PR #1, opened 2021-02-23
Initial openapi.yaml

lukehinds:


openapi.yaml
```diff
@@ -0,0 +1,69 @@
...
+            type: string
+          description: Invalid Data
+        500:
+          schema:
+            type: string
+          description: Server error
+definitions:
+  Submit:
+    type: object
+    required: [email,common_name,country,province,organization,organization_unit]
```
lukehinds, line 47: Not sure if this should be outlined here, as I expect the input would be a single JWT base64 encoded. We would then verify the signing and decode to then serialise the values from there.

openapi.yaml
```diff
@@ -0,0 +1,69 @@
+swagger: '2.0'
+info:
+  version: '1.0.0'
+  title: Fulcio
+schemes: [http, https]
+host: fulcio.sigstore.dev
+basePath: /v1
```
bobcallaway, line 7: I'd prepend `/api/` to this so if you wanted to have a webpage with instructions/etc you can safely do that.

openapi.yaml
```diff
@@ -0,0 +1,69 @@
...
+  Bearer:
+    type: apiKey
+    name: Authorization
+    in: header
+paths:
+  /submitcsr:
+    post:
+      description: 'Request API for signing certifcate'
+      operationId: SubmitCSR
+      consumes:
```
bobcallaway, line 20: this is redundant since you specify it above

openapi.yaml
```diff
@@ -0,0 +1,69 @@
...
+        - "application/json"
+      parameters:
+        - name: 'submitcsr'
+          in: 'body'
+          required: true
+          description: 'Submit CSR JWT Payload'
+          schema:
+            $ref: '#/definitions/Submit'
+      responses:
+        200:
```
bobcallaway, line 30: should be 201 if you created the cert

openapi.yaml
```diff
@@ -0,0 +1,69 @@
...
+      locality:
+        type: string
+      organization:
+        type: string
+      organization_unit:
+        type: string
+  SubmitSuccess:
+    type: object
+    properties:
+      success:
```
bobcallaway, line 66: not needed; let the HTTP response code indicate success/failure

openapi.yaml
```diff
@@ -0,0 +1,69 @@
...
+basePath: /v1
+consumes: [application/json]
+produces: [application/json]
+securityDefinitions:
+  Bearer:
+    type: apiKey
+    name: Authorization
+    in: header
+paths:
+  /submitcsr:
```
bobcallaway, line 16: wondering about making the externals:

POST `/signingCert` : create a cert, return content with a location header (with URL to CTL entry)

openapi.yaml
```diff
@@ -0,0 +1,69 @@
...
+          schema:
+            $ref: '#/definitions/Submit'
+      responses:
+        200:
+          description: Successful CSR Submit
+          schema:
+            $ref: '#/definitions/SubmitSuccess'
+        400:
+          description: Bad Request
+        404:
```
bobcallaway, line 36: you shouldn't get a 404 ever from a POST since you're creating a resource. should be 400 if the content is malformed

openapi.yaml
```diff
@@ -0,0 +1,69 @@
...
+          schema:
+            type: string
+          description: Invalid Data
+        500:
```
bobcallaway, line 40: You need 401 unauthorized here since we're requiring authorization to call this

openapi.yaml
```diff
@@ -0,0 +1,69 @@
...
+info:
+  version: '1.0.0'
+  title: Fulcio
+schemes: [http, https]
+host: fulcio.sigstore.dev
+basePath: /v1
+consumes: [application/json]
+produces: [application/json]
+securityDefinitions:
+  Bearer:
```
bobcallaway, line 11: I think this should be OAuth2

openapi.yaml
bobcallaway: Since you'll have a signed JWT, I don't think it makes sense to specify the schema here. I would just have a single property:

```
type: string
format: byte
```

openapi.yaml
lukehinds: don't we need to outline the return (certificate)

```
certificate:
        type: string
```

openapi.yaml
lukehinds: understood, was from an example , but agree a 404 should only be for a GET

openapi.yaml
lukehinds: OAuth2 would be the Authentication Server {keycloak,kong}, I expect this would instead be a scoped JWT token?

openapi.yaml
bobcallaway: yup - the boolean was just redundant

lukehinds: ack, updated.

openapi.yaml
lukehinds: what's view on this now @bobcallaway ?

bobcallaway (approved):

PR #4, opened 2021-03-02
initial server impl with go-swagger

bobcallaway:


Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM golang:1.16.0 AS builder
```
lukehinds, line 1: This is not about your pull request as such, but do we fix to a particular release of go? I grabbed the CI testing yaml from rekor and that's running on 1.15?

lukehinds (approved):

Dockerfile
bobcallaway: I think in rekor we have a minimum of 1.15 for some of the crypto APIs used.

PR #21, opened 2021-03-08
Docker compose / deployment files for ct server

lukehinds:
This is still a WIP and not yet ready to merge.

Just some issues to fix up around mount points


lukehinds:
Just some notes:

* ca-roots.pem, may actually be present in the gcr image. if so will remove. However we will want to add the fulcio root cert there at some point, although it could be a test dummy one
* I still have some issues around the ct server accessing the config file
* k8s yaml needs work still (ct config file again)

PR #27, opened 2021-03-10
Use 'Authorization Code with PKCE' OAuth2 flow

bobcallaway:
This change adds the use of the PKCE challenge and verifier into the
OAuth2 flow in the CLI to be compliant with RFC 8252. It also adds the
optional nonce parameter.

It also converts the Fulcio server to using the raw id_token as the
Authorization header Bearer token, and verifies that:
 - the token was signed by the specified OIDC identity provider
 - the client ID matches what is expected
 - the email address in the id_token was verified at some point in past

This change also moves the OIDC settings into command line arguments,
and shrinks the number of HTTP calls used in the OAuth2 dance.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 .PHONY: all test clean lint gosec
 
-all: server
+all: server client
 
 GENSRC = pkg/generated/models/%.go pkg/generated/restapi/%.go
 OPENAPIDEPS = openapi.yaml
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -23,11 +23,14 @@ gosec:
...
 
+client: $(SRCS)
+	go build ./cmd/client
+
 test:
 	go test ./...
 
 clean:
-	rm -rf server
+	rm -rf server client
```
lukehinds, line 33: tiny nit, may as well prefix the current path `./server` `./client` as someone may not have it in their $PATH (which is likely there fault, but it does no harm having them there).

lukehinds (approved):
nice patch bob!

Makefile
bobcallaway: not following you - these paths mean the files named `client` and `server` in `$CWD`... even if you run make clean from a different dir, it only refers to files in the same directory as the Makefile

PR #35, opened 2021-03-12
Improve PKCE code and use ksuid for state/nonce

bobcallaway:
Previously the code presumed the support of PKCE by a configured OIDC
identity provider. This patch verifies support from the OIDC discovery 
document and forces the use of the S256 method.

Also, OAuth2 state and nonce values (as well as the PKCE code verifier)
have been switched to using ksuid instead of simply random values to
improve protection against replay attacks.

### DIFF
--- a/pkg/oauthflow/flow.go
+++ b/pkg/oauthflow/flow.go
@@ -18,8 +18,6 @@ package oauthflow
 
 import (
 	"context"
-	"crypto/rand"
-	"encoding/base64"
 	"errors"
 	"fmt"
 	"net/http"
@@ -27,7 +25,8 @@ import (
 	"os"
 	"time"
 
-	"github.com/coreos/go-oidc"
+	"github.com/coreos/go-oidc/v3/oidc"
+	"github.com/segmentio/ksuid"
 	"github.com/skratchdot/open-golang/open"
 	"github.com/spf13/viper"
 	"golang.org/x/oauth2"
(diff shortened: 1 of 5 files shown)


pkg/oauthflow/pkce.go
```diff
@@ -18,38 +18,57 @@ package oauthflow
...
 	}
 
-	value, err := randStr()
-	if err != nil {
+	if err := provider.Claims(&providerClaims); err != nil {
+		// will only error out if the JSON was malformed, which shouldn't happen at this point
 		return nil, err
 	}
 
+	var chosenMethod string
```
dlorenc, line 50: This section terrifies me for downgrade attacks: https://tools.ietf.org/html/rfc7636#section-7.2

Who doesn't support S256? Do we care?

pkg/oauthflow/flow.go
```diff
@@ -162,15 +160,10 @@ func getCodeFromLocalServer(state string, redirectURL *url.URL) (string, error)
...
-func randStr() (string, error) {
-	len := 32
-	b := make([]byte, len)
-	n, err := rand.Read(b)
-	if err != nil {
-		return "", err
-	} else if n != len {
-		return "", errors.New("short read")
-	}
+func randStr() string {
```
dlorenc, line 163: nit: we can just drop the randStr method rather than using it as a stub

pkg/oauthflow/pkce.go
```diff
@@ -18,38 +18,57 @@ package oauthflow
...
+	for _, method := range providerClaims.CodeChallengeMethodsSupported {
```
dlorenc, line 51: This whole section scares me for downgrade attacks.It looks like you got it right but I still don't like it. Does anyone not support S256? Do we care?

https://tools.ietf.org/html/rfc7636#section-7.2

pkg/oauthflow/flow.go
bobcallaway: Yeah, I debated doing that but liked having it there for the comments and rationale for why we're using ksuid. I can be convinced otherwise though :)

dlorenc: Ah ok, that's fair.

dlorenc (approved):

PR #72, opened 2021-04-02
bump swagger to v0.27.0

bobcallaway:
we should make sure that rekor & fulcio stay in sync if at all possible on swagger versions; bumping to latest since v0.26.1 has a defect that is incompatible with Rekor's codebase.

Also sets GOVERSION from Dockerfile rather than having a separate variable in GitHub actions definition.

Signed-off-by: Bob Callaway <bob.callaway@gmail.com>


lukehinds (approved):

pkg/generated/restapi/operations/signing_cert.go
```diff
@@ -61,7 +61,7 @@ type SigningCert struct {
 func (o *SigningCert) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 	route, rCtx, _ := o.Context.RouteInfo(r)
 	if rCtx != nil {
-		r = rCtx
+		*r = *rCtx
```
lukehinds, line 64: do you know why these got converted to pointers?

bobcallaway: this comes from the commit https://github.com/go-swagger/go-swagger/pull/2553/commits/43f6f80154a4259054764f809b83a624e9c7b171 

since I think they're trying to actually update the value at `r`, they do a shallow copy of the values into the actual memory VS just updating the local copy of the pointer.

PR #77, opened 2021-04-07
Add generate subcommand to client to create certs

priyawadhwa:
This lets users create certs without having to go through the OIDC flow!

used like this:

```
./client generate --public-key-file <> --gcp-private-ca-parent <> --uri <> --ct-log-url <>
```

I used it to create a fulcio signed cert for the rekor public key (https://github.com/sigstore/rekor/issues/256).

### DIFF
--- a/cmd/client/app/generate.go
+++ b/cmd/client/app/generate.go
@@ -1,0 +1,125 @@
+/*
+Copyright © 2021 Dan Lorenc <lorenc.d@gmail.com>
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package app
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+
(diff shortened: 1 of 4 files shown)


cmd/client/app/generate.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	parent := viper.GetString("gcp-private-ca-parent")
+	req := cert(parent, uri, pemBytes)
+	resp, err := ca.Client().CreateCertificate(ctx, req)
+	if err != nil {
+		return errors.Wrap(err, "creating cert")
+	}
+
+	// Submit to CTL
+	log.Logger.Info("Submitting cert to CT log")
```
lukehinds, line 68: Just thought, i wonder if we might want this to be optional. I can certainly see why in most cases a CTL entry would be desirable, but there may be cases where someone just wants a cert and would like to skip entry. I say that assuming this is intended as an offline type tool for an admin to use.

cmd/client/app/generate.go
```diff
@@ -0,0 +1,133 @@
...
+				},
+			},
+		},
+	}
+}
+
+func init() {
+	generateCmd.PersistentFlags().String("gcp-private-ca-parent", "projects/project-rekor/locations/us-central1/certificateAuthorities/sigstore", "private ca parent: /projects/<project>/locations/<location>/<name>")
+	generateCmd.PersistentFlags().StringVar(&pubKeyFile, "public-key-file", "", "path to the PEM encoded public key")
+	generateCmd.PersistentFlags().StringVar(&uri, "uri", "", "uri for the cert")
```
lukehinds, line 126: what does the uri do?

priyawadhwa:
closing in favor of https://github.com/sigstore/fulcio/pull/87

PR #78, opened 2021-04-12
Owners file

lukehinds:


dlorenc:
I kind of prefer "MAINTAINERS" if we're going to list these, it also starts to raise the question of how people get added to the file. Should we tackle all of that at once?

PR #86, opened 2021-05-02
Add Key Encipherment to satisfy Microsoft code signing cert requirements

pearcec:
Signed-off-by: Christian Pearce <christian@pearcec.com>


pearcec:
Was able to extract a PFX file by hacking up cosign. This validates we don't need this Enhanced Key Usage.

PR #87, opened 2021-05-05
Add CLI option to generate a cert for timestamping

asraa:
Let's users also create a cert issued for timestamping through the CLI. Built on top of:
https://github.com/sigstore/fulcio/pull/77

Co-authored-by: Priya Wadhwa <priyawadhwa@google.com>
Signed-off-by: Asra Ali <asraa@google.com>

### DIFF
--- a/cmd/client/app/generate.go
+++ b/cmd/client/app/generate.go
@@ -87,27 +87,34 @@ func generateCert(ctx context.Context) error {
 
 func cert(parent, uri string, pemBytes []byte, timestamping bool) *privatecapb.CreateCertificateRequest {
 	baseKeyUsage := privatecapb.KeyUsage_KeyUsageOptions{}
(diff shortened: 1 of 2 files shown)


lukehinds:
We should try to move this forward. Is this needed for the offline signature bundles. This could really use a linked issue to describe what work this is part of.

cmd/client/app/generate.go
```diff
@@ -0,0 +1,159 @@
+/*
+Copyright © 2021 Dan Lorenc <lorenc.d@gmail.com>
```
hectorj2f, line 2: Shouldn't this be `Copyright 2021 The Sigstore Authors.` ?

hectorj2f: Oh this is an old PR, nvm

asraa:
@hectorj2f @lukehinds Right now I can close it, we shouldn't need it except to generate certs by hand for rekor's keys. End users shouldn't need this

PR #93, opened 2021-05-10
Query PublicKey_KeyType

lukehinds:
This is required to send the correct Key Type needed
by GCP CreateCertificateRequest and allow the signing
of both RSA and ECDSA


dlorenc (approved):

dlorenc:
Looks like there's a lint issue still

PR #96, opened 2021-05-16
Allow for multiple OIDC providers.

dlorenc:

### DIFF
--- a/cmd/app/root.go
+++ b/cmd/app/root.go
@@ -52,10 +52,15 @@ func Execute() {
 	}
 }
 
+var defaultIssuers = []string{
+	"https://oauth2.sigstore.dev/auth",
+	"https://accounts.google.com",
+}
+
 func init() {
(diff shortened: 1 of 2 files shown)


dlorenc:
I think we might want a way to "hint" at the proper issuer via a header or query parameter here. We could pull it out of the JWT itself, but that seems more error-prone.

Going through them in a list is fine but we should probably add the hint if we want to support more than a few.

lukehinds:
What's the use case here, is it fail over type system if the first provider is unavailable (iterate over several providers?)

dlorenc:
> What's the use case here, is it fail over type system if the first provider is unavailable (iterate over several providers?)

Left some more text in the description, but I think this would make eventual federation and headless modes easier.

bobcallaway (approved):
use case makes sense... would be nice to specify client ID separately for each issuer but for now this should be fine

lukehinds (approved):

dlorenc:
> use case makes sense... would be nice to specify client ID separately for each issuer but for now this should be fine

Opened #100 to track that

PR #99, opened 2021-05-20
Fix paths in Dockerfile for local development

priyawadhwa:
Update the paths since #91 was merged!


lukehinds (approved):
nice catch, I missed this during the client code refactor.

Issue #44, opened 2021-03-15
Capture the IDP somewhere in the cert


dlorenc:
We should log **who** said the challenge was correct somewhere. So instead of just a cert issued to an email, we say "google validated the identity token", in case of an issue with the OIDC providers.

dlorenc:
From @bobcallaway:

> we should record the "issuer" and the other claim that is the identifer for the identity


I'm not sure we should store the other one, I'll have to think about it. If multiple email addresses are tied to the same identifier, we probably don't want to track that.

dlorenc:
Ah yeah - we can find a field for this or use our own OID.

dlorenc:
ref #80 as well.

raesene:
Just to expand on what my thought from Slack, what I was thinking about is the threat model where an attacker is able to register an e-mail address at one of the providers that a user doesn't have an account on, and then sign an artifact using that identity. 

Addressing this at a basic level seems like it needs a couple of things:

- The OIDC provider used to authenticate the user needs to be recorded at the time of the signing.
- The UI of the signing tool needs to display the provider alongside the validated e-mail address.
- The documentation needs to state that people who are signing artifacts should make it clear which provider(s) they are using for the process to their users, and also that people who want to verify a signature need to check that the combination of e-mail address and provider matches what was expected for that artifact.

dlorenc:
Yeah!

dlorenc:
For reference: https://github.com/sigstore/fulcio/pull/211

PR #102, opened 2021-05-22
Remove the common name field from the certs.

dlorenc:
This was overkill and the field is deprecated anyway. We only check the SAN.


bobcallaway (approved):
this is fine; we still need to solve #44

Issue #100, opened 2021-05-20
Figure out how to properly configure multiple OIDC issuers


dlorenc:
#96 

Right now we're fine with a single "audience" or "client id" field across our issuers, but it's possible we might need different ones (and even client secrets) for each individual one.

cpanato:
fixed by https://github.com/sigstore/fulcio/pull/105

closing

Issue #22, opened 2021-03-08
Figure out how to support other OIDC providers


dlorenc:
There's no canonical mapping of email address to authoritative OIDC provider. This means we can't just let users provide their own configs. For example:

* Someone could stand up their own OIDC instance and issue tokens for any email address, if we trusted all providers blindly we would mis-issue certs.
  * We would be able to detect this and recover from it with transparency logs, but it's still not good.

So, we'll need a way to add OIDC providers **that we trust** one by one to cover most email addresses.

For users that truly want to run their own email systems, we have a couple options:

- ACME-style DNS challenges on MX records to ensure the user has control of email routing for that domain
- Email-based challenges - we send a token, they enter it. Email is wildly insecure, but if we want to support complete federation we don't have too many other options. This is hard to automate as well.
- Our own login system
  - I don't love this idea, but it does give us flexibility. Login to sigstore with an account (OIDC is fine too), and strengthen that account with other challenges (similar to keybase). We can verify access to an email address here however we need to once a week, month, etc. manually, then automatically issue certs using our own oauth challenges.

bobcallaway:
Note here for myself and others: the `email_verified` claim simply notes that when verification was requested, the provider was able to successfully verify ownership of the email account... however this could have been quite some time in the past and as such isn't sufficient to trust a claim.

cpanato:
fixed by https://github.com/sigstore/fulcio/pull/105

closing

PR #105, opened 2021-05-28
Move OIDC configuration to a nested JSON config file.

dlorenc:
#22

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,0 +1,83 @@
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package config
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+
+	"github.com/sigstore/fulcio/pkg/log"
+)
+
+type FulcioConfig struct {
+	OIDCIssuers map[string]OIDCIssuer
+}
+
+type OIDCIssuer struct {
+	IssuerURL string
+	ClientID  string
+}
+
+func ParseConfig(b []byte) (FulcioConfig, error) {
+	cfg := FulcioConfig{}
+	if err := json.Unmarshal(b, &cfg); err != nil {
+		return FulcioConfig{}, err
+	}
+	return cfg, nil
+}
+
+var DefaultConfig = FulcioConfig{
+	OIDCIssuers: map[string]OIDCIssuer{
+		"https://oauth2.sigstore.dev/auth": {
+			IssuerURL: "https://oauth2.sigstore.dev/auth",
+			ClientID:  "sigstore",
+		},
+		"https://accounts.google.com": {
+			IssuerURL: "https://accounts.google.com",
+			ClientID:  "sigstore",
+		},
+	},
+}
+
+var config *FulcioConfig
+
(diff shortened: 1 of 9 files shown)


pkg/config/config.go
```diff
@@ -0,0 +1,83 @@
...
+		"https://accounts.google.com": {
+			IssuerURL: "https://accounts.google.com",
+			ClientID:  "sigstore",
+		},
+	},
+}
+
+var config *FulcioConfig
+
+func Config() FulcioConfig {
```
cpanato, line 58: why you return the config thing if that is a global variable? asking to understand and to see if this is a good practice.

maybe we not make the config global  and have something like

func (c * FulcioConfig) Config()....

cpanato: ahh, right it is global in the package.

ignore me. 🤦

cpanato (approved):

pkg/config/config.go
dlorenc: Ah yeah, I don't love this approach but I like it more than stuffing this kind of thing into context variables... Not sure if there's a better way.

pkg/generated/restapi/configure_fulcio_server.go
```diff
@@ -195,13 +194,8 @@ func logAndServeError(w http.ResponseWriter, r *http.Request, err error) {
...
 	if e, ok := err.(goaerrors.Error); ok && e.Code() == http.StatusUnauthorized {
-		issuerURL, parseErr := url.Parse(viper.GetString("oidc-issuer"))
-		if parseErr != nil {
-			log.RequestIDLogger(r).Error(fmt.Errorf("error parsing OIDC provider: %w", parseErr))
-		} else {
-			// this is set directly so the header name is not canonicalized
-			w.Header()["WWW-Authenticate"] = []string{fmt.Sprintf("Bearer realm=\"%v\",scope=\"openid email\"", issuerURL.Host)}
-		}
+		// this is set directly so the header name is not canonicalized
+		w.Header()["WWW-Authenticate"] = []string{"Bearer realm=\"oauth2.sigstore.dev\",scope=\"openid email\""}
```
bobcallaway, line 198: nit: this should probably iterate across all issuers, like this:

```go
var b strings.Builder
for i, iss := range fulcioCfg.OIDCIssuers {
	if i != 0 {
		b.WriteString(", ")
	}
	b.WriteString("Bearer realm=\""+iss.IssuerURL+"\",scope=\"openid email\"")
}
w.Header()["WWW-Authenticate"] = []string{b.String()}
```

note this doesn't create multiple WWW-Authenticate response headers but instead returns 1 with all relevant challenges concatenated per https://datatracker.ietf.org/doc/html/rfc7235#section-4.1

bobcallaway:
Also this

Fixes #100

pkg/generated/restapi/configure_fulcio_server.go
dlorenc: nice!

dlorenc: had to tweak it a bit since it's a map not a slice...

pkg/generated/restapi/configure_fulcio_server.go
```diff
@@ -195,13 +194,14 @@ func logAndServeError(w http.ResponseWriter, r *http.Request, err error) {
...
-		} else {
-			// this is set directly so the header name is not canonicalized
-			w.Header()["WWW-Authenticate"] = []string{fmt.Sprintf("Bearer realm=\"%v\",scope=\"openid email\"", issuerURL.Host)}
+		// this is set directly so the header name is not canonicalized
+		fulcioCfg := config.Config()
+		issuers := []string{}
+		for iss := range fulcioCfg.OIDCIssuers {
+			issuers = append(issuers, "Bearer realm=\""+iss+"\",scope=\"openid email\"")
 		}
+		strings.Join(issuers, ", ")
```
bobcallaway, line 203: Is this duplicative? LGTM otherwise

dlorenc: Whoa, surprised the linter didn't catch that.

bobcallaway (approved):

PR #107, opened 2021-05-30
Support SPIFFE challenges instead of just emails.

dlorenc:
This also refactors the code a bit so each "issuer" can declare which
type of subject it will grant. The current types are email and spiffe.
Each type can set different fields in the final cert.

### DIFF
--- a/pkg/api/signing_cert.go
+++ b/pkg/api/signing_cert.go
@@ -16,12 +16,15 @@
 package api
 
 import (
-	"crypto/x509"
+	"context"
 	"encoding/pem"
 	"errors"
 	"fmt"
 	"net/http"
 	"strings"
+
+	"github.com/sigstore/fulcio/pkg/challenges"
+	"github.com/sigstore/fulcio/pkg/config"
 
 	"github.com/coreos/go-oidc/v3/oidc"
 	"github.com/go-openapi/runtime/middleware"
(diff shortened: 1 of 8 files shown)


pkg/api/signing_cert.go
```diff
@@ -31,7 +34,7 @@ import (
 	"github.com/sigstore/fulcio/pkg/ctl"
 	"github.com/sigstore/fulcio/pkg/generated/restapi/operations"
 	"github.com/sigstore/fulcio/pkg/log"
-	"github.com/sigstore/fulcio/pkg/oauthflow"
```
lukehinds, line 34: It would be nice if we could make this modular. I feel like this is locking into one provider and we are going to have to refactor it at a later point to support other CAs.

dlorenc: This PR doesn't actually change or add the CA integration, it just refactors it a bit. Agree we should add some kind of intermediate layer.

lukehinds: It's a fair point in that there is no other CA right now to have a modular layer, I will open an issue for this.

dlorenc: Thanks!

priyawadhwa (approved):

lukehinds (approved):

PR #115, opened 2021-06-08
Implement modular CA and SoftHSM integration

lukehinds:


README.md
```diff
@@ -54,9 +54,87 @@ We encourage auditors to monitor this log, and aim to help people access the dat
...
+
+* Note: you can import existing keys and import using pkcs11-tool, see pkcs11-tool manual for details
+
+### Create a root CA
+
+Now that your keys are generated, you can use the fulcio `createca` command to generate a Root CA. This command
+will also store the generated Root CA into the HSM by the delegated id passed to `--hsm-caroot-id`
+
+```
+fulcio createca--org=acme --country=UK --locality=SomeTown --province=SomeProvince --postal-code=XXXX --street-address=XXXX --hsm-caroot-id 99 --out myrootCA.pem
```
bobcallaway, line 121: space b/t `createca` and `--org`?

README.md
```diff
@@ -54,9 +54,87 @@ We encourage auditors to monitor this log, and aim to help people access the dat
...
+```
+
+### Run fulcio with the SoftHSM
+
+```
+fulcio serve --ca fulcioca --hsm-caroot-id 99
```
bobcallaway, line 127: should we name this `softhsm` instead of `fulcioca`?

cmd/app/createca.go
```diff
@@ -0,0 +1,153 @@
...
+			log.Logger.Fatal(err)
+		}
+		defer p11Ctx.Close()
+
+		rootID := []byte(viper.GetString("hsm-caroot-id"))
+
+		// Check if CA already exists (or a cert within the provided ID)
+		findCA, err := p11Ctx.FindCertificate(rootID, nil, nil)
+		if err != nil {
+			panic(err)
```
bobcallaway, line 58: why log.Fatal() some places but panic() in others?

cmd/app/createca.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+
+		// Find the existing Key Pair
+		// TODO: We could make the TAG customizable
+		log.Logger.Info("finding slot for private key \"FulcioCA\"")
+		privKey, err := p11Ctx.FindKeyPair(nil, []byte("FulcioCA"))
+		if err != nil {
+			log.Logger.Fatal(err)
+		}
+		pubKey := privKey.Public().(crypto.PublicKey).(*ecdsa.PublicKey)
```
bobcallaway, line 71: do you need to check the algorithm for the public key before assuming its ecdsa?

cmd/app/createca.go
```diff
@@ -0,0 +1,153 @@
...
+			panic(err)
+		}
+		if findCA != nil {
+			panic("certificate already exists with this ID")
+		}
+
+		// Find the existing Key Pair
+		// TODO: We could make the TAG customizable
+		log.Logger.Info("finding slot for private key \"FulcioCA\"")
+		privKey, err := p11Ctx.FindKeyPair(nil, []byte("FulcioCA"))
```
bobcallaway, line 67: if the tag isn't customizable then put it into a `const` at top of file?

cmd/app/createca.go
```diff
@@ -0,0 +1,153 @@
...
+		// TODO: We could make the TAG customizable
+		err = p11Ctx.ImportCertificateWithLabel(rootID, []byte("FulcioCA"), certParse)
+		if err != nil {
+			log.Logger.Fatal(err)
+		}
+		log.Logger.Info("root CA created with PKCS11 ID: ", viper.GetString("hsm-caroot-id"))
+
+		// Save out the file in pem format for easy import to CTL chain
+		if viper.IsSet("out") {
+			certOut, err := os.Create(viper.GetString("out"))
```
bobcallaway, line 125: do you need to clean the file path here?

cmd/app/createca.go
```diff
@@ -0,0 +1,153 @@
...
+			if err != nil {
+				log.Logger.Fatal(err)
+			}
+			if err := pem.Encode(certOut, &pem.Block{ //nolint
+				Type:  "CERTIFICATE",
+				Bytes: caBytes},
+			); err != nil {
+				log.Logger.Fatal(err)
```
bobcallaway, line 133: `certOut.Close()` needs to go above this line or deal with a deferred call to `Close`

pkg/api/fulcioca_signing_cert.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+func FulcioCASigningCertHandler(params operations.SigningCertParams, principal *oidc.IDToken) middleware.Responder {
+
+	// none of the following cases should happen if the authentication path is working correctly; checking to be defensive
+	if principal == nil {
+		return handleFulcioAPIError(params, http.StatusInternalServerError, errors.New("no principal supplied to request"), invalidCredentials)
+	}
+
+	emailAddress, emailVerified, err := oauthflow.EmailFromIDToken(principal)
```
bobcallaway, line 45: will this work with the SPIFFE code?

pkg/api/fulcioca_signing_cert.go
```diff
@@ -0,0 +1,129 @@
...
+	// Perform a proof challenge verification
+	if err := fulcioca.CheckSignature(pkixPubKey, *params.CertificateRequest.SignedEmailAddress, emailAddress); err != nil {
+		return handleFulcioAPIError(params, http.StatusBadRequest, err, invalidSignature)
+	}
+
+	// Use a generic Cert failure message as we don't want to tell the client about a HSM failure and risk exposure of
+	// internal security controls
+	p11Ctx, err := pkcs11.InitHSMCtx()
+	if err != nil {
+		return handleFulcioAPIError(params, http.StatusInternalServerError, err, failedToCreateCert)
```
bobcallaway, line 67: change `err` to `errors.Wrap(err, "initializing HSM context")` so when this gets logged you have some additional information present

pkg/api/fulcioca_signing_cert.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	defer p11Ctx.Close()
+
+	rootID := []byte(viper.GetString("hsm-caroot-id"))
+
+	// get the existing root CA from the HSM
+	rootCA, err := p11Ctx.FindCertificate(rootID, nil, nil)
+	if err != nil {
+		return handleFulcioAPIError(params, http.StatusInternalServerError, err, failedToCreateCert)
```
bobcallaway, line 76: same as 67

pkg/api/fulcioca_signing_cert.go
```diff
@@ -0,0 +1,129 @@
...
+	})
+
+	// Format in PEM
+	rootPEM := pem.EncodeToMemory(&pem.Block{
+		Type:  "CERTIFICATE",
+		Bytes: rootCA.Raw,
+	})
+
+	// Submit to CTL
+	log.Logger.Info("Submitting CTL inclusion for OIDC grant: ", subject)
```
bobcallaway, line 105: please change all the logging statements to be:
`log.RequestIDLogger(params.HTTPRequest).Info()` so we can correlate with the HTTP transaction

pkg/ca/fulcioca/fulcioca.go
```diff
@@ -0,0 +1,72 @@
...
+			return fmt.Errorf("signature could not be verified: %v", err)
+		}
+	}
+
+	return nil
+}
+
+func CreateClientCertificate(rootCA *x509.Certificate, emailAddress string, publicKeyPEM interface{}, privKey crypto11.Signer) ([]byte, error) {
+	// TODO: Track / increment serial nums instead, although unlikely we will create dupes, it could happen
+	serialNumber, err := rand.Int(rand.Reader, new(big.Int).SetInt64(math.MaxInt64))
```
bobcallaway, line 53: I'd do this instead:
```
import "github.com/google/uuid"`
uuid = uuid.New()
var serialNumber big.Int
serialNumber.SetBytes(uuid[:])
```

pkg/generated/restapi/configure_fulcio_server.go
```diff
@@ -95,10 +98,29 @@ func configureAPI(api *operations.FulcioServerAPI) http.Handler {
 		if idToken == nil {
 			return nil, goaerrors.New(http.StatusUnauthorized, strings.Join(errs, ","))
 		}
+		if _, ok, err := oauthflow.EmailFromIDToken(idToken); !ok || err != nil {
```
bobcallaway, line 101: I think this will break SPIFFE support since it won't have an email in the token

pkg/generated/restapi/configure_fulcio_server.go
```diff
@@ -95,10 +98,29 @@ func configureAPI(api *operations.FulcioServerAPI) http.Handler {
...
+	// Currently supported:
+	// googleca: Google Certficate Authority Service
+	// fulcio: Generic PKCS11 / HSM backed service
+	CAType := viper.GetString("ca")
+	switch CAType {
+	case "googleca":
+		api.SigningCertHandler = operations.SigningCertHandlerFunc(pkgapi.GoogleCASigningCertHandler)
+	case "fulcioca":
+		api.SigningCertHandler = operations.SigningCertHandlerFunc(pkgapi.FulcioCASigningCertHandler)
+	default:
```
bobcallaway, line 121: if it is neither, should we bail since we don't have the information to actually interact with the googleca?

README.md
lukehinds: The thinking is that this is a pkcs11 implementation and so it should work with HW HSM's as well (the SoftHSM specifics are in the config file). So this becomes fulcioca (as it creates and signs the client certificates in much the same manner as googleca does).

cmd/app/createca.go
lukehinds: Good point, I read up more on both approaches and went with Log.Fatal, if these should be panic I can swap approach.

cmd/app/createca.go
lukehinds: I certainly do, addressed in next commit!

cmd/app/createca.go
lukehinds: yup, done.

pkg/api/fulcioca_signing_cert.go
lukehinds: done

pkg/ca/fulcioca/fulcioca.go
lukehinds: nice change, done!

pkg/generated/restapi/configure_fulcio_server.go
lukehinds: good call!

lukehinds:
@bobcallaway , solid review thanks. Everything above is addressed, apart from spiffe/spire which I need to look into a little more. I will tag you when that is pushed.

cmd/app/createca.go
lukehinds: could you expand on this? how do you mean by clean?

pkg/api/fulcioca_signing_cert.go
lukehinds: done

pkg/api/fulcioca_signing_cert.go
lukehinds: done

cmd/app/createca.go
bobcallaway: do `certOut, err := os.Create(filepath.Clean(viper.GetString("out")))`

cmd/app/createca.go
```diff
@@ -19,22 +19,40 @@ import (
...
+)
 
-	"github.com/sigstore/fulcio/pkg/log"
+const (
+	label = "FulcioCA"
 )
 
+func publicKeyFromPrivate(privKey crypto.PrivateKey) (crypto.PublicKey, error) {
+	var err error
+	switch key := privKey.(type) {
```
bobcallaway, line 45: you're returning a `crypto.PublicKey` here, which means there's no point in casting the return value to the algorithm-specific value. This would work just the same:

```
switch key := privKey.(type) {
case *ecdsa.PrivateKey, *rsa.PrivateKey:
	return key.Public(), nil
default:
	return nil, errors.New("error generating public key")
}
```

cmd/app/createca.go
```diff
@@ -19,22 +19,40 @@ import (
...
 
 	"github.com/ThalesIgnite/crypto11"
+	"github.com/sigstore/fulcio/pkg/log"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+)
 
-	"github.com/sigstore/fulcio/pkg/log"
+const (
+	label = "FulcioCA"
```
bobcallaway, line 40: for readability, I might do all caps for constant `LABEL` or even something slightly more specific

bobcallaway (commented):
a couple other small comments

cmd/app/createca.go
lukehinds: This does not work, having the case match both leave Key undefined:

`cmd/app/createca.go:49:13: key.Public undefined (type crypto.PrivateKey is interface with no methods)`

lukehinds: I get what you mean here, in fact I don't need to any checks at all.

this works fine: `pubKey := privKey.Public().(crypto.PublicKey)`

This also made me check rsa i the HSM and its all good.

pkg/api/googleca_signing_cert.go
```diff
@@ -0,0 +1,67 @@
...
+	log.Logger.Infof("requesting cert from %s for %v", parent, subject)
+
+	resp, err := googleca.Client().CreateCertificate(ctx, req)
+	if err != nil {
+		return "", nil, err
+	}
+	return resp.PemCertificate, resp.PemCertificateChain, nil
+}
+
+func subject(ctx context.Context, tok *oidc.IDToken, cfg config.FulcioConfig, publicKey, challenge []byte) (*challenges.ChallengeResult, error) {
```
bobcallaway, line 57: this method should be moved somewhere outside of the google-specific file

config/deployment.yaml
```diff
@@ -42,14 +42,16 @@ spec:
         - containerPort: 2112 # metrics
         args: [
           "serve",
-          "--host=0.0.0.0", "--port=5555", "--gcp_private_ca_parent=$(CA_PARENT)", "--ct-log-url=http://ct-log/test",
+          "--host=0.0.0.0", "--port=5555", "--ca fulcioca", "--hsm-caroot-id 1", "--ct-log-url=http://ct-log/test",
```
cpanato, line 45: change to 
```
"--host=0.0.0.0", "--port=5555", "--ca=fulcioca", "--hsm-caroot-id=1", "--ct-log-url=http://ct-log/test",
```

config/deployment.yaml
```diff
@@ -42,14 +42,16 @@ spec:
...
         ]
         env:
         - name: CA_PARENT
           valueFrom:
             configMapKeyRef:
               name: private-ca
               key: connection
+        - name: CGO_ENABLED
```
cpanato, line 53: do we need this?

cpanato: tested locally and works without this env var

bobcallaway (commented):
a few small nits but overall looking good

pkg/api/ca.go
```diff
@@ -55,24 +52,28 @@ func SigningCertHandler(params operations.SigningCertParams, principal *oidc.IDT
...
+	case "googleca":
+		PemCertificate, PemCertificateChain, err = GoogleCASigningCertHandler(ctx, *subj, publicKeyPEM)
+	case "fulcioca":
+		PemCertificate, PemCertificateChain, err = FulcioCASigningCertHandler(*subj, publicKey)
+	default:
+		return handleFulcioAPIError(params, http.StatusInternalServerError, err, fmt.Sprintf(invalidCA, viper.GetString("ca")))
+	}
 	if err != nil {
-		return handleFulcioAPIError(params, http.StatusInternalServerError, err, failedToCreateCert)
+		return handleFulcioAPIError(params, http.StatusInternalServerError, err, fmt.Sprintf(invalidCA, viper.GetString("ca")))
```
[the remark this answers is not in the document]
lukehinds: this merits some discussion, as we really want to be careful of kicking back internal errors to users. I am of mind we use a very non descriptive error here.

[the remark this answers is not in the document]
bobcallaway: +1 on not passing back the value of `err` unfiltered... how about something like `"error communicating with CA backend"`?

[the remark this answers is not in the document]
lukehinds: agree

pkg/challenges/challenges.go
```diff
@@ -25,15 +29,40 @@ import (
...
+	EmailValue ChallengeType = iota
+	SpiffeValue
+)
+
+type ChallengeResult struct {
+	TypeVal ChallengeType
+	Value   string
+}
+
+func CheckSignature(pub crypto.PublicKey, proof []byte, email string) error {
```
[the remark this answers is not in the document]
lukehinds: any ideas on where?

I thought as its part of the challenge (Check the signed email proof) stick it in challenges?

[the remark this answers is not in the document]
bobcallaway: never mind - it is just fine here

bobcallaway (approved):

config/deployment.yaml
```diff
@@ -42,7 +42,7 @@ spec:
         - containerPort: 2112 # metrics
         args: [
           "serve",
-          "--host=0.0.0.0", "--port=5555", "--gcp_private_ca_parent=$(CA_PARENT)", "--ct-log-url=http://ct-log/test",
+          "--host=0.0.0.0", "--port=5555", "--ca=fulcioca", "--hsm-caroot-id=1", "--ct-log-url=http://ct-log/test",
```
lukehinds, line 45: revert this back for PGI

PR #123, opened 2021-06-22
Implement standalone CLI command

### LINKED ISSUE #95: Usage outside of sigstore
The fulcio client code recently moved from this repo to the sigstore cli in the sigstore repo. I think fulcio is useful on its own as well, independently of rekor. The sigstore cli (correct me if I'm wrong) currently doesn't allow to just retrieve a fulcio cert without signing and sending to rekor. 

Is fulcio intended to stay as an independent service? If so, does it make sense to have a stand-alone client here (or elsewhere) for it? Or alternatively extend the sigstore cli to retrieve a cert for custom usage?

mbestavros:
May be used to generate an OIDC certificate without any signing or tlog
interaction. Resolves #95 

Open to feedback! I'm not sure if the output format is exactly what's needed, so testing would be appreciated.

Signed-off-by: Mark Bestavros <mbestavr@redhat.com>


lukehinds:
awesome! we might need to wait for https://github.com/sigstore/sigstore/pull/69 or https://github.com/sigstore/sigstore/pull/70 to shake out first, but changes needed here should be pretty simple!

letmaik:
I may be missing something but if the private key is generated in memory and discarded what would someone do with just the cert?

lukehinds:
> I may be missing something but if the private key is generated in memory and discarded, what would someone do with just the cert?

They would extract the public key (which was part of the ephemeral private key) and along with the signature, verify the signing event.

letmaik:
Sure, but you can't sign something if you don't have the private key. I would have expected the CLI to either dump the generated private key to a file or take an existing private key as input.

lukehinds:
> Sure, but you can't sign something if you don't have the private key. I would have expected the CLI to either dump the generated private key to a file or take an existing private key as input.

What use case do you have in mind here? Fulcio's scope is OIDC based signing signing certificates. It sounds to me like you more looking for a generic CA. Can you outline what sort of use you're hoping to get out of fulcio?

letmaik:
OIDC based signing certificates is exactly my use case. Maybe we can clear it up by adding a sample or test to this PR.

lukehinds:
> OIDC based signing certificates is exactly my use case. Maybe we can clear it up by adding a sample or test to this PR.

I prefer that we scope this out properly first in an issue. This is a CLI operation that is run on the server itself, not via a client API. Would you really want OIDC based signing certificates distributed to users who have make the request directly on the server itself? I guess there might be a valid use of this, but we need to consider security impacts on fulcio here. 

Could you create an issue please and outline the UX flow and what the expected end result would be?

letmaik:
I think https://github.com/sigstore/fulcio/issues/95#issuecomment-840642011 and following covers it.

lukehinds:
> I think [#95 (comment)](https://github.com/sigstore/fulcio/issues/95#issuecomment-840642011) and following covers it.

It does not. This really needs to be properly defined and not left up to interpretation (e.g. there was no mention of outputting private keys to disk).

dlorenc:
I think the other direction would make more sense. I don't want to generate keys here and write them to disk, but I would be ok with the user doing that then calling fulcio to get an certificate based on a public key.

The difference is subtle, but it's clear in the second case that the end user is the one responsible for managing the lifecycle of the key and protecting it.

cmd/client/app/root.go
```diff
@@ -0,0 +1,148 @@
...
+			return err
+		}
+
+		pub, err := signer.PublicKey(ctx)
+		if err != nil {
+			return err
+		}
+
+		pubBytes, err := x509.MarshalPKIXPublicKey(pub)
+
```
jeffvance, line 72: super nit: blank line should be deleted, imo.

cmd/client/app/root.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		cert, err := x509.ParseCertificate(certBlock.Bytes)
+		if err != nil {
+			return err
+		}
+
+		outputFileStr := viper.GetString("output")
+		outputFile := os.Stdout
+		if outputFileStr != "-" {
```
jeffvance, line 115: Is it possible for outputFileStr to be empty?

cmd/client/app/root.go
mbestavros: Good catch, I've tidied it up. Thanks!

cmd/client/app/root.go
mbestavros: `viper.GetString("output")` gets the command-line argument `output` from the app's invocation. That's all set up at the bottom of the file (lines 137-147). If the invocation didn't provide anything for the `output` flag, it'll default to `-` -- which is checked for on line 113. If that happens, the app won't attempt to write anything to a file.

mbestavros:
This has become out of date, and I filed the PR using my `main` branch, which caused problems on my end. I can fix up and re-file the PR if desired.

PR #145, opened 2021-07-14
fulcio-server: add html page when humans reach the server via the browser

cpanato:
When we access via browser the Fulcio server (for example) https://fulcio.sigstore.dev/ it returns 
```
{
  "code": 404,
  "message": "path / was not found"
}
```

maybe is good to add a webpage and say something :) instead of return a json 404.

This PR adds a very simple homepage when humans access that page

Screenshot 2021-07-14 at 17 32 01

if we don't want that and this is by design feel free to close this PR


openapi.yaml
```diff
@@ -19,14 +19,27 @@ info:
   title: Fulcio
 schemes: [http, https]
 host: fulcio.sigstore.dev
-basePath: /api/v1
+basePath: /
```
cpanato, line 22: @bobcallaway push back the basepath because I don't know if we can set a root Path and override the basepath, let me know if this is a good practice or not, thanks in advance

docker-compose.yml
```diff
@@ -24,7 +24,7 @@ services:
       "serve",
       "--host=0.0.0.0",
       "--port=5555",
-      "--ca googleca"
+      "--ca googleca",
```
bobcallaway, line 27: it also needs an equals sign in between ca and googleca

pkg/api/homepage.go
```diff
@@ -0,0 +1,53 @@
...
+//
+
+package api
+
+import (
+	"github.com/go-openapi/runtime/middleware"
+	"github.com/sigstore/fulcio/pkg/generated/restapi/operations"
+)
+
+const homePage = `<!DOCTYPE html>
```
bobcallaway, line 23: could we use https://pkg.go.dev/embed instead of inlining this here?

pkg/generated/restapi/configure_fulcio_server.go
```diff
@@ -105,12 +105,16 @@ func configureAPI(api *operations.FulcioServerAPI) http.Handler {
...
 
+	api.HomepageHandler = operations.HomepageHandlerFunc(pkgapi.HomePageHandler)
+
 	api.PreServerShutdown = func() {}
 
 	api.ServerShutdown = func() {}
 
 	api.AddMiddlewareFor("POST", "/api/v1/signingCert", middleware.NoCache)
 
+	api.AddMiddlewareFor("GET", "/", middleware.NoCache)
```
bobcallaway, line 116: instead of `middleware.NoCache`, how about just writing the bytes for the HTML here:
```
api.AddMiddlewareFor("GET", "/", func(w http.ResponseWriter, req *http.Request) {
    fmt.Fprintf(w, homepageBytes)
  })
```

bobcallaway (commented):
I wouldn't touch the openapi file at all for this; we can handle it by declaring the middleware handler directly for the "/" path.

docker-compose.yml
cpanato: yep 😅

pkg/generated/restapi/configure_fulcio_server.go
cpanato: thanks for your Review and time Bob, i will address the comments!

cpanato:
thanks @bobcallaway for your review and insights! really appreciate
PTAL

dlorenc (approved):

dlorenc:
LGTM now, I'll wait for Bob.

bobcallaway (approved):

PR #154, opened 2021-07-28
makefile: add rule to download and set swagger and make rule to build the dist

cpanato:
- Adding the changes we applied for rekor in this PR: https://github.com/sigstore/rekor/pull/391


bobcallaway:
looks fine, two thoughts:

1. should we update README.md (which currently has the developer manually install swagger)?
2. is there any benefit in having a "dummy" `go install` of swagger in a GitHub action (not actually used but at least referenced) so that @dependabot knows to flag updates for it?

cpanato:
@bobcallaway
1. already removed the reference in the readme that says to install swagger
2. the swagger update should not be our decision when need to upgrade? they might have any breaking changes? no strong opinions on that, I can take a look and see if dependabot can act in this case

thanks for the review

bobcallaway:
> @bobcallaway
> 
> 1. already removed the reference in the readme that says to install swagger

https://github.com/sigstore/fulcio/blob/485846c13d586e6e2d213ddb40d7ba05a993530a/README.md#build-for-development

> 2. the swagger update should not be our decision when need to upgrade? they might have any breaking changes? no strong opinions on that, I can take a look and see if dependabot can act in this case

They might, but since go-swagger is a big part of our code base I want to make sure we're at least notified when an upgrade is released so we can evaluate it. This can be a separate PR but just wanted to capture the idea

> 
> thanks for the review

README.md
```diff
@@ -34,7 +34,6 @@ We **WILL** change this and add intermediaries in the future.
 After cloning the repository:
 
 ```
-$ go get -u github.com/go-swagger/go-swagger/cmd/swagger@v0.26.1
```
cpanato, line 37: moreved here @bobcallaway

bobcallaway: I swear I looked for this but totally missed it... sorry @cpanato :(

bobcallaway (approved):

cpanato:
yep, will take a look in the dependabot review and see what we can do, otherwise we can make an action workflow to check that and notify us :)


> They might, but since go-swagger is a big part of our code base I want to make sure we're at least notified when an upgrade is released so we can evaluate it. This can be a separate PR but just wanted to capture the idea

cpanato:
after discussing with @bobcallaway i will update this PR to use something similar we have in https://github.com/kubernetes-sigs/cluster-api/blob/master/hack/tools/go.mod

bobcallaway (approved):
lgtm, thanks!

PR #170, opened 2021-08-25
go: upgrade golang to 1.17

cpanato:
- golangci-lint: bump to use release 1.42
- go: upgrade golang to 1.17

### DIFF
--- a/.github/workflows/verify-k8s.yml
+++ b/.github/workflows/verify-k8s.yml
@@ -23,9 +23,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2.3.4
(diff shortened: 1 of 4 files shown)


lukehinds (changes requested):
I honestly prefer we don't do this yet , unless there is a specific real need for 1.17.

A lot of Linux distributions will be a long way off supporting 1.17 (most are only just rolling out 1.16 on their latest releases). You are only going to find 1.17 on a rolling based distro. This means stable platforms (debian, opensuse, fedora (not including rawhide) won't be able to package fulcio.

cpanato:
got it. I will close the PRs

cpanato:
closing for now

PR #204, opened 2021-10-11
Add issuer information to code signing certificates

bobcallaway:
This extracts a claim from the OIDC token (defaults to `iss`) and inserts it into the final code signing certificate issued by Fulcio in an X509 extension. 

This provides an explicit link between the identity (i.e. the email or the URI that is the subject of the certificate) to the identity provider who is vouching for an entity holding valid credentials for said identity.

The value is stored in an X509 extension with OID:
1.3.6.1.4.1.57264.1.1

WIP due to https://github.com/dexidp/dex/issues/2289

Signed-off-by: Bob Callaway <bob.callaway@gmail.com>

### DIFF
--- a/config/fulcio-config.yaml
+++ b/config/fulcio-config.yaml
@@ -24,9 +24,9 @@ data:
             },
             "https://oauth2.sigstore.dev/auth": {
               "IssuerURL": "https://oauth2.sigstore.dev/auth",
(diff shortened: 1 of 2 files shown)


lukehinds (approved):
just needs a rebase, but good for me.

PR #228, opened 2021-10-30
Add support for recoginizing allow.pub as an spiffe issuer

evanphx:
#### Summary

Add support for fulcio to allow spiffe tokens signed by allow.pub

#### Description

I'm building an OCI registry that would like to perform server side
cosign support via fulcio. The OCI registry is called vcr.pub and should
be available in beta form soon.

The reason that the domain used here is allow.pub rather than vcr.pub is
one of the future services we'd like to support is other OIDC related
authentication tasks, so we've separated it from the OCI registry off
the bat.

### DIFF
--- a/federation/external/allow.pub/config.yaml
+++ b/federation/external/allow.pub/config.yaml
@@ -1,0 +1,18 @@
+# Copyright 2021 The Sigstore Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+url: https://allow.pub
+contact: evan@phx.io
+description: "Server side signing support for the OCI registry vcr.pub"
+type: "spiffe"
--- a/config/fulcio-config.yaml
+++ b/config/fulcio-config.yaml
@@ -22,6 +22,11 @@ data:
               "ClientID": "sigstore",
               "Type": "email"
             },
+            "https://allow.pub": {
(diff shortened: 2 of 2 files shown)


federation/external/allow.pub/config.yaml
```diff
@@ -0,0 +1,18 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+url: https://allow.pub
+contact: evan@phx.io
+description: "Server side signing support for the OCI registry vcr.pub"
```
mattmoor, line 17: Neither vcr.pub or allow.pub resolve to anything meaningful in the browser.  Perhaps include (comment?) a way to learn more?

evanphx: Certainly! The websites are absolutely blank atm because I'm just getting those services up and going.

vcr.pub is the focus here, it's a simple, credit based OCI registry. Credit based meaning it's a bit like tarsnap where you pay upfront for credits and usage deducts from there. The idea is to build the best OCI registry that I want to use, which includes image signing, thus my interest in sigstore/fulcio.

vcr.pub is in alpha atm but I'd like to get the automated signing going before promoting it more, since this sort of integration is a great feature to talk about.

More than happy to answer any questions about it or anything else!

evanphx: You make a great point, since I'm talking about these things outside my own tiny circle I should go ahead and at least provide something meaningful for browsers. I'll get that sorted out this weekend.

dlorenc:
Once you get the DCO fixed this should be good!

evanphx:
@dlorenc Ah! Ok, DCO fixed!

federation/external/allow.pub/config.yaml
evanphx: I've at least provided some redirects at https://allow.pub and https://vcr.pub for now.

dlorenc (approved):

dlorenc:
This should be live!

evanphx:
@dlorenc Thank you!

PR #244, opened 2021-11-23
plumb through !cgo golang tags that removes pkcs11 support

### LINKED ISSUE #240: Disable PKCS11 with `purego` build tag
**Description**

The need for `cgo` to build `PKCS11` support makes using an M1 macbook -> GKE kind of a pain because `CGO_ENABLED=1` doesn't work.  The go community has adopted the build tag `purego` to disable things requiring `cgo` in environments that only support `purego`, so I think we should plumb through logic to disable the PKCS11 support when `purego` is specified.

cc @dlorenc @vaikas @bobcallaway

vaikas:
#### Summary

If specifying golang env `CGO_ENABLED=0`, remove the code that uses PKCS11, because it depends on C. By default, nothing changes, this is just purely an option you must specify.

#### Ticket Link

Fixes #240 

#### Release Note

```release-note
 * Make it possible to build the binary in pure go mode by specifying `CGO_ENABLED=0`. Because pkcs11 support requires C support, when specifying this option, it does remove the support for `createca` command.
 * Fixes #240 
```

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -3,9 +3,7 @@ builds:
 - main: .
   env:
   - CGO_ENABLED=1
-# If you are deploying from M1, you can use this (uncomment this, and
-# comment above), though it does remove the support for the "createca" command.
+# If you are deploying from M1, you can use this (uncomment below, and
+# comment out above), though it does remove the support for the "createca" command.
 # But at least you can deploy it from M1 using this.
-#- main: .
-#  flags:
-#  - -tags=purego
+#  - CGO_ENABLED=0
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -3,3 +3,9 @@ builds:
 - main: .
   env:
   - CGO_ENABLED=1
+# If you are deploying from M1, you can use this (uncomment this, and
+# comment above), though it does remove the support for the "createca" command.
+# But at least you can deploy it from M1 using this.
+#- main: .
+#  flags:
+#  - -tags=purego
--- a/cmd/app/createca.go
+++ b/cmd/app/createca.go
@@ -1,3 +1,6 @@
+//go:build !purego
+// +build !purego
+
 // Copyright 2021 The Sigstore Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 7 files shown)


.ko.yaml
```diff
@@ -3,3 +3,9 @@ builds:
 - main: .
   env:
   - CGO_ENABLED=1
+# If you are deploying from M1, you can use this (uncomment this, and
+# comment above), though it does remove the support for the "createca" command.
+# But at least you can deploy it from M1 using this.
+#- main: .
+#  flags:
+#  - -tags=purego
```
mattmoor, line 11: This may be contentious, but my personal bias would be towards making the configs we check in by default maximally approachable by contributors.

I believe here that translates to opting INTO pkcs11 support here (like we do in cosign) vs. opting out (as this PR currently does).

Here's where cosign is doing this for `pivkey`: https://github.com/sigstore/cosign/blob/a064fabac961be65bee810369e8b34c66ef265bd/cmd/cosign/cli/piv_tool_disabled.go#L2

It looks like it is using a `cgo` tag, that I didn't come across in my search, which I wonder if it's set automatically 🤔

mattmoor: cc @dlorenc @lukehinds @bobcallaway

vaikas: Make up your mind, you want purego or cgo? :) 

A data point (maybe crappy, but a data point regardless), is that while I was struggling with why the config was not working (spoiler, I was using older version of ko), things didn't work. Then I updated ko and things started working (I reckon it started passing the tags flag to go build), so not sure it would work out of the box. But, in any case, I'm happy to change the tags to whatever we want them to be.

Also wanted to make sure I don't break any folks that are expecting the pkcs11 support by turning off something by this change, so went for the opt-out.

mattmoor: I just played a little locally and it seems that the `cgo` tag is automatically defined unless `CGO_ENABLED=0` is set, at which point the `purego` tag is set.

What didn't work with `ko` was likely the support for `goreleaser` config in the `.ko.yaml`, which is relatively new.  Previously you had to `export GOFLAGS=...`

mattmoor: I take that back.  Only `cgo` is automatically set, `purego` is not.  Let's use `cgo` instead?

lukehinds: If this keeps cgo by default and provides a toggle to implement purego as an option, I am good with that.

vaikas: I missed this when reading this the first time:
```
I believe here that translates to opting INTO pkcs11 support here (like we do in cosign) vs. opting out (as this PR currently does).
```
The PR as was (I'm changing the tag) does opt INTO pkcs11, because you had to uncomment the lines to get the non-pkcs11 support? @mattmoor did I misunderstand what you meant by this?

mattmoor (approved):
Let's land the support and then we can talk about what defaults make the most sense for `cgo`, no need to hold this up.

.ko.yaml
mattmoor: My bias was towards `CGO_ENABLED=0` by default, but I'd rather land this as-is, and change that later since it seems at least mildly contentious.

PR #248, opened 2021-11-24
fix: go install complain missing version when dir not in module

tuananh:
Signed-off-by: Tuan Anh Tran <me@tuananh.org>

#### Summary

go install `delve` will complain missing version when directory not in module (in Dockerfile) so we will include the `@latest` version when install.

```
go install: version is required when current directory is not in a module
```

#### Ticket Link

Fixes

#### Release Note

```release-note

```


cpanato (commented):
lgtm

bobcallaway (approved):

PR #218, opened 2021-10-27
Upgrade fulcios to use of the google privateca api at v1

n3wscott:
Signed-off-by: Scott Nichols <n3wscott@chainguard.dev>

#### Summary

We were still on a v1beta1 api for the googleca private ca. This adds a flag to use the v1 of the api. We are still needing to update the cert to `v1`.

The new flag `--gcp_private_ca_version` selects `v1` by default to allow us to drop the flag when we move to v1 certs and then delete the flag and the v1beta1 codepaths.

#### Ticket Link

Relates to a checkbox in https://github.com/sigstore/fulcio/issues/191

#### Release Note

```release-note
Fulcio now can use the v1 api for google privateca.
```

Issue #191, opened 2021-09-29
Release 1.0 planning


lukehinds:
The MVP is established now and other projects are leveraging fulcio, so its time to plan for 1.0

- [x] #194
- [x] #187
- [x] #195
- [x] #13 

cc @cpanato

cpanato:
same from rekor:

I would like to add a cloudbuild + goreleaser when we release Fulcio

then I have some questions:

1. We will use the same KMS key that we use to sign `cosign`? or we create another?
2. To push the data to the Rekor server how can we do that in an automated fashion way? there is any guide? or we just don't do it at this time?

@dlorenc @lukehinds @dekkagaijin

mattmoor:
The googlecs stuff needs to move to the 1.0 API

jdolitsky:
Should probably lock down the configuration prior to 1.0: https://github.com/sigstore/fulcio/issues/304

nsmith5:
Not 100% configuration change should be 1.0 blocker. If only because it seems like it might take some time to finish up a refactor. I discussed it a bit here: https://github.com/sigstore/fulcio/issues/304#issuecomment-1019345636

lukehinds:
I have been thinking the same. fuclio has been operating in a stable manner for a while now. I will freeze the milestone and we can look at closing the release.

cc @cpanato

cpanato:
if need to change any configuration, maybe we can do that now and make a pre-release with deprecations and then plan the 1.0 removing the deprecations.

lukehinds:
@cpanato as I understand, it won't be a change of flags or a non backwards compatible change. 

At the moment we have a pattern of folks adding to 1.0, but not able to do the work and so we are seeing 1.0 pushed back and a moving target. For me the most important role of fulcio is the public good instance and what we have at present, which although I am sure can be improved over time, is currently functioning well.

cpanato:
ahh got it, i think i misunderstood :(

haydentherapper:
Everything here has been completed. Work is tracked here - https://github.com/orgs/sigstore/projects/5/views/1

PR #252, opened 2021-12-01
[Correction] Upgrade fulcios to use of the google privateca api at v1

n3wscott:
Relands and corrects #218 

Signed-off-by: Scott Nichols <n3wscott@chainguard.dev>

#### Summary

We were still on a v1beta1 api for the googleca private ca. This adds a flag to use the v1 of the api. We are still needing to update the cert to `v1`.

The new flag `--gcp_private_ca_version` selects `v1` by default to allow us to drop the flag when we move to v1 certs and then delete the flag and the v1beta1 codepaths.

#### Ticket Link

Relates to a checkbox in https://github.com/sigstore/fulcio/issues/191

#### Release Note

```release-note
Fulcio now can use the v1 api for google privateca.
```


pkg/api/ca.go
```diff
@@ -53,7 +53,15 @@ func CA() certauth.CertificateAuthority {
 		var err error
 		switch viper.GetString("ca") {
 		case "googleca":
-			ca, err = googlecabeta.NewCertAuthorityService()
+			version := viper.GetString("gcp_private_ca_version")
+			switch version {
+			case "v1":
```
n3wscott, line 58: In the original pr, these where switches and no one caught it...

dlorenc: ooops!

dlorenc (approved):

PR #257, opened 2021-12-02
Remove `viper` from `pkg/`.

mattmoor:
This change builds on https://github.com/sigstore/fulcio/pull/255.  It moves the CA creating to the `serve` entrypoint, and attaches the appropriate `CertificateAuthority` to the request contexts.  The same for the `ct-log-url`.

With this change, all of the `viper` logic lives under `./cmd` and the logic in `./pkg` receives its configurations through parameters (possible attached to `ctx`).

Signed-off-by: Matt Moore <mattmoor@chainguard.dev>

WIP until the base change lands.

#### Ticket Link

N/A

#### Release Note
```release-note
NONE
```

### DIFF
--- a/cmd/app/serve.go
+++ b/cmd/app/serve.go
@@ -17,17 +17,23 @@ package app
 
 import (
 	"flag"
+	"fmt"
 	"net/http"
 
 	"github.com/go-openapi/loads"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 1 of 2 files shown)


mattmoor:
@bobcallaway I think the next thing I'm going to chase after `viper` is `sync.Once` singletons since I think we have a bunch of those that seem gratuitous 🤔

dlorenc (approved):

PR #262, opened 2021-12-02
[WIP] Experiment with dropping OpenAPI

mattmoor:
Signed-off-by: Matt Moore <mattmoor@chainguard.dev>

#### Release Note
```release-note

```


mattmoor:
This needs fairly careful testing and scrutiny before it goes in.  This drops a number of things on the floor, some of which we clearly need a replacement for including:
1. `pkg/client`: Used by `cosign`, so kinda important.  We probably don't need swagger for this, but we need something.
2. Discovery docs: I think the API used to serve this, but now it doesn't.  We should confirm if this is a problem, and probably add test coverage if we want it.

I like the whole "dropping 3500 LoC outside of vendor/" aspect of this change, though. 🤣

mattmoor:
Alright, I think that this is relatively complete, assuming we're cool with ditching the discovery stuff.  I staged the update to cosign to consume the new client library here: https://github.com/sigstore/cosign/pull/1126

I'm not sure what else might be relying on the existing fulcio client logic, so any feedback would be appreciated.

pkg/config/config.go
```diff
@@ -189,27 +207,29 @@ func FromContext(ctx context.Context) *FulcioConfig {
...
 // Load a config from disk, or use defaults
 func Load(configPath string) (*FulcioConfig, error) {
-	var config *FulcioConfig
 	if _, err := os.Stat(configPath); os.IsNotExist(err) {
 		log.Logger.Infof("No config at %s, using defaults: %v", configPath, DefaultConfig)
-		config = DefaultConfig
-		cache, err := lru.New2Q(100 /* size */)
-		if err != nil {
+		config := DefaultConfig
+		if err := config.prepare(); err != nil {
```
mattmoor, line 213: This is a bug at HEAD.  We don't set up the OIDCIssuers (just the lru).

pkg/challenges/challenges.go
```diff
@@ -270,14 +275,15 @@ func workflowInfoFromIDToken(token *oidc.IDToken) (map[AdditionalInfo]string, er
...
-
-	if spiffeDomain == host {
+	u, err := url.Parse(spiffeID)
+	if err != nil {
+		return false
+	}
+	if u.Scheme != "spiffe" {
+		return false
+	}
+	if u.Hostname() == host {
```
mattmoor, line 285: I changed this because writing my unit test (`api_test.go`) this doesn't properly handle port numbers at HEAD.

pkg/ca/x509ca/common.go
```diff
@@ -37,10 +37,6 @@ type X509CA struct {
 }
 
 func (x *X509CA) CreateCertificate(_ context.Context, subject *challenges.ChallengeResult) (*ca.CodeSigningCertificate, error) {
-	return x.CreateCertificateWithCA(x, subject)
```
mattmoor, line 40: IDK why this was like this, but it is unnecessary.

mattmoor: I put this into: https://github.com/sigstore/fulcio/pull/263

mattmoor:
TODO: incorporate: https://github.com/sigstore/fulcio/pull/264

pkg/api/client.go
```diff
@@ -0,0 +1,112 @@
...
+
+// Client is the interface for accessing the Fulcio API.
+type Client interface {
+	// SigningCert sends the provided CertificateRequest to the /api/v1/singingCert
+	// endpoint of a Fulcio API, authenticated with the provided bearer token.
+	SigningCert(cr CertificateRequest, token string) (*CertificateResponse, error)
+}
+
+// NewClient creates a new Fulcio API client talking to the provided URL.
+func NewClient(url *url.URL) Client {
```
bobcallaway, line 47: could be an additional patch, but would be nice to be able to override the client here (e.g. to set a specific timeout value)

pkg/api/client.go
```diff
@@ -0,0 +1,112 @@
...
+	"errors"
+	"io"
+	"net/http"
+	"net/url"
+	"path"
+)
+
+type CertificateResponse struct {
+	CertPEM  []byte
+	ChainPEM []byte
```
bobcallaway, line 32: worth parsing them into `x509.Certificate and []x509.Certificate` for increased usability?

openapi.yaml
```diff
@@ -1,116 +0,0 @@
-#
```
bobcallaway, line 1: do we need to delete this? I get the desire to remove go-swagger and the code-gen, but we've had several requests for API documentation.

cmd/app/serve.go
```diff
@@ -110,31 +100,41 @@ var serveCmd = &cobra.Command{
...
 		go func() {
-			_ = http.ListenAndServe(":2112", nil)
+			_ = prom.ListenAndServe()
 		}()
 
-		if err := server.Serve(); err != nil {
+		port := os.Getenv("PORT")
+		if port == "" {
+			port = "8080"
+		}
```
bobcallaway, line 130: My personal preference would be to leave the `--host` and `--port` command line args implemented.

bobcallaway (commented):
overall, seems like this is a functionally equivalent swap on a per-request basis so if your goal was to remove thousands of lines of generated code, I think you've achieved it :)

we'd also need to remove the dep on go-swagger from `hack/tools/` even though it isn't called.

pkg/api/client.go
mattmoor: Let's follow up with that, I opened: https://github.com/sigstore/fulcio/issues/266

openapi.yaml
mattmoor: I can bring it back, but we'd need to find a way to ensure it stays in sync with the code.  Any ideas?

pkg/api/client.go
mattmoor: Lemme dig into how this is used on the `cosign` side more, since this came from there originally.  Generally SGTM, but I just want to avoid a nightmare picking up this change.

mattmoor: Looks like these are passed around a fair amount.  First to `Signer` here: 

https://github.com/sigstore/cosign/blob/a47a835480c45d6db7a9751d7cb8bf87772799fc/cmd/cosign/cli/fulcio/fulcio.go#L185-L187

Next to `SignerVerifier` here:

https://github.com/sigstore/cosign/blob/7e5ff000f6af25793631eb590964b9a861d672dc/cmd/cosign/cli/sign/sign.go#L413-L415

Chasing the usages of the `SignerVerifier`'s `.Cert` it seems like the Rekor uploads want things like this, but only the policy code (to extract email/issuer) wants the structured form.


I'm mildly inclined to "Nack" on this, since it's a bit of plumbing and (given current usage) I'm not sure it helps usability a ton.  Maybe we can add an accessor to get the `x509.Certificate` later.  I can change if you feel strongly though, just LMK.

cmd/app/serve.go
mattmoor: I'll add this back now!

mattmoor: Ok these are back, and I've reverted the changes to `deployment.yaml`

bobcallaway (commented):
LGTM, I dont see any other functional issues. Just a couple minor things I saw but I think this is fine.

pkg/config/config.go
```diff
@@ -132,6 +132,24 @@ func (fc *FulcioConfig) GetVerifier(issuerURL string) (*oidc.IDTokenVerifier, bo
 	return verifier, true
 }
 
+func (fc *FulcioConfig) prepare() error {
+	fc.verifiers = make(map[string]*oidc.IDTokenVerifier, len(fc.OIDCIssuers))
+	for _, iss := range fc.OIDCIssuers {
```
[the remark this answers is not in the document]
mattmoor: It is possible now that they only configure meta issuers.  I think our KinD e2e tests have a leg that does that now.

mattmoor:
Fixed the two suggestions, thanks!

mattmoor:
I'll start dusting off my changes to import this into `sigstore/cosign`

dlorenc (approved):

bobcallaway (approved):

PR #267, opened 2021-12-08
Wrap the server with the Prometheus so we get metrics + add an e2e te…

vaikas:
Wrap the server with the Prometheus so we get metrics + add an e2e test for it. Note that
the default prometheus metrics (go_*) were already exposed.

I wrapped test in a job, since otherwise it will be trickier to get access to the
metrics from outside the kind cluster.

Fix the dangling pointer from #262 and fix a tiny typo, though I kind of do
like exposing a singingCert that might come handy (should be required?) in karaoke?

#### Ticket Link

Fixes

#### Release Note

```release-note
* Expose prometheus metrics for: fulcio_api_latency (histogram) and fulcio_new_certs (counter).
```

### DIFF
--- a/cmd/app/serve.go
+++ b/cmd/app/serve.go
@@ -100,19 +100,23 @@ var serveCmd = &cobra.Command{
 		}
 
 		decorateHandler := func(h http.Handler) http.Handler {
-			return http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
-				ctx := r.Context()
+			// Wrap the inner func with instrumentation to get latencies
+			// that get partitioned by 'code' and 'method'.
+			return promhttp.InstrumentHandlerDuration(
+				api.MetricLatency,
+				http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+					ctx := r.Context()
 
-				// For each request, infuse context with our snapshot of the FulcioConfig.
-				// TODO(mattmoor): Consider periodically (every minute?) refreshing the ConfigMap
-				// from disk, so that we don't need to cycle pods to pick up config updates.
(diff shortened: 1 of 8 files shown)


cpanato (commented):
lgtm

very cool

cmd/app/serve.go
```diff
@@ -100,19 +100,23 @@ var serveCmd = &cobra.Command{
...
-				// Alternately we could take advantage of Knative's configmap watcher.
-				ctx = config.With(ctx, cfg)
-				ctx = api.WithCA(ctx, baseca)
-				ctx = api.WithCTLogURL(ctx, viper.GetString("ct-log-url"))
-
-				h.ServeHTTP(rw, r.WithContext(ctx))
-			})
+			// Wrap the inner func with instrumentation to get latencies
+			// that get partitioned by 'code' and 'method'.
+			return promhttp.InstrumentHandlerDuration(
```
lukehinds, line 105: Will the decorate handler mean fulcio will not work without a Prometheus instance?What would happen to requests in that scenario?

lukehinds (changes requested):

cmd/app/serve.go
dlorenc: The nice part of the prometheus model is that it's pull based - exposing metrics with no instance is safe and a noop. A running collector would be needed to actually scrape these and do something useful with them.

dlorenc (approved):

cmd/app/serve.go
lukehinds: gotcha, good for me.

lukehinds (approved):

PR #271, opened 2021-12-10
Add verbose logging to fulcio client

haydentherapper:
Inspired by https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/transport/logger.go, this adds verbose logging to debug calls to fulcio. 

We redact the the OIDC token in the authorization header. I've chosen to not redact the body of the request and response, because without the OIDC token or private key, the contents of the body can't be used (the public key, challenge, and returned certificates).

By default, the logs are off. I'll send another PR to cosign to turn on verbose logging.

#### Release Note
```release-note
Added optional verbose client logs
```

### DIFF
--- a/pkg/api/client.go
+++ b/pkg/api/client.go
@@ -21,12 +21,30 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"errors"
+	"fmt"
 	"io"
+	"io/ioutil"
+	"log"
 	"net/http"
+	"net/http/httputil"
 	"net/url"
 	"path"
+	"strings"
 	"time"
 )
+
+var (
+	// DebugLog is used to log information that is useful for debugging.
+	// Enable with: api.DebugLog.SetOutput(os.Stderr)
+	DebugLog = log.New(ioutil.Discard, "", log.LstdFlags)
+)
+
+// Enabled checks to see if the logger's writer is set to something other
+// than ioutil.Discard. This allows callers to avoid expensive operations
+// that would end up in /dev/null.
+func Enabled(l *log.Logger) bool {
+	return l.Writer() != ioutil.Discard
+}
 
 type CertificateResponse struct {
 	CertPEM  []byte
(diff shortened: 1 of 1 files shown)


pkg/api/client.go
```diff
@@ -149,7 +167,57 @@ type roundTripper struct {
...
+	DebugLog.Printf("--> %s %s", req.Method, req.URL)
+	// Save these headers so we can redact Authorization.
+	savedHeaders := req.Header.Clone()
+	if req.Header != nil && req.Header.Get("authorization") != "" {
+		req.Header.Set("authorization", "<redacted>")
+	}
+
+	b, err := httputil.DumpRequestOut(req, true)
+	if err == nil {
+		DebugLog.Println(string(b))
```
dlorenc, line 187: Looks like codeql thinks this contains sensitive info even though you redact right above - is there a way to suppress the codeql warning?

haydentherapper: I've redacted the body of the request too now. This makes the logging less useful, so maybe we can find some middle ground, but I want to see if that's the issue.

Would you be able to kick off the CI, I'm not able to since I haven't committed to this repo before.

dlorenc: Kicked it, still seems upset about the URL...

haydentherapper: This error comes from c# - https://codeql.github.com/codeql-query-help/csharp/cs-log-forging/

It looks like we're running the go version of codeql though, so not sure why we're seeing this

dlorenc:
Any ideas how to suppress this?

haydentherapper:
Let's see if that helps. I changed `req.URL` to `req.URL.String()`, and made a copy of the headers to redact.

haydentherapper:
Well that didn't work...
It seems like the analysis is just too sensitive to variable use. We can only disable it per-project, not per-line. I'll see if I can find another approach, if not, I'll close this for now.

haydentherapper:
I ran codeql locally and cannot replicate these failures. I'm likely just going to drop this PR.

PR #287, opened 2021-12-20
Add a Root Cert method to the CA interface, and implement it.

dlorenc:
This will make the ephemeral and HSM-based CAs easier to use, in addition
to giving us an API to retrieve the existing root cert. This isn't available
in-band today.

Signed-off-by: Dan Lorenc <lorenc.d@gmail.com>

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/stretchr/testify v1.7.0
 	go.uber.org/multierr v1.7.0 // indirect
 	go.uber.org/zap v1.19.1
+	google.golang.org/api v0.63.0
 	google.golang.org/genproto v0.0.0-20211208223120-3a66f561d7aa
 	google.golang.org/protobuf v1.27.1
 	gopkg.in/square/go-jose.v2 v2.6.0
--- a/pkg/api/ca.go
+++ b/pkg/api/ca.go
@@ -48,12 +48,16 @@ type CertificateRequest struct {
 	SignedEmailAddress []byte `json:"signedEmailAddress"`
 }
 
-const signingCertPath = "/api/v1/signingCert"
+const (
+	signingCertPath = "/api/v1/signingCert"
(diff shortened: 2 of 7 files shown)


dlorenc:
I don't really know if this is the best idea, but right now the ephemeral certs and the HSM stuff can be tricky to use if you don't remember the cert you loaded in. In the case of the GCP pools, the root certs can actually get created dynamically as well, with no way for anyone to know which are used.

This is particularly important when automating setup with the CT log - where you need to configure the CT log with this set of certs.

I'm open to other ideas here, just wanted to try this out as a prototype.

haydentherapper:
I like this idea. CA Service implements something similar, [FetchCaCerts](https://cloud.google.com/certificate-authority-service/docs/reference/rest/v1/projects.locations.caPools/fetchCaCerts).

What do you think of changing the method from `rootCert` to `caCertChain` or something that returns a list of certificates, so when there's more support for intermediates, it won't require a new API method?

dlorenc:
> What do you think of changing the method from `rootCert` to `caCertChain` or something that returns a list of certificates, so when there's more support for intermediates, it won't require a new API method?

I thought about that - but would that work? The chain could vary per reach issued leaf, and we have a spot for that in the API already I think.

Would it be something like "return the currently active chain"?

haydentherapper:
Yea, the currently active chain. For previously issued certificates, like you said, you'd be able to verify as long as you've persisted the chain used to issue that cert. This would just be for clients that want the latest chain.

dlorenc:
Renamed so we can return them all!

vaikas:
Fantastic! LGTM

mattmoor (approved):
We should add this to the `client` and `api_test.go` as well.

haydentherapper:
@dlorenc You mentioned you renamed the method, it doesn't look like that change got included?

PR #288, opened 2021-12-21
Fix the SCT header return value from the API to base64 encode it.

dlorenc:
I don't actually know what's correct here - but the current client expects
it to be base64 encoded. We could keep it "raw", but then we have to fix the client
that's currently in cosign.

Signed-off-by: Dan Lorenc <lorenc.d@gmail.com>

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/pkg/api/ca.go
+++ b/pkg/api/ca.go
@@ -203,7 +203,7 @@ func signingCert(w http.ResponseWriter, req *http.Request) {
 	}
 
 	// Set the SCT and Content-Type headers, and then respond with a 201 Created.
-	w.Header().Add("SCT", string(sctBytes))
+	w.Header().Add("SCT", base64.StdEncoding.EncodeToString(sctBytes))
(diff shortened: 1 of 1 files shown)


pkg/api/ca.go
```diff
@@ -203,7 +203,7 @@ func signingCert(w http.ResponseWriter, req *http.Request) {
 	}
 
 	// Set the SCT and Content-Type headers, and then respond with a 201 Created.
-	w.Header().Add("SCT", string(sctBytes))
+	w.Header().Add("SCT", base64.RawStdEncoding.EncodeToString(sctBytes))
```
vaikas, line 206: Since the client uses StdEncoding when decoding, should this use that too?

dlorenc: Doh, changed.

vaikas (approved):

vaikas:
I'm also adding a test for this as part of #289

bobcallaway (approved):

PR #290, opened 2021-12-21
Add RootCert method to client + tests

### LINKED ISSUE #289: Add the new RootCert this to the `client` and `api_test.go` as well.
We should add this to the `client` and `api_test.go` as well.

_Originally posted by @mattmoor in https://github.com/sigstore/fulcio/pull/287#pullrequestreview-836949571_

vaikas:
* Add tests for exercising STL
* Fix #289
* Validates that the new STL tests failed unless backporting #288

#### Ticket Link

Fixes

#### Release Note

```release-note
API Additions: Add RootCert method into the pkg/api/client
```

### DIFF
--- a/pkg/api/client.go
+++ b/pkg/api/client.go
@@ -34,6 +34,10 @@ type CertificateResponse struct {
 	SCT      []byte
 }
 
+type RootResponse struct {
+	ChainPEM []byte
+}
+
 // SigstorePublicServerURL is the URL of Sigstore's public Fulcio service.
 const SigstorePublicServerURL = "https://fulcio.sigstore.dev"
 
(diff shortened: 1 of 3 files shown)


pkg/api/client.go
```diff
@@ -112,6 +118,9 @@ func (c *client) SigningCert(cr CertificateRequest, token string) (*CertificateR
 
 	// Split the cert and the chain
 	certBlock, chainPem := pem.Decode(body)
+	if certBlock == nil {
```
vaikas, line 121: FYI, since Decode returns nil for certBlock if it can't find a PEM, I added this check since it certainly seemed like it must return a cert or it should be considered a failure?

dlorenc (approved):

PR #297, opened 2021-12-22
Change ports for docker compose to avoid conflict with Rekor

haydentherapper:
I've been working through issues with docker-compose. This is one of the simpler ones to resolve, which is changing the ports so that we can run both Rekor and Fulcio locally at the same time.

Other issues are plumbing through the GCP CA Service CA resource path without having to change the yaml, and there is an issue spinning up a local Dex service.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -77,7 +77,7 @@ services:
       - ctfeConfig:/etc/config/:ro
     command: [
         "--log_config" ,"/etc/config/ct_server.cfg",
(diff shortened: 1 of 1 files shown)


dlorenc:
This looks good! I don't know enough about docker compose to know off the top of my head, but would it be possible to share the same trillian instance between the two somehow? Something like a "shared service" that only gets spun up once?

dlorenc (approved):

PR #306, opened 2022-01-03
Extract additional claims from github-workflow token

### LINKED ISSUE #305: `job_workflow_ref` is insufficient to disambiguate actions workflows
**Description**

Github launched the ability to call other workflows like a step: https://github.blog/2021-11-29-github-actions-reusable-workflows-is-generally-available/  In these cases the launched job shows up as the `job_workflow_ref`, and NOT the workflow definition that lives in the user's repository.

Prior to this feature, the `job_workflow_ref` effectively codified the `repo`, `workflow`, and `ref` attributes into a single string, but with this feature all of that context is now lost!

I believe that the `job_workflow_ref` is still the correct identity for us to encode (it is what's doing the work), but we should also include (the no longer redundant) metadata about the originating workflow.

We already have a handful of Github extensions here: https://github.com/sigstore/fulcio/blob/f4746cc4ff3dc5f12b325b0c7a28af1217fa8194/pkg/challenges/challenges.go#L254-L270

I'd propose we also add: `repo`, `workflow`, and `ref` to enable folks to disambiguate callers.

ckotzbauer:
With reusable github-workflows the "job_workflow_ref" will reference the shared workflow instead the actual calling workflow.

#### Summary
This adds the additional infos `repository`, `workflow` and `ref` from the Github-Workflow OIDC token as described in #305. According to the [Github-Docs](https://docs.github.com/en/enterprise-cloud@latest/actions/deployment/security-hardening-your-deployments/using-openid-connect-with-reusable-workflows#how-the-token-works-with-reusable-workflows) I added the `repository` field instead of the `repo`.

#### Ticket Link
Fixes #305

#### Release Note
```release-note
job_workflow_ref info was insufficient when using with shared github-workflows
```

### DIFF
--- a/OID_INFO.md
+++ b/OID_INFO.md
@@ -15,4 +15,9 @@ Note that all values begin from the root OID 1.3.6.1.4.1.57264 [registered by Da
     - This contains the `event_name` claim from the GitHub OIDC Identity token that contains the name of the event that triggered the workflow run. [(docs)](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#understanding-the-oidc-token)
 - *1.3.6.1.4.1.57264.1.3*: (GithubWorkflowSha)
     - This contains the `sha` claim from the GitHub OIDC Identity token that contains the commit SHA that the workflow run was based upon. [(docs)](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#understanding-the-oidc-token)
-
+- *1.3.6.1.4.1.57264.1.4*: (GithubWorkflowName)
+    - This contains the `workflow` claim from the GitHub OIDC Identity token that contains the name of the executed workflow. [(docs)](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#understanding-the-oidc-token)
+- *1.3.6.1.4.1.57264.1.5*: (GithubWorkflowRepository)
+    - This contains the `repository` claim from the GitHub OIDC Identity token that contains the repository that the workflow run was based upon. [(docs)](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#understanding-the-oidc-token)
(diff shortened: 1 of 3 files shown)


nsmith5:
I believe this additional information also needs to be embedded into the x509 code signing certificate to fix #305. You can see [here](https://github.com/sigstore/fulcio/blob/main/pkg/ca/x509ca/common.go#L102-L120) where the trigger and SHA are added. I believe you'd need to create new extensions for this new data and additionally document them [here](https://github.com/sigstore/fulcio/blob/main/OID_INFO.md) like the existing ones.

ckotzbauer:
Thanks for your comment @nsmith5, I will add the extensions and the docs.

ckotzbauer:
@nsmith5 I added the new claims to the signing-cert and the docs now.

nsmith5 (approved):
Nice, great work! I'm not a maintainer so can't approve the workflow / merge, but this looks good to me

nsmith5:
cc @mattmoor for 👀

mattmoor (approved):

mattmoor:
cc @dlorenc to merge/deploy once CI passes.

bobcallaway:
The code looks good to me; one thought though: should we consider using a separate child OID for GitHub-specific fields rather than keeping them all at the same depth? If we add information from other CI/CD systems it might make sense to separate them, but I'm curious to hear what others think.

dlorenc:
> The code looks good to me; one thought though: should we consider using a separate child OID for GitHub-specific fields rather than keeping them all at the same depth? If we add information from other CI/CD systems it might make sense to separate them, but I'm curious to hear what others think.

I'd be fine with that. I think we can nest them arbitrarily long.

ckotzbauer:
@dlorenc @bobcallaway Should the nesting be part of this PR? If yes, how is it supposed to look like?

dlorenc:
I'm fine either way since we already have some un-nested Github OIDs. We could start nesting other providers going forward.

Wdyt @bobcallaway?

bobcallaway:
this PR is fine as is, we can nest them in a separate one.

PR #308, opened 2022-01-03
Remove hack/tools

nsmith5:
I believe this directory doesn't serve a purpose now that we don't use go-swagger / openapi anymore.


Makefile
```diff
@@ -19,7 +19,6 @@ all: fulcio
 # Ensure Make is run with bash shell as some syntax below is bash-specific
 SHELL:=/usr/bin/env bash
 
-OPENAPIDEPS = openapi.yaml
 SRCS = $(shell find cmd -iname "*.go") $(shell find pkg -iname "*.go")
 TOOLS_DIR := hack/tools
 TOOLS_BIN_DIR := $(abspath $(TOOLS_DIR)/bin)
```
bobcallaway, line 24: is there a reason not to delete these references to the directory too? same for line 62 and lines 74-81?

nsmith5: Ah! Nice catch. No reason. Just didn't notice that reference

bobcallaway (approved):
lgtm, thanks!

cpanato (approved):

PR #309, opened 2022-01-04
Switch to use fileca in e2e tests

### LINKED ISSUE #301: Replace `ephemeralca` in cosign e2e tests with `fileca`
**Description**

In the `fileca` PR we dicussed using `fileca` instead of `emphemeralca` in e2e tests.

jdolitsky:
Resolves #301


jdolitsky:
@mattmoor @vaikas - is it more involved than this? Need to point to a CA file?

mattmoor:
cc @nsmith5 

Yeah, we need to set up the secret, mount it, and pass the flag to Fulcio.

.github/workflows/verify-k8s.yml
```diff
@@ -136,8 +136,8 @@ jobs:
           # Reduce the resource requests of Fulcio
           sed -i -e 's,memory: "1G",memory: "100m",g' ${{ github.workspace }}/config/deployment.yaml
           sed -i -e 's,cpu: ".5",cpu: "50m",g' ${{ github.workspace }}/config/deployment.yaml
-          # Switch to the ephemeralca for testing.
-          sed -i -e 's,--ca=googleca,--ca=ephemeralca,g' ${{ github.workspace }}/config/deployment.yaml
+          # Switch to fileca for testing.
+          sed -i -e 's,--ca=googleca,--ca=fileca,g' ${{ github.workspace }}/config/deployment.yaml
```
nsmith5, line 140: Once the secret is set up and mounted, the `fileca` needs a couple more flags to just figure out where the certs are mounted and the password to decrypt the key. They're as follows

- `--fileca-cert` for the cert path
- `--fileca-key` for the key path
- `--fileca-passwd` for the key password
- `--fileca-watch=true/false` if you want to watch the cert and key path for updates and reload on change

I'm less familiar with how the deployment is configured, but these flags will need to be set somewhere to get fileca working

jdolitsky:
Current short-term plan here (please let me know if any objections):

1. Add a simple main.go for generating a new CA cert+key (based on `ephemeralca` logic)
2. Run main.go mentioned in previous step to generate a temporary CA cert+key
3. Create new `fulcio-secret.yaml` in config/ on the fly, adding the contents of CA cert+key
4. Modify config/deployment.yaml to mount new secret (optionally)
5. Fix sed to pass in `--fileca-cert` and `--fileca-key`
6. Extra: Remove sed for replica count, keep 3 replicas (since distributed should work properly now using `fileca`)

nsmith5:
Maybe we could skip 1. and just use openssl to generate the key pair? Would save us from maintaining that extra bit of go code. There are lots of examples of generating key-pairs using openssl in `pkg/ca/fileca/testdata/generate.sh`. E.g 

```bash
password=password123
duration=36500 # 100 years

# ed25519
openssl req -x509 \
    -newkey ed25519 \
    -sha256 \
    -keyout ed25519-key.pem \
    -out ed25519-cert.pem \
    -subj "/CN=ed25519" \
    -days $duration \
    -addext basicConstraints=critical,CA:TRUE,pathlen:1 \
    -passout pass:"$password"
```

jdolitsky:
@nsmith5 that sounds good, thanks for sharing paths to example scripts

jdolitsky:
Regarding my comment "Remove sed for replica count, keep 3 replicas", going to skip this for now as it appears this was done for prometheus vs. CA-related

.github/workflows/verify-k8s.yml
```diff
@@ -176,6 +190,21 @@ jobs:
...
+          # Create secret needed to use fileca
+          cat <<EOF > config/fulcio-secret.yaml
+          apiVersion: v1
+          kind: Secret
+          metadata:
+            name: fulcio-config
+            namespace: fulcio-dev
+          data:
+            key.pem: $(cat "${{ github.run_id }}-key.pem" | base64)
+            cert.pem: $(cat "${{ github.run_id }}-cert.pem" | base64)
```
mattmoor, line 202: FYI you can use `stringData` and drop the `base64` (it's k8s magic ✨ )

.github/workflows/verify-k8s.yml
```diff
@@ -131,13 +131,27 @@ jobs:
...
       - name: Deploy fulcio-dev
         run: |
           # Reduce the resource requests of Fulcio
           sed -i -e 's,memory: "1G",memory: "100m",g' ${{ github.workspace }}/config/deployment.yaml
           sed -i -e 's,cpu: ".5",cpu: "50m",g' ${{ github.workspace }}/config/deployment.yaml
-          # Switch to the ephemeralca for testing.
-          sed -i -e 's,--ca=googleca,--ca=ephemeralca,g' ${{ github.workspace }}/config/deployment.yaml
+          # Switch to fileca for testing.
+          sed -i -e \
+            's,--ca=googleca,--ca=fileca --fileca-key=/etc/fulcio-secret/key.pem --fileca-cert=/etc/fulcio-secret/cert.pem --fileca-passwd=${{ github.run_id }},g' \
```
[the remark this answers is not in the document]
jdolitsky: resolved by using config file, see #314

.github/workflows/verify-k8s.yml
jdolitsky: hm. is that going to work alright with a multiline file?

mattmoor: It would, but something something yaml 😞

jdolitsky:
Pulled in changes from #314 and tests passing now

config/deployment.yaml
```diff
@@ -52,6 +52,9 @@ spec:
         volumeMounts:
         - name: fulcio-config
           mountPath: /etc/fulcio-config
+        - name: fulcio-secret
+          mountPath: /etc/fulcio-secret
+          readOnly: true
```
mattmoor, line 57: I believe this will cause production to break, but I believe you can make it optional (and then it won't)

.github/workflows/verify-k8s.yml
```diff
@@ -173,12 +185,27 @@ jobs:
             server.yaml: |-
               host: 0.0.0.0
               port: 5555
-              ca: ephemeralca
+              ca: fileca
+              fileca-cert: /etc/fulcio-secret/cert.pem
+              fileca-key: /etc/fulcio-secret/key.pem
+              fileca-key-passwd: "${{ github.run_id }}"
               gcp_private_ca_version: v1
```
mattmoor, line 192: may as well drop this

config/deployment.yaml
```diff
@@ -62,6 +65,10 @@ spec:
       - name: fulcio-config
         configMap:
           name: fulcio-config
+      - name: fulcio-secret
+        secret:
+          secretName: fulcio-secret
+          optional: true
```
jdolitsky, line 71: ^^ @mattmoor

config/deployment.yaml
jdolitsky: Does the `optional` field added in `volumes` section below accomplish this?

jdolitsky: I think so. Tried adding optional in another branch and got yelled at:
```
error: error validating "STDIN": error validating data: ValidationError(Deployment.spec.template.spec.containers[0].volumeMounts[1]): unknown field "optional" in
```

config/deployment.yaml
vaikas: That should make it so that it doesn't block starting it. The one interesting bit is what happens if you do want to specify the secret and it is not there yet but you'd like it to actually block. IIRC though, we watch that mount point so I _think_ it will pick it up once it becomes available. @nsmith5 check me out :)

jdolitsky: Looks like `--fileca-watch` flag defaults to true, so I think thats correct @vaikas

nsmith5: `fileca` will never start without a valid key-pair. We never want to boot up the server if we can't serve requests so I wrote it to try and load an initial key-pair and bail hard if the doesn't work (ref https://github.com/sigstore/fulcio/blob/main/pkg/ca/fileca/fileca.go#L48)

I don't think I know what Kubernetes would do in this case. My hope would be a pod crashloop until the secret mounts, but I don't know the `optional` spec well enough to know that is the case. Once the pod from this deployment is scheduled and the secret wasn't available at schedule time, does that mean it will never get that secret volume mounted by pods scheduled later when the secret exists _will_ get it? Or does it mean that Kubernetes will try to mount it even to an existing pod once the secret does exist?

nsmith5: The watch behaviour is also pretty paranoid about putting the server in a state where it can't serve requests so it looks for updates, but if, for instance, the cert doesn't match the key or the key can't be decrypted with the password it will just keep using the existing certs and ignore the updates from the filesystem

vaikas: I think the real issue here is that we should not require the pod to crashloop if the secret is missing, unless the `fileca` is being specified, because in those cases, there's no spoon (err, secret).
That was the thinking behind having the secret be optionally mounted, and then depending on the flags specified, we'd do something different.
When secrets change after they've been mounted, they will eventually be updated on the volume mount. But the question about if theres' no secret at startup time and then it shows up if that will still hold true is a good one, and for that I do not know the answer. I'd expect that an empty volume is mounted and when the secret shows up it will be populated, but that's just my expectation as a user.

vaikas: Oops, I read your reply incorrectly :) Yes, if the secret is missing, the pod will crashloop because there's no secret. In my head I read that the pod won't start. So, yes.
But the question is still if non-existent secret at startup time will still get updated once it shows up. And I do not know answer to that.

jdolitsky: So the default config is using `googleca` - so is this really any concern? If the files are missing using googleca, then .. no issue right?

jdolitsky: see https://github.com/sigstore/fulcio/blob/main/config/fulcio-config.yaml#L61

mattmoor: > The one interesting bit is what happens if you do want to specify the secret and it is not there yet but you'd like it to actually block

I think what we need to resolve this via Dr. Empirical is:
1. Create a Pod with an optional secret volume and an image that crash loops if the secret isn't there.
2. Create the secret and see if it stops crash looping.

I don't believe K8s will mount an optional secret into a running pod, but it is plausible that it might when restarting containers.  An interesting corollary here is: Does K8s refresh secrets projected into env vars when a container crashes?

vaikas: We can totally do an experiment, my only thought is that if the behaviour changes from version to version if it's not explicitly specified. But mosdef worth testing.

jdolitsky: Ran Matt's suggestion, and things appear to work as we expect:

1. Create a Pod with an optional secret volume and an image that crash loops if the secret isn't there:

```
cat << EOF | kubectl apply -f -
apiVersion: apps/v1
kind: Deployment
metadata:
  name: boop-deployment
  labels:
    app: boop
spec:
  replicas: 1
  selector:
    matchLabels:
      app: boop
  template:
    metadata:
      labels:
        app: boop
    spec:
      containers:
      - name: boop
        image: bash
        command:
        - "bash"
        - "-c"
        - "[[ -f /boop/bop.txt ]] || (echo bad && exit 1) && (echo good && while true; do sleep 1; done)"
        volumeMounts:
        - name: boop-secret
          mountPath: /boop
          readOnly: true
      volumes:
      - name: boop-secret
        secret:
          secretName: boop-secret
          optional: true
EOF
```

```
$ k get pods
NAME                               READY   STATUS             RESTARTS   AGE
boop-deployment-5c48fd856b-tv9lz   0/1     CrashLoopBackOff   2          33s
```

2. Create the secret and see if it stops crash looping.

```
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Secret
metadata:
  name: boop-secret
data:
  bop.txt: $(echo boop | base64)
EOF
```

```
$ k get pods
NAME                               READY   STATUS    RESTARTS   AGE
boop-deployment-5c48fd856b-tv9lz   1/1     Running   3          49s
```

yay?

mattmoor: @vaikas @nsmith5 Did you have anything else on this before we merge?

nsmith5: Nope looks good from me! Good work @jdolitsky :D

nsmith5: Thanks for checking in with Dr. Empirical 😆

vaikas: No. I think it might behoove us to file an issue in k8s to document the behaviour in case they reserve the right to change this behaviour. But I'm happy with the change for sure.

mattmoor (approved):

PR #315, opened 2022-01-05
Enable server settings via config file and env vars

### LINKED ISSUE #314: Allow passing configuration via config file
Make use of Viper's `viper.AddConfigPath` to allow configuring fulcio using a YAML config file

Example: https://github.com/spf13/viper#reading-config-files

Discovered trying to modify sed statements in #309

cc @mattmoor

jdolitsky:
Modify the viper settings in the serve subcommand to allow users to provide a config file using -c/--config vs. setting everything using command line flags.

Sinc this is based upon viper, the following config file extensions are supported: "json", "toml", "yaml", "yml", "properties", "props", "prop", "hcl", "tfvars", "dotenv", "env", and "ini".

This change also allows using env vars prefixed with FULCIO_SERVE to provide configuration settings as well.

Resolves #314

### DIFF
--- a/.github/workflows/verify-k8s.yml
+++ b/.github/workflows/verify-k8s.yml
@@ -136,10 +136,6 @@ jobs:
           # Reduce the resource requests of Fulcio
           sed -i -e 's,memory: "1G",memory: "100m",g' ${{ github.workspace }}/config/deployment.yaml
           sed -i -e 's,cpu: ".5",cpu: "50m",g' ${{ github.workspace }}/config/deployment.yaml
-          # Switch to the ephemeralca for testing.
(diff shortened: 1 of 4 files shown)


.github/workflows/verify-k8s.yml
```diff
@@ -170,10 +170,17 @@ jobs:
             name: fulcio-config
             namespace: fulcio-dev
           data:
+            server.yaml: |-
+              host: 0.0.0.0
+              port: 5555
+              ca: googleca
```
mattmoor, line 176: We should change this to `ephemeralca` and start removing the `sed` below?  In fact I guess this will fail.

jdolitsky: ah right

mattmoor (approved):

PR #316, opened 2022-01-05
Fix docker-compose dexidp startup

haydentherapper:
There was some change in the last year to use a non-root
user for the Dex image. This causes permission issues
when mounting files, since the non-root user cannot
access the mount. The workaround is to run as root,
which should be fine for development.

See https://github.com/dexidp/dex/issues/1649

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -47,6 +47,7 @@ services:
       - ct_server
   dex-idp:
     image: dexidp/dex:v2.30.0
+    user: root
     command: [
       "dex",
       "serve",


docker-compose.yml
```diff
@@ -47,6 +47,7 @@ services:
       - ct_server
   dex-idp:
     image: dexidp/dex:v2.30.0
+    user: root
```
cpanato, line 50: what if we change the location of the config file to a tmp directory?

haydentherapper: I tried this out too, I get permission issues with any directory. For example, if I mount ./config/dex to /tmp, the default user can't access /tmp. When I shell into the image without mounting a file, I'm able to access /tmp though, so I assume this is from the docker process that mounts the directory overriding the directory permissions.

haydentherapper:
@mattmoor Could we get this merged?

bobcallaway (approved):
nothing against the fix, just wish it wasn't required...

mattmoor:
I'm also curious if we can find a way to not need this (do we in prod?), but I'm unfamiliar with docker compose to be of any real help.

haydentherapper:
In prod, we load the config in directly, i.e 

```
data:
  config.yaml: |
    issuer: https://oauth2.sigstore.dev/auth
    ...
```

I'll look at making that change.

Issue #302, opened 2021-12-29
Add usage / docs for `fileca`


nsmith5:
**Description**

We need a basic example of using `fileca` backing in our docs for operators. Should show how to create a CA cert with openssl and then use is in fulcio.

nsmith5:
As pointed out by @haydentherapper, with the support of intermediate CA in `fileca` we need to document the expected order of certificates in the cert bundle.

NB: if #320 is merged that order will be signing intermedate first -> root last

PR #320, opened 2022-01-08
Support intermediate CA with `fileca` backend

### LINKED ISSUE #300: Support CA chain in `fileca`
**Description**

The current `fileca` implementation only supports a root CA certificate, but should support a chain of intermediates of arbitrary length

nsmith5:
#### Summary

Adds support for an intermediate CA using the fileca backend

#### Ticket Link

Fixes #300

#### Release Note

```release-note
* Added support for intermediate CA to the `fileca` backend
```


pkg/ca/fileca/load.go
```diff
@@ -27,21 +27,34 @@ import (
...
-		key  crypto.Signer
+		certs []*x509.Certificate
+		err   error
+		key   crypto.Signer
 	)
 
-	// TODO: Load chain of certs (intermediates and root) instead of just one
-	// certificate.
-	cert, err = pemutil.ReadCertificate(certPath)
+	certs, err = pemutil.ReadCertificateBundle(certPath)
```
haydentherapper, line 38: It's worth documenting the expected order of certificates. From ReadCertBundle, this would be intermediate first, root last.

pkg/ca/fileca/load.go
```diff
@@ -27,21 +27,34 @@ import (
...
 	if err != nil {
 		return nil, nil, err
 	}
 
+	// Verify certificate chain
+	{
```
haydentherapper, line 44: This block of code treats all certificates in the bundle as root certificates, trying to build a valid chain from certs[0] to any other cert in the root cert pool.

Assuming the order of certs is intermediate1 (that will be used to issue leaf certs), i2, i3...root, you'll want to create a second cert pool for intermediates and use the Intermediates option from [VerifyOptions](https://pkg.go.dev/crypto/x509#VerifyOptions). So something like: 
* Add certs[last index] to root pool
* Add certs[1 .. last index - 1] to intermediate pool
* Verify certs[0]

pkg/ca/fileca/testdata/intermediate-cert.pem
```diff
@@ -0,0 +1,18 @@
+-----BEGIN CERTIFICATE-----
```
haydentherapper, line 1: Could we also add test data for a 3-long chain, with two intermediates and a root? This should exercise the code that validates a pool of roots and a pool of intermediates.

nsmith5:
Thanks for the timely review @haydentherapper! I'll add the len(chain) = 3 test case and modify the cert verification logic soon :D

pkg/ca/fileca/load.go
haydentherapper: One other small thing, VerifyOptions will need to include the extended key usages that are used in all certificates. 

```
&VerifyOptions {
  ...
  KeyUsages: []x509.ExtKeyUsage{
    x509.ExtKeyUsageCodeSigning,
  },
}
```

haydentherapper: Go's x509 library won't enforce that CA certificates contain the same EKUs, but some certificate verification libraries will. For Go, it'll just skip over any cert in the chain that contains no EKUs - https://cs.opensource.google/go/go/+/refs/tags/go1.17.6:src/crypto/x509/verify.go;l=1065-1067;drc=refs%2Ftags%2Fgo1.17.6. At a minimum, you could include the code signing EKU in just the leaf cert and include the EKU in `VerifyOptions` and everything would work.

For some background, it's called EKU chaining, where all intermediate certificates need to have the EKUs of the certificates they issue (root certs don't require EKUs). The idea is it restricts the types of certs a certificate authority can issue. This comes up with web PKI, as it's a requirement from the CA/Browser (CAB) forum (https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.8.0.pdf). See *7.1.5 Name constraints*:
> For a Subordinate CA Certificate to be considered Technically Constrained, the certificate MUST include an Extended Key Usage (EKU) extension specifying all extended key usages that the Subordinate CA Certificate is authorized to issue certificates for. The anyExtendedKeyUsage KeyPurposeId MUST NOT appear within
this extension."

FWIW, this is still useful with private PKI, to limit what an intermediate CA can issue, but technically Go won't enforce this. 

Totally unrelated to this PR, in case you're curious, there was an issue [last year](https://groups.google.com/g/mozilla.dev.security.policy/c/EzjIkNGfVEE/m/XSfw4tZPBwAJ) ([another explanation](https://www.hardenize.com/blog/revocation-of-misissued-ocsp-signing-intermediates)) that was related to EKU chaining where intermediate CAs were being incorrectly issued with the EKU for OCSP because they thought it was necessary for chaining.

nsmith5: Amazing!!! Thanks so much for this detailed explanation!

pkg/ca/fileca/load.go
```diff
@@ -27,21 +27,46 @@ import (
...
+			Intermediates: intermediates,
+			KeyUsages: []x509.ExtKeyUsage{
+				x509.ExtKeyUsageCodeSigning,
+			},
+		}
+		if _, err := certs[0].Verify(opts); err != nil {
+			return nil, nil, err
+		}
+
+		// TODO: Additionally verify that cert[0] is a certificate authority
```
haydentherapper, line 66: nit: You already checked that cert[0] is a certificate authority below, line 87

haydentherapper (approved):

pkg/ca/fileca/load.go
```diff
@@ -27,21 +27,46 @@ import (
...
 	}
 
+	// Verify certificate chain
+	{
+		roots := x509.NewCertPool()
+		roots.AddCert(certs[len(certs)-1])
+
+		intermediates := x509.NewCertPool()
+		if len(certs) > 2 {
+			for _, intermediate := range certs[1 : len(certs)-2] {
```
haydentherapper, line 50: The array slice is [inclusive, exclusive), so you'd need to do [1 : len(certs)-1] instead or else you'd exclude one cert

pkg/ca/fileca/load.go
nsmith5: I've documented this in a comment in the code, but the entire `fileca` backend needs documentation as well. I've added a note to #302 to explicitly add this to the docs. Are you alright with catching the docs in that issue / in an iteration?

dlorenc:
The codeql check is failing everywhere, so you can ignore that one!

pkg/ca/fileca/load.go
haydentherapper: Good with me! Thanks for adding the comment.

pkg/ca/fileca/load.go
```diff
@@ -27,21 +27,64 @@ import (
...
 		return nil, nil, err
 	}
 
+	// Verify certificate chain
+	{
+		roots := x509.NewCertPool()
+		roots.AddCert(certs[len(certs)-1])
+
+		intermediates := x509.NewCertPool()
+		if len(certs) > 2 {
```
haydentherapper, line 51: Should this be `if len(certs) > 1`?

nsmith5: I don't think so. In the case of `len(certs) == 2` we add `certs[1]` to the root pool and then verify `certs[0]` and the pool of intermediates is empty. Unless perhaps I misunderstood your previous comments. You don't want to add `certs[0]` to the pool of intermediates of that is the certificate your verifying right?

That said, I don't actually have a test cast for `len(certs) == 2` so I'll add that just to be certain.

haydentherapper: Never mind, this is good! Forgot this was verifying the intermediate. Yea, you don't need to add certs[0] to the pool.

nsmith5: Added test data and test case for `len(certs) == 2` 🚀

haydentherapper (approved):
This looks great! Thanks for adding all of the comprehensive tests!

nsmith5:
Sweeeet! Thanks for the comprehensive review. It really taught me a lot! I really appreciate the time you put into it ❤️

pkg/ca/fileca/fileca.go
```diff
@@ -103,8 +104,19 @@ func (fca *fileCA) CreateCertificate(_ context.Context, subject *challenges.Chal
...
+		err := pem.Encode(buf, &pem.Block{
+			Type:  "CERTIFICATE",
+			Bytes: cert.Raw,
+		})
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return buf.Bytes(), nil
```
bobcallaway, line 121: add import for `github.com/sigstore/sigstore/pkg/cryptoutils` and then:

```suggestion
	return cryptoutils.MarshalCertificatesToPEM(fca.certs)
```

bobcallaway (approved):
one minor change to drive re-use but otherwise LGTM

pkg/ca/fileca/fileca.go
nsmith5: An nice catch! Discovered the some of the test data didn't have the 100year expiration set and had expired so regenerated it.

PR #337, opened 2022-01-20
Add some reasonable timeouts to API server

### LINKED ISSUE #336: Lack of server timeouts can lead to connection exhaustion
**Description**

Fulcio does not set any timeouts on the `http.Server` used to serve the API. This can be seen here

- https://github.com/sigstore/fulcio/blob/6f40e90992a4141b4aa5b2cf43721ce098b9bd0c/cmd/app/serve.go#L196-L199

In this case, the zero value means disabling the timeouts entirely as documented [here](https://pkg.go.dev/net/http@go1.17.6#Server). The implications of disabling these timeouts is discussed in [this Cloudflare article](https://blog.cloudflare.com/exposing-go-on-the-internet/).  To summarize, if an attacker can open enough connections to the server to make the Fulcio server process hit its open file limit the server will fail to open any more connections and deny service to all other requests. Because there are no timeouts, an attacker can hold these connections open for an arbitrarily long time.

**Mitigation**

Configure `http.Server` with reasonable default read and write timeouts.

nsmith5:
#### Summary

Adds a 1 minute hard coded read, header read, write and idle timeout. This helps prevents accidental or malicious connection exhaustion of the API server.

_Implementation note_

While simply setting `ReadTimeout` would have set the header read and write timeouts to the same value, I think setting them explicitly helps inform the future code readers about the server behaviour in case they're unfamiliar with the behaviour of `http.Server` timeouts.

#### Ticket Link

Fixes #336

#### Release Note

```release-note
* Added 1 minute read, write and idle timeouts to API server
```

### DIFF
--- a/cmd/app/serve.go
+++ b/cmd/app/serve.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -206,6 +207,12 @@ func runServeCmd(cmd *cobra.Command, args []string) {
 	api := http.Server{
 		Addr:    host + ":" + port,
 		Handler: handler,
+
+		// Timeouts
+		ReadTimeout:       60 * time.Second,
+		ReadHeaderTimeout: 60 * time.Second,
+		WriteTimeout:      60 * time.Second,
+		IdleTimeout:       60 * time.Second,
 	}
 
 	if err := api.ListenAndServe(); err != nil && err != http.ErrServerClosed {


haydentherapper:
Will this close all open connections, i.e. will this stop the connection to Trillian? Or should we also add a timeout for outbound ?

nsmith5:
This is just for inbound API requests. Timeouts on the outbound requests to the CT log are controllered here and it seems are set to 30 seconds https://github.com/sigstore/fulcio/blob/main/pkg/ctl/ctl.go#L37

haydentherapper (approved):

dlorenc:
LGTM but it hit a merge conflict with your other one!

nsmith5:
Rebased on main and should be good to go now :)

PR #338, opened 2022-01-20
Set max request size to 4MiB

### LINKED ISSUE #335: Unbound request size can lead to OOM
**Description**

There is no size limit imposed on the body of requests to the `/api/v1/signingCert` endpoint. This means that an request can create an
arbitrarily large signing request and force the Fulcio process to be OOM killed by the kernel. The request need only be in the right format because the request body is send into a JSON parser immediately as we can see [here](https://github.com/sigstore/fulcio/blob/6f40e90992a4141b4aa5b2cf43721ce098b9bd0c/pkg/api/ca.go#L125).

An example request that would achieve this would be:

```
{
  "signedEmailAddress": "AAAAAAAAAAAAAAAAAAA.............
```

Where the string of `AAAAA.....` simply continues forever. I believe this request body is valid and would be read until the process runs out of memory.

**Mitigation**

To mitigate, I think we can use `http.MaxBytesReader` to limit the size of requests to this endpoint. I believe we can find a reasonable limit that prevents this malicious behaviour but shouldn't break and valid use cases.

nsmith5:
#### Summary

Fixes the the max request body size to 4MiB (Just a guess at a size.... may it could be smaller?) so that it isn't unbound. Implemented using a new `api.WithMaxBytes` middleware. There is a small refactor of how we wire up stuff for the API server here as well to make it a little easier to understand. As adding the body limiting middleware was part of that wire up it seemed a good thing to clean up with this work.

#### Ticket Link

Fixes #335 

#### Release Note

```release-note
* Set max request size to 4MiB so that request size isn't unbound
```

### DIFF
--- a/cmd/app/serve.go
+++ b/cmd/app/serve.go
@@ -163,26 +163,6 @@ func runServeCmd(cmd *cobra.Command, args []string) {
 		log.Logger.Fatal(err)
 	}
 
-	decorateHandler := func(h http.Handler) http.Handler {
-		// Wrap the inner func with instrumentation to get latencies
-		// that get partitioned by 'code' and 'method'.
-		return promhttp.InstrumentHandlerDuration(
-			api.MetricLatency,
-			http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
-				ctx := r.Context()
-
-				// For each request, infuse context with our snapshot of the FulcioConfig.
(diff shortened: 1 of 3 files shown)


nsmith5:
The diff algorithm made this one a little odd to read. Sorry ya'll 😅

dlorenc (approved):

dlorenc:
Thanks!

PR #339, opened 2022-01-21
Add vaikas to CODEOWNERS

vaikas:
I wasn't quite sure what the process is or the requirements, but here goes :) 

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -6,3 +6,4 @@
 # dlorenc
 # lukehinds
 # mattmoor
+# vaikas


lukehinds:
We don't really have a decent system outlined yet, but I think you would be a really good individual to get on board. 

An observation looking at your commit history that I think what would really bolster the case for making you a maintainer would be if you contributed a core feature. What do you think?

vaikas:
Sounds reasonable, thanks @lukehinds and @bobcallaway :)

lukehinds:
I'll close this for now, can always make another PR again later.

PR #355, opened 2022-01-27
Update README for V1 Fulcio cert

haydentherapper:

#### Ticket Link

Fixes

#### Release Note

```release-note

```


cpanato (commented):
what do you think to add in the PR summary how you get the certificate, just for the records, and maybe for the next update someone that will do that and does not know, can do it.

thanks for this!!

lukehinds:
> what do you think to add in the PR summary how you get the certificate, just for the records, and maybe for the next update someone that will do that and does not know, can do it.
> 
> thanks for this!!

That's actually a good idea and @haydentherapper has a very good handle on the TUF root setup. Any ideas on what might be a better way to show users how to get the root CA Hayden? I don't think we should really have it in a readme any more, it was fine when we were in early development, but not as excusable now.

dlorenc (approved):

lukehinds:
@dlorenc I think this one was in review still, see @cpanato comment

dlorenc:
> @dlorenc I think this one was in review still, see @cpanato comment

Oh my mistake! my fingers went too fast here. The cert here is accurate but we'd probably be better off linking people to the canonical location which is here: https://github.com/sigstore/root-signing/blob/main/repository/repository/targets/fulcio_v1.crt.pem

lukehinds:
@dlorenc is it possible a user (visiting this repo) could validate the cert as a TUF target? I might not be using the right TUFism's here, but someone could verify the cert as part of the root signing verify operation that others did after the key signing party (or perhaps even a subset of the verification)?

dlorenc:
Yeah - it's not trivial but it should be doable with a TUF client. @asraa might have that incantation handy.

asraa:
> Yeah - it's not trivial but it should be doable with a TUF client. @asraa might have that incantation handy.

I can update the README with these instructions! But you can use the go-tuf CLI to do the following

1. Extract trusted root keys from, let's say, the embeded repository in a cosign checkout
```
go get github.com/theupdateframework/go-tuf/cmd/tuf
tuf -d $(COSIGN_DIR)/pkg/cosign/tuf/ root-keys > sigstore-root.json
```
2. Init and get `fulcio_v1.crt.pem`.
```
go get github.com/theupdateframework/go-tuf/cmd/tuf-client
tuf-client init https://raw.githubusercontent.com/sigstore/root-signing/main/repository/repository/ sigstore-root.json
tuf-client get https://raw.githubusercontent.com/sigstore/root-signing/main/repository/repository/ fulcio_v1.crt.pem
```

dlorenc:
Opened #360 to track fixing my mistake on the merge here and enhancing the docs! Sorry about that again.

PR #380, opened 2022-02-02
RFC: An RFC process

### LINKED ISSUE #379: Create a simple RFC process for important changes
**Description**

Sometimes, we there is a need to propose a change to Fulcio requires more space, discussion and scrutiny than Github issues is good for. I believe these change proposals should be both documented once accepted and placed under the code review process for acceptance.

Two examples of such changes I can think off right now are the refactor of are #275 and #304. The are big enough that no single PR should seek to implement the whole thing, but I think as a project we should go through the RFC process to discuss what we'd like it to look like once completed.

We often see Google docs etc floating about for this purpose, but they're not very discoverable once complete.

nsmith5:
#### Summary

This is an RFC for the RFC process itself. I've documented a bit of background in #379 but to summarize:

- We should discuss some big changes upfront with the code review process
- We should document those changes

The attached RFC describes what that process could look like

#### Ticket Link

Fixes #379 

#### Release Note

```release-note
NONE
```

### DIFF
--- a/docs/rfc/00-rpc-process.md
+++ b/docs/rfc/00-rpc-process.md
@@ -1,0 +1,52 @@
+author: Nathan Smith <fulcio@nfsmith.ca>
+date: 2022-02-02
+---
+
+# RFC-00 An RFC process
+
+At the moment on the Fulcio project we have a lightweight planning process that
+roughly involves creating an issue on Github for a bug or enhancement and then
+submitting a patch to fix that issue. This works reasonably well for small
+changes, but sometimes a larger change requires more upfront discussion and a
+clear specification for what the finished change will look like. Additionally
+it would be ideal if these important design decision where documented for
+future contributors.
+
+## The RFC process
+
+To create an RFC, create a copy the RFC template to a new file
+
+```
+cp docs/rfc/_template.md docs/rfc/04-my-rfc.md
+```
+
+Prefix your RFC with its number. Increment the number from the latest RFC.
+Next, describe in detail the change you're proposing to make. This should
+include background information on the problem space so that little prior
+knowledge is assumed of the reader.
+
+Once you're happy with your initial draft, submit your RFC as a pull request.
+The pull request code review process is used to iterate, discuss and come to a
+final conclusion about whether or not the Fulcio project should accept an RFC.
+
+## When to submit an RFC
+
+An RFC should be submitted if any of the following things is true
+
+- Work to implement something will span many pull requests and you want to
+  inform maintainers and other contributors about what the final goal of your
+work is.
+- Your work will change how Fulcio functions in an important way you'd like to
+  document for users and contributors
+
+## Conclusions
+
+If the proposed RFC process is followed the following benefits are gained
+
+- Maintainers become aware of larger strategic work that folks would like to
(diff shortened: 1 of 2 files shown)


docs/rfc/00-rpc-process.md
```diff
@@ -0,0 +1,52 @@
...
+roughly involves creating an issue on Github for a bug or enhancement and then
+submitting a patch to fix that issue. This works reasonably well for small
+changes, but sometimes a larger change requires more upfront discussion and a
+clear specification for what the finished change will look like. Additionally
+it would be ideal if these important design decision where documented for
+future contributors.
+
+## The RFC process
+
+To create an RFC, create a copy the RFC template to a new file
```
dlorenc, line 17: I'd suggest we start in a Google doc or hackmd actually, then move to a pull request toward the end.

GitHub PRs are unfortunately really bad for quick, small changes that would probably occur at the start of a design discussion.

The template can still get filled in and sent as a PR, but as a rule of thumb there probably shouldn't be many comments/revisions at that point otherwise the process becomes painful.

dlorenc: So to make a concrete suggestion: To create an RFC, open an issue with a link to a google doc or hackmd. After general consensus has been reached, convert to markdown and open a PR in a new file.

You can of course just skip the google doc portion if you're feeling lucky :)

nsmith5: Hmmm I just worry that the google docs / hackmd never really make it into the documented code base if that is where they start out. Buttt to be fair, I haven't really participated in this kind of open design work before so I don't know the short coming of Github versus other places to review. Happy to go with prior art or experience. 

The really big thing for me is that we have a pattern for discussing and agreeing on the big work and then making those decisions documented and discoverable after

dlorenc: Ah yeah, it's definitely a risk. It's better than not having them though. Starting in code review is pretty hard because small stuff like wordsmithing at the start tends to result in a million comments, and GitHub makes that kind of thing hard to track. With a doc, the reviewer can just fix typos or address small formatting things directly without needing to leave a comment for the author to do it.

Basically I see it as a pre-step, to build consensus and circulate ideas even earlier than getting to a PR/code review stage. Docs also have the benefit that they can have multiple authors, rather than it just being one person submitting a PR being responsible for making all the changes/edits.

But if you want to skip that step for a specific PR/RFC, then I guess you can always go for it!

nsmith5: Yeah that totally makes sense! Better to have that first low-barrier idea out there and shared early in easy to collaborate format and open the PR once you're pretty set its a good idea. Ok, I'll write that into the process!

lukehinds:
It might be good to fold (or link / reference) this into the work @haydentherapper kicked off 'maintainer friction' over in sigstore/community

Overall I support this though, fulcio is expected to be pumping out what could be millions of certificates a day, so we need decent change control in place. I would really like to see the same on rekor as well.

nsmith5:
Going to close this one per @lukehinds suggestion. Maybe how this works could live in the community repo? Anyways I want to discuss and think about it so I won't leave this PR hanging like this.

Issue #373, opened 2022-02-01
Architecture docs


nsmith5:
**Description**

It takes a really long time to get started with Fulcio development because of the lack of architecture docs. The reader needs to spin everything up and look the the code to get the gist of how things work. I think we can make this much better with basic architecture docs including

- [x] Detailed explanation of how a certificate request works from start to finish
- [ ] Flow diagram of OIDC authentication
- [ ] Clear diagrams and explanation of upstream and downstream dependencies
  - [ ] Explain how the rest of sigstore relies on Fulcio
  - [ ] Explain how Fulcio relies on a certificate transparency log and certificate authority

lukehinds:
Agree, I would add a nice flow diagram of OIDC as well.

nsmith5:
I've got a very detailed description of the certificate request lifecycle in this doc I can repurpose for the "how a certificate request works" part of this https://docs.google.com/document/d/1PKAou7wEmEob4VOLq6rFlx5maOM5hkJ3XffZizfwdFw/edit

endorama:
I've been looking at OICD authentication flow is it correct the one to document is the Basic/Authentication flow? 

Am I correct saying that IDP redirects to fulcio, which then uses the ID token as issuer for the sign certificate? 
(At least this is my understanding from [`actualAuthorize`](https://github.com/sigstore/fulcio/blob/2ce3e6cec9fba93bf632c2e3e257e0aca9632fd1/pkg/api/ca.go#L98) and it's usage)

haydentherapper:
Work is complete

PR #383, opened 2022-02-03
Document the certificate issuing process

nsmith5:
#### Summary

Adds a basic overview of the certificate issuing process to our docs to help bring new collaborators up to speed. Its a little light on the certificate transparency and SCT stuff, but I think its a good starting point.

NB: The images all have the excalidraw source embedded in them so they can be changed / edited by anyone

#### Ticket Link

Relates to #373 

#### Release Note

```release-note
NONE
```


docs/how-certificate-issuing-works.md
```diff
@@ -0,0 +1,70 @@
...
+
+- An OIDC identity token. This is a JWT containing information about the
+  principal (identity of our client), the issuer (who verified this identity?
+Google, Github etc) and some other metadata.
+- The public key. This is the public half of an asymmetric key-pair generated
+  by the client. The public key will be embedded in the final x509 certificate
+if this request is successful
+- A challenge. This challenge proves the client is in possession of the private
+  key that corresponds to the public key provided. The challenge created by
+signing the subject portion of the OIDC ID token
```
lukehinds, line 24: The challenge description reads correct, but is it stuffed into the cert now?

docs/how-certificate-issuing-works.md
```diff
@@ -0,0 +1,70 @@
...
+## 2 | Authentication
+
+The first step in processing the certificate request is to authenticate the
+OIDC ID token.
+
+![OIDC Authentication diagram](img/authenticate-token.png)
+
+To authorize the token Fulcio must:
+
+- Use the issuer claim from the token find the issuers OIDC discovery endpoint
```
lukehinds, line 35: should that be "from the token *to* find the issuers..."

docs/how-certificate-issuing-works.md
```diff
@@ -0,0 +1,70 @@
...
+- Download the issuers signing keys from the discovery endpoint
+- Verify the ID token signature
+
+## 3 | Verifying the challenge
+
+Once the client has been authenticated, the next step is to verify the client
+is in possession of the private key of the public key they’ve submitted. To do
+this we must verify the challenge. This is simply a signature of the `sub`
+claim so we verify the signature using the public key supplied.
```
lukehinds, line 44: I guess you could mention here that the challenge uses the email as the element that is signed.

lukehinds (commented):
Thanks for this Nathan, very needed bit of work. 

I made a few comments, but nothing major

regarding "docs/img/return-certificate.png", I wonder if instead of having "Google CA" we could put "CA Backend" or "interface", something to show its not fixed to Google, as we support other backends.

lukehinds:
cc @bobcallaway / @dlorenc for their input.

docs/how-certificate-issuing-works.md
nsmith5: I don't _think_ so. Are you saying it reads like that or it doesn't read like that and it should to be accurate?

docs/how-certificate-issuing-works.md
nsmith5: The subject isn't always the email right? It can be the workflow URL or SPIFFE ID as well as I understand it. Thay is why I kept it general as the subject in the ID token, but perhaps I can enumerate the different things that can be?

docs/how-certificate-issuing-works.md
lukehinds: actually reading it again, I missed the reference to "request", so I misread it as being in the certificate 

> "The certificate request contains three items:"

docs/how-certificate-issuing-works.md
lukehinds: fair point

nsmith5:
Rewrote the end and expanded on the certificate authority backend, ct log upload etc. I find it easiest to review the rendered content here https://github.com/nsmith5/fulcio/blob/921a3e5b74fafe219d6a240351ae74c7c3991568/docs/how-certificate-issuing-works.md

docs/how-certificate-issuing-works.md
nsmith5: Ahhh right I see. Yeah I just meant the API request if that makes sense. I think I can clarify that

dlorenc:
This looks good to me!

dlorenc (approved):

PR #407, opened 2022-02-14
Automate release

k4leung4:
#### Summary

Create a Github Actions Workflow that will mostly automate the cutting of a release
- draft a release with specified tag [this uses the gh cli]
- kick off cloud build to create artifacts [this uses gcloud cli]; this will also require the creation of a new service account to leverage the oidc login flow, the commands/script to create the service account is included in the PR.

only inputs needed are 
- release tag
- cosign key ring name
- cosign key name

The last two parameter can be hardcoded or if it is sensitive, stored as a secret and referenced that way. Let me know if this is desirable and I can update the workflow.

To avoid just anyone starting the workflow to create a release, it is hardcoded to allow one of
"bobcallaway","cpanato","dlorenc","lukehinds" to start it.
Anyone else that starts it will result in an immediate failure.

if this approach makes sense, i plan to check in a reusable workflow into sigstore/sigstore and have fulcio|rekor|cosign all have release workflows using it.

#### Ticket Link

Fixes

https://github.com/sigstore/public-good-instance/issues/50

#### Release Note

```release-note
NONE
```

### DIFF
--- a/.github/workflows/cut-release.yml
+++ b/.github/workflows/cut-release.yml
@@ -1,0 +1,54 @@
+name: Cut Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_tag:
+        required: true
+        type: string
+        description: 'Release tag'
+      key_ring:
+        required: true
+        type: string
+        description: 'Key ring for cosign key'
+      key_name:
+        required: true
+        type: string
+        description: 'Key name for cosign key'
+
+concurrency: cut-release
+
+jobs:
+  cut-release:
+    name: Cut release
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    env:
+      GIT_TAG: ${{ github.event.inputs.release_tag }}
+      PROJECT_ID: 'projectsigstore'
+    steps:
+      - name: Check actor access
+        if: ${{ !contains(fromJson('["bobcallaway","cpanato","dlorenc","lukehinds"]'), github.actor) }}
+        run: exit 1
+
+      - name: Checkout out repo
+        uses: actions/checkout@v2
+        with:
+          path: ./src/github.com/sigstore/fulcio
+
+      - name: Authenticate to Google Cloud
+        uses: google-github-actions/auth@v0
+        with:
+          workload_identity_provider: 'projects/498091336538/locations/global/workloadIdentityPools/githubactions/providers/sigstore-fulcio'
+          service_account: 'github-actions-fulcio@projectsigstore.iam.gserviceaccount.com'
+
+      - name: Setup gcloud
+        uses: 'google-github-actions/setup-gcloud@v0'
+        with:
+          project_id: ${{ env.PROJECT_ID }}
+
+      - name: Start cloudbuild job
+        working-directory: ./src/github.com/sigstore/fulcio
+        run: gcloud builds submit --config release/cloudbuild.yaml --substitutions _GIT_TAG=${{ env.GIT_TAG }},_TOOL_ORG=sigstore,_TOOL_REPO=fulcio,_STORAGE_LOCATION=fulcio-releases,_KEY_RING=${{ github.event.inputs.key_ring }},_KEY_NAME=${{ github.event.inputs.key_name }} --project=${{ env.PROJECT_ID }}
--- a/hack/github-oidc-setup.sh
+++ b/hack/github-oidc-setup.sh
@@ -1,0 +1,70 @@
+#!/usr/bin/env bash
+
+# Copyright 2022 The Sigstore Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


cpanato (commented):
we already have a release job similar we have for cosign/rekor https://github.com/sigstore/fulcio/tree/main/release do we want to switch that to github actions or keep it in the same way we do for others?

@k4leung4 @dlorenc

k4leung4:
> we already have a release job similar we have for cosign/rekor https://github.com/sigstore/fulcio/tree/main/release do we want to switch that to github actions or keep it in the same way we do for others?
> 
> @k4leung4 @dlorenc

Thanks for the comment @cpanato 
My intent is to reduce the amount of copy and paste needed to do the release, and provide the option to cut bi-weekly, or monthly minor releases automatically.
This is a change I aim to apply to the other repos [rekor|cosign] as well.

cpanato:
> > we already have a release job similar we have for cosign/rekor https://github.com/sigstore/fulcio/tree/main/release do we want to switch that to github actions or keep it in the same way we do for others?
> > @k4leung4 @dlorenc
> 
> Thanks for the comment @cpanato My intent is to reduce the amount of copy and paste needed to do the release, and provide the option to cut bi-weekly, or monthly minor releases automatically. This is a change I aim to apply to the other repos [rekor|cosign] as well.

I'm sorry @k4leung4 I did not see the last part of the job, that submits the cloudbuild job 🤦 , so sorry :(

that looks great to me!

.github/workflows/cut-release.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Checkout out repo
+        uses: actions/checkout@v2
+        with:
+          path: ./src/github.com/sigstore/fulcio
+
+      - name: Draft release
+        working-directory: ./src/github.com/sigstore/fulcio
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: gh release create --repo github.com/sigstore/fulcio --draft ${{ env.GIT_TAG }}
```
cpanato, line 45: we don't need this, the goreleaser that will run in the cloudbuild job will create the gh release for us

cpanato: then maybe we can remove the `contents` permission

hack/github-oidc-setup.sh
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
```
cpanato, line 1: this script is needed? i did not see that being used in the release action

.github/workflows/cut-release.yml
k4leung4: thanks for pointing that out.
removed.

hack/github-oidc-setup.sh
k4leung4: apologies for not referencing in the README. 
it is needed to setup the oidc access for github actions to authenticate against GCP to kick off the gcloud builds command.

updated README.md to explain it is a one time setup.

.github/workflows/cut-release.yml
```diff
@@ -0,0 +1,54 @@
...
+          service_account: 'github-actions-fulcio@projectsigstore.iam.gserviceaccount.com'
+
+      - name: Setup gcloud
+        uses: 'google-github-actions/setup-gcloud@v0'
+        with:
+          project_id: ${{ env.PROJECT_ID }}
+
+      - name: Start cloudbuild job
+        working-directory: ./src/github.com/sigstore/fulcio
+        run: gcloud builds submit --config release/cloudbuild.yaml --substitutions _GIT_TAG=${{ env.GIT_TAG }},_TOOL_ORG=sigstore,_TOOL_REPO=fulcio,_STORAGE_LOCATION=fulcio-releases,_KEY_RING=${{ github.event.inputs.key_ring }},_KEY_NAME=${{ github.event.inputs.key_name }} --project=${{ env.PROJECT_ID }}
```
lukehinds, line 54: Do we need google cloud, is it not something we do in github?

lukehinds:
also a commit squash / rebase (which every method you prefer) would be good (unless you have some more pending commits)

.github/workflows/cut-release.yml
cpanato: we are building the release using GCP cloudbuild and GCP KMS and the oidc token there as well, this will help to send the job to cloudbuild and not need to run locally

however if we decide to run the release in GitHub actions we will need to refactor somethings, but is possible

lukehinds: I guess a better way to frame the question is if there is a cost involved in running it on GCP. Right now that is OK as big G is footing the bill for us, but that might not always be the case.

cpanato: it costs `$0.016 / build-minute` today, plus the costs of storing the secret manager and the KMS key for the cosign private key.

if we revamp this to run on GH actions we will just need to check if we will continue to use the KMS key or we will use another key or move completely to keyless signature.

the refactor to GH should be very straightforward. We just need to make a decision

dlorenc: Let's keep it consistent for now. We had to use gcb initially for the auth to registries, but now that github supports OIDC we can reevaluate moving everything here.

k4leung4:
thanks for all the feedback.

commits squashed.

.github/workflows/cut-release.yml
cpanato: sounds good to me

cpanato (approved):

Issue #372, opened 2022-02-01
Break down mega-README into bite size articles under /docs


nsmith5:
**Description**

The README has sort of just grown organically over time and I think its high time it gets broken down into more focused documents under something like a `/docs` folder. We can link to that folder form the README so the information is still easily discoverable, but keep the copy in the README a tighter narrative about what Fulcio does. The focus here would be leaning towards contributors primarily as the end-user documentation lives at https://docs.sigstore.dev which lives over here https://github.com/sigstore/sigstore-website.

dlorenc:
Big +1 to this! For GitHub "SEO" it's usually good to have a small blurb/diagram/picture targeted to end-users "above the fold" in the README, but the rest of the end-user stuff should move to the docs website and this can stay contributor focused.

nsmith5:
Yes absolutely!

endorama:
Ok so first step is breaking down the README, then I'll start moving things around.

nsmith5:
Readme is a lot more reasonable now. Closing this

PR #409, opened 2022-02-15
Extract CA/KMS support from README

endorama:
Part of https://github.com/sigstore/fulcio/issues/372

This PR extract the part related to CA/KMS support from README to it's own file in `docs/` folder.

/cc @nsmith5


nsmith5:
I think section of the README would generalize well into an operator manual and could include other details of configuration. I think it ultimately should live in the main docs website as well.

This is a good move for right now, but definitely some follow up to think about.

dlorenc (approved):

PR #430, opened 2022-02-23
Upload code coverage report to CodeCov

nsmith5:
#### Summary

Uploading code coverage reports helps nudge contributors to add tests as the coverage diff is posted to their PR.

#### Ticket Link

Relates to ##368. Reporting helps us get a sense of poorly tested code areas.

#### Release Note

```release-note
* NONE
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,6 +36,10 @@ jobs:
       - name: Build
         run: make -C $GITHUB_WORKSPACE all
       - name: Test
-        run: go test -v ./...
+        run: go test -v -covermode=atomic -coverprofile=coverage.out ./...
+      - name: Upload Coverage Report
+        uses: codecov/codecov-action@f32b3a3741e1053eb607407145bc9619351dc93b # v2.1.0
+        with:
+          files: coverage.out
       - name: Ensure no files were modified as a result of the build
         run: git update-index --refresh && git diff-index --quiet HEAD -- || git diff --exit-code


.github/workflows/main.yml
```diff
@@ -36,6 +36,10 @@ jobs:
       - name: Build
         run: make -C $GITHUB_WORKSPACE all
       - name: Test
-        run: go test -v ./...
+        run: go test -v -covermode=atomic -coverprofile=coverage.out ./...
+      - name: Upload Coverage Report
+        uses: codecov/codecov-action@f32b3a3741e1053eb607407145bc9619351dc93b # v2.1.0
```
haydentherapper, line 41: Is this isolated such that if codecov was compromised in the same way that it was previously compromised, the action couldn't exfiltrate secrets from our workflow?

nsmith5: Someone would need to add a secret to this workflow (there are currently none) or add additional permissions to the github token to do anything as I understand it

nsmith5: FWIW codecov completely rewrote the uploader so the attack vector would also have to be different this time as well

nsmith5: For some background, CodeCov also changed their uploader significantly after the incident. Its no longer a bash script that you download via curl. Its now a static binary. This action downloads the binary and checks the signature using their signing key.

The manual process to do this is documented here https://docs.codecov.com/docs/codecov-uploader

nsmith5:
Ah derp. I didn't realize there was an existing and un-concluded discussion on this front already https://github.com/sigstore/rekor/pull/676

I'm going to close this for now and focus the discussion over there until we've settled the approach together

PR #434, opened 2022-02-23
add indent to fix yaml error

bobcallaway:

### DIFF
--- a/.github/workflows/verify-k8s.yml
+++ b/.github/workflows/verify-k8s.yml
@@ -212,6 +212,6 @@ jobs:
         env:
           KO_DOCKER_REPO: registry.local:5000/fulcio
 
-    - name: Collect diagnostics and upload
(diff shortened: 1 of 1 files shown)


cpanato:
i was thinking if we can run some yaml lint to catch those things

mattmoor (approved):

mattmoor:
@cpanato Kenny found an action linter we just started running downstream: https://github.com/chainguard-dev/actions/blob/4ba8d060251254fc0e65500a8d3a90013a22a8d7/.github/workflows/actionlint.yaml#L20-L21

Not sure if would catch this, but it sure is annoying that these fail so quietly.  Apologies for the break 🤦

k4leung4:
I tested yaml in their playground, it would have caught it.
playground at https://rhysd.github.io/actionlint/

PR #445, opened 2022-03-02
Generate release yaml for non-CI builds.

k4leung4:
Signed-off-by: Kenny Leung <kleung@chainguard.dev>

#### Summary

This will generate a YAML artifact for releases.

As I understand it, the current release involves building 5 images, each for a different architecture. With these 5 images, we use docker compose to create a single manifest list from these individual images to support the different architecture. It is this composed image that is signed.

For generating the YAML, we leverage one of the existing ko publish commands, changing it to ko resolve, which generates the YAML. At this point, it will point to one of the single architecture images. After composing the final image, we update the YAML with the manifest list. 

#### Ticket Link

Fixes

#### Release Note

```release-note
Generate a release YAML artifact
```

### DIFF
--- a/release/release.mk
+++ b/release/release.mk
@@ -23,35 +23,35 @@ ALL_ARCH = amd64 arm arm64 ppc64le s390x
 ko-release:
 # amd64
 	LDFLAGS="$(LDFLAGS)" GIT_HASH=$(GIT_HASH) GIT_VERSION=$(GIT_VERSION) \
-	ko publish --base-import-paths --bare \
+	ko resolve --base-import-paths \
 		--platform=linux/amd64 --tags $(GIT_VERSION)-amd64 --tags $(GIT_HASH)-amd64 \
-		github.com/sigstore/fulcio
+		--filename config/ > $(FULCIO_YAML)
 
 # arm64
 	LDFLAGS="$(LDFLAGS)" GIT_HASH=$(GIT_HASH) GIT_VERSION=$(GIT_VERSION) \
 	CC=aarch64-linux-gnu-gcc \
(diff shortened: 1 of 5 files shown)


release/release.mk
```diff
@@ -22,37 +22,37 @@ ALL_ARCH = amd64 arm arm64 ppc64le s390x
 .PHONY: ko-release
 ko-release:
 # amd64
-	LDFLAGS="$(LDFLAGS)" GIT_HASH=$(GIT_HASH) GIT_VERSION=$(GIT_VERSION) \
-	ko publish --base-import-paths --bare \
-		--platform=linux/amd64 --tags $(GIT_VERSION)-amd64 --tags $(GIT_HASH)-amd64 \
-		github.com/sigstore/fulcio
+	ARCH="amd64" LDFLAGS="$(LDFLAGS)" GIT_HASH=$(GIT_HASH) GIT_VERSION=$(GIT_VERSION) \
+	ko resolve --base-import-paths \
+		--platform=linux/$(ARCH) --tags $(GIT_VERSION)-$(ARCH) --tags $(GIT_HASH)-$(ARCH) \
```
cpanato, line 27: the $(ARCH) does not get filled with the data, maybe there is another way to do that, but maybe we keep as the original for now, i was planning to do a refactor on this and use a for loop

cpanato: besides that i was able to generate the yaml :)

k4leung4: thanks for finding that issue. i just dont understand shell sometimes.

cpanato (approved):

dlorenc (approved):

PR #463, opened 2022-03-09
Update Username OIDC flow based on comments 

### LINKED ISSUE #398: OIDC enhancement: Support for additional OIDC subjects
## Background

As discussed in https://github.com/rubygems/rfcs/pull/37, there is a desire to specify other values in OIDC tokens besides emails, SPIFFE IDs, or Github OIDC subjects. RubyGems is proposing using the profile URI as the subject of the OIDC token. As other IDPs integrate with Fulcio, subjects could also simply be usernames without any domain.

Integration with IDPs can be done with Dex or directly through Fulcio's list of OIDC issuers. For this issue, I will focus on the latter. There are benefits to integrating directly with Fulcio, primarily automation.

I propose that we add support for two types of domain-scoped subjects:
1. URIs that contain a domain and user identifier
2. Usernames along with a domain for which the username is valid

## Details

We will add new [Types](https://github.com/sigstore/fulcio/blob/4d0f5399c71c3cf2f42ce458b748e6f62a6ed566/pkg/config/config.go#L156) for each of these flows.

### Type: `uri`

OIDC tokens must have a subject with the following structure: `domain/user-identifier`

For example, consider RubyGems profiles. For `https://rubygems.org/profiles/username`, the domain is `https://rubygems.org` and the user identifier is `profiles/username`. We will require that the issuer URL host (where the openid configuration is hosted) matches the domain specified in the token. It is up to the client to interpret the user identifier.

Issued certificates will include the subject as a URI subject alternative name, `URI:domain/user-identifer`.

Example configuration, assuming RubyGems sets up the IDP at `/auth`:

```
"https://rubygems.org": {
      "IssuerURL": "https://rubygems.org/auth",
      "ClientID": "sigstore",
      "Type": "uri"
    },
```

### Type: `user-with-domain` (name WIP)

OIDC tokens must have a subject without a domain, for example a username or a user ID. As part of configuring the OIDC issuer, a domain will be specified. Like with `uri`, we will require that the issuer URL host matches the domain specified in the configuration. 

Issued certificates will include the subject as an email subject alternative name. Fuclio will concatenate the subject and domain specified in the config, `EMAIL:username@domain`.

Example configuration:

```
"https://rubygems.org": {
      "IssuerURL": "https://rubygems.org/auth",
      "ClientID": "sigstore",
      "Domain": "rubygems.org"
      "Type": "user-with-domain"
    },
```

### Domain verification

Requiring that the issuer URL matches the domain specified in the token or configuration will mitigate the risk of a misconfigured IDP issuing tokens scoped to other domains. For example, if another trusted IDP attempted to issue a token with a subject of `https://rubygems.org/profiles/user`, on verification, Fulcio will only use the configuration named `https://rubygems.org`.

For a future improvement, we can use a domain verification system like ACME to verify that a requester who wants to add an OIDC IDP to Fulcio has control over that domain.

haydentherapper:
#### Summary

#### Ticket Link

Fixes #398

#### Release Note

```release-note
Added OIDC issuer type for username subjects
```

### DIFF
--- a/pkg/challenges/challenges.go
+++ b/pkg/challenges/challenges.go
@@ -277,6 +277,10 @@ func URI(ctx context.Context, principal *oidc.IDToken, pubKey crypto.PublicKey, 
 func Username(ctx context.Context, principal *oidc.IDToken, pubKey crypto.PublicKey, challenge []byte) (*ChallengeResult, error) {
 	username := principal.Subject
 
+	if strings.Contains(username, "@") {
+		return nil, errors.New("username cannot contain @ character")
+	}
+
 	cfg, ok := config.FromContext(ctx).GetIssuer(principal.Issuer)
 	if !ok {
 		return nil, errors.New("invalid configuration for OIDC ID Token issuer")
@@ -293,7 +297,7 @@ func Username(ctx context.Context, principal *oidc.IDToken, pubKey crypto.Public
 	if err != nil {
 		return nil, err
 	}
-	if err := isDomainAllowed(cfg.SubjectDomain, uIssuer.Hostname()); err != nil {
+	if err := validateAllowedDomain(cfg.SubjectDomain, uIssuer.Hostname()); err != nil {
 		return nil, err
 	}
 
@@ -412,12 +416,12 @@ func isURISubjectAllowed(subject, issuer *url.URL) error {
 		return fmt.Errorf("subject (%s) and issuer (%s) URI schemes do not match", subject.Scheme, issuer.Scheme)
 	}
 
-	return isDomainAllowed(subject.Hostname(), issuer.Hostname())
+	return validateAllowedDomain(subject.Hostname(), issuer.Hostname())
 }
 
-// isDomainAllowed compares two hostnames, returning an error if the
+// validateAllowedDomain compares two hostnames, returning an error if the
 // top-level and second-level domains do not match
-func isDomainAllowed(subjectHostname, issuerHostname string) error {
+func validateAllowedDomain(subjectHostname, issuerHostname string) error {
 	// If the hostnames exactly match, return early
 	if subjectHostname == issuerHostname {
 		return nil
@@ -427,14 +431,14 @@ func isDomainAllowed(subjectHostname, issuerHostname string) error {
 	sHostname := strings.Split(subjectHostname, ".")
 	iHostname := strings.Split(issuerHostname, ".")
 	if len(sHostname) < minimumHostnameLength {
-		return fmt.Errorf("subject URI hostname too short: %s", subjectHostname)
+		return fmt.Errorf("URI hostname too short: %s", subjectHostname)
 	}
 	if len(iHostname) < minimumHostnameLength {
-		return fmt.Errorf("issuer URI hostname too short: %s", issuerHostname)
+		return fmt.Errorf("URI hostname too short: %s", issuerHostname)
 	}
 	if sHostname[len(sHostname)-1] == iHostname[len(iHostname)-1] &&
 		sHostname[len(sHostname)-2] == iHostname[len(iHostname)-2] {
 		return nil
 	}
-	return fmt.Errorf("subject and issuer hostnames do not match: %s, %s", subjectHostname, issuerHostname)
+	return fmt.Errorf("hostname top-level and second-level domains do not match: %s, %s", subjectHostname, issuerHostname)
 }
--- a/pkg/challenges/challenges_test.go
+++ b/pkg/challenges/challenges_test.go
@@ -147,6 +147,32 @@ func TestUsername(t *testing.T) {
 	}
 }
 
+func TestUsernameInvalidChar(t *testing.T) {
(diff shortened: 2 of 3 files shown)


haydentherapper:
cc @znewman01

dlorenc (approved):
Nice! Sorry for the delay.

pkg/challenges/challenges.go
```diff
@@ -363,13 +408,16 @@ func isSpiffeIDAllowed(host, spiffeID string) bool {
...
 	if subject.Scheme != issuer.Scheme {
 		return fmt.Errorf("subject (%s) and issuer (%s) URI schemes do not match", subject.Scheme, issuer.Scheme)
 	}
 
+	return isDomainAllowed(subject.Hostname(), issuer.Hostname())
+}
+
+// isDomainAllowed compares two hostnames, returning an error if the
+// top-level and second-level domains do not match
+func isDomainAllowed(subjectHostname, issuerHostname string) error {
```
znewman01, line 420: nit: I think of functions starting with `is` as returning bools -- maybe `validateDomainAllowed` instead?

znewman01, line 420: Also the error messages talk about "subject URI" and "issuer URI" which to me doesn't seem like enough context—if you call `Username`, it implicitly fetches the config and it's an error in the *configuration* that it's complaining about, not really an error in what you passed it

pkg/challenges/challenges.go
```diff
@@ -273,6 +274,50 @@ func URI(ctx context.Context, principal *oidc.IDToken, pubKey crypto.PublicKey,
...
+	if err := CheckSignature(pubKey, challenge, username); err != nil {
+		return nil, err
+	}
+
+	issuer, err := oauthflow.IssuerFromIDToken(principal, cfg.IssuerClaim)
+	if err != nil {
+		return nil, err
+	}
+
+	emailSubject := fmt.Sprintf("%s@%s", username, cfg.SubjectDomain)
```
znewman01, line 310: Is there any validation we want to do on `username`? I worry about passing things like `zack@example.com` as the username for issuer `somethingelse.com`  and some validation code getting confused down the line.

Less nefariously, we might have `zack@example.com` with issuer `example.com` and getting email `zack@example.com@example.com`

pkg/challenges/challenges_test.go
```diff
@@ -177,6 +215,67 @@ func Test_isURISubjectAllowed(t *testing.T) {
...
+		subject: "example",
+		issuer:  "example.com",
+		want:    fmt.Errorf("subject URI hostname too short: example"),
+	}, {
+		name:    "issuer domain too short",
+		subject: "example.com",
+		issuer:  "issuer",
+		want:    fmt.Errorf("issuer URI hostname too short: issuer"),
+	}, {
+		name:    "domain mismatch",
```
znewman01, line 255: nit: maybe a test case for "domain mismatch, subdomain match"

znewman01, line 255: nit: maybe a test case for "domain mismatch, subdomain match"

haydentherapper:
Whoops, #483 submitted the change too. I'll rebase and update based on the comments.

dlorenc:
> Whoops, #483 submitted the change too. I'll rebase and update based on the comments.

Oh wow, somehow I only saw the second commit there with the tests.

pkg/challenges/challenges_test.go
haydentherapper: To me this is already covered by domain mismatch, because the function documents that it only checks top-level and second-level domains. The test for `subdomain mismatch` exists to show that subdomains are not checked.

pkg/challenges/challenges.go
haydentherapper: Great point! Ultimately it'd be up to the validator to be aware of this, and I'd hope that they're checking matching strings rather than parsing the username. I'll add a check that `@` is not present in the username, that seems very reasonable. If for some reason this comes up again later on, we can relax this requirement.

Another option could be to call [`ParseAddress`](https://pkg.go.dev/net/mail#ParseAddress) after constructing the subject, but I don't think this is necessary, as we aren't requiring this to be a routable email.

haydentherapper (commented):
Thanks!

pkg/challenges/challenges.go
znewman01: +1 to do the bare minimum check for `@`, but nothing fancier than that

pkg/challenges/challenges_test.go
znewman01: IMO the test is for triple-checking that we weren't lying when we wrote the documentation :)

I could imagine a sloppy implementation where that would fail (e.g., if you accidentally used index `0` instead of `len(..)-2`.

znewman01:
All looks great!

pkg/challenges/challenges_test.go
haydentherapper: Added another test!

dlorenc (approved):

PR #472, opened 2022-03-12
WIP: add GRPC interface

bobcallaway:
This adds a GRPC interface to `fulcio` and moves the HTTP interface to be a wrapper around GRPC (via grpc-gateway) to ensure compatibility. 

The purpose of this is to provide a simpler, cleaner interface for client code generation across sigstore libraries/services and other consumers. I am also looking to move Rekor over to a similar model but wanted to start here since the API is simpler.

TODO:

- [ ] add unit tests
- [ ] add integration tests
- [ ] remove duplicate code in existing HTTP-based impl in `ca.go` and elsewhere


dlorenc:
> This adds a GRPC interface to `fulcio` and moves the HTTP interface to be a wrapper around GRPC (via grpc-gateway) to ensure compatibility.
> 
> The purpose of this is to provide a simpler, cleaner interface for client code generation across sigstore libraries/services and other consumers. I am also looking to move Rekor over to a similar model but wanted to start here since the API is simpler.

Big +1, OpenAPI/swagger just don't add enough value for the complexity.

bobcallaway:
@loosebazooka @mattmoor @dlorenc any chance I can get one of you to look at and/or recreate the failed test in kind outside of CI? I can't seem to reproduce it running locally (either with kind or docker-compose). 

The error from the `check-oidc` job in the latest run is:
```
2022-03-13T18:32:08.144448806Z stderr F Generating ephemeral keys...
2022-03-13T18:32:08.14469571Z stderr F Retrieving signed certificate...
2022-03-13T18:32:08.759737048Z stderr F tlog entry created with index: 1651798
2022-03-13T18:32:08.75984375Z stderr F Error: signing [registry.local:5000/fulcio/fulcio-bd4d600f674b081a535a886a15cb8d9a@sha256:cb7bf620bb822701f48b440114f846a0aef307bab344346a7d676777467a2f88]: signing digest: error during PEM decoding
2022-03-13T18:32:08.759853651Z stderr F main.go:46: error during command execution: signing [registry.local:5000/fulcio/fulcio-bd4d600f674b081a535a886a15cb8d9a@sha256:cb7bf620bb822701f48b440114f846a0aef307bab344346a7d676777467a2f88]: signing digest: error during PEM decoding
```

I read this as: `cosign`'s interaction with `fulcio` went fine (denoted by the fact we were successfully able to upload the signature to rekor), the failure seems to occur when adding the signature back to the registry for the container where there is some issue parsing the cert or the chain for the OCI signature. 

I'm also able to query and parse the root cert for fulcio just fine:

```
[bcallawa@bcallawa-m900z cosign]$ curl -s http://127.0.0.1:5555/api/v1/rootCert | openssl x509 -noout -text
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            b6:4d:00:f1:5d:c4:73:f0:8d:e0:e5:a0:3c:32:60:28:40:3b:fe
        Signature Algorithm: ecdsa-with-SHA384
        Issuer: O = sigstore.dev, CN = sigstore
        Validity
            Not Before: Oct  7 13:56:59 2021 GMT
            Not After : Oct  5 13:56:58 2031 GMT
        Subject: O = sigstore.dev, CN = sigstore
        Subject Public Key Info:
            Public Key Algorithm: id-ecPublicKey
                Public-Key: (384 bit)
                pub:
                    04:fb:5d:e1:53:e2:b6:f7:3d:01:b0:4b:82:1a:8e:
                    d2:e4:df:f3:a5:9e:98:1a:9e:06:81:72:56:29:b1:
                    80:6b:e6:2f:b8:ca:70:74:ed:c7:9b:dc:b3:f4:38:
                    83:99:77:17:b1:5f:af:5c:e6:25:6e:c8:94:50:f8:
                    7c:f4:e7:28:be:50:5d:ee:05:60:25:1e:98:92:e6:
                    c8:74:f8:7d:86:1c:4e:d2:5e:b9:35:10:2e:66:d5:
                    3a:f5:f4:bf:60:83:dd
                ASN1 OID: secp384r1
                NIST CURVE: P-384
        X509v3 extensions:
            X509v3 Key Usage: critical
                Certificate Sign, CRL Sign
            X509v3 Basic Constraints: critical
                CA:TRUE
                58:C0:1E:5F:91:45:A5:66:A9:7A:CC:90:A1:93:22:D0:2A:C5:C5:FA
                keyid:58:C0:1E:5F:91:45:A5:66:A9:7A:CC:90:A1:93:22:D0:2A:C5:C5:FA

    Signature Algorithm: ecdsa-with-SHA384
         30:66:02:31:00:8f:59:c7:79:76:69:fb:5d:cd:58:13:5a:f8:
         40:ec:0c:ff:06:d5:65:a0:d6:d0:8c:58:ff:d6:1c:fa:a9:69:
         5a:34:8e:1b:30:78:d1:59:81:2b:34:78:4e:f0:60:8e:2a:02:
         31:00:d9:60:7d:a2:df:7c:b0:89:28:17:7b:d9:61:d7:77:fd:
         5b:56:07:96:fd:4c:d3:1e:6b:b2:31:fe:cb:49:e5:37:dc:2c:
         b7:80:04:b1:38:04:d2:4e:b1:0e:2f:9c:11:c9
```

fulcio.proto
```diff
@@ -0,0 +1,86 @@
...
+service CA {
+    /**
+     * Returns information about the current state of the transparency log
+     */
+    rpc GetSigningCert(CertificateRequest) returns (CertificateResponse){
+        option (google.api.http) = {
+          post: "/api/v1/signingCert"
+          body: "*"
+        };
+    }
```
mattmoor, line 33: Can we document some of the custom response munging we have to do on the HTTP path?  I could see folks naively trying to reimplement this proto API and having a bad time.

pkg/api/error.go
```diff
@@ -39,6 +42,11 @@ func handleFulcioAPIError(w http.ResponseWriter, req *http.Request, code int, er
 		message = http.StatusText(code)
 	}
 
-	log.RequestIDLogger(req).Errorw("exiting with error", append([]interface{}{"handler", req.URL.Path, "statusCode", code, "clientMessage", message, "error", err}, fields...)...)
+	// log.RequestIDLogger(req).Errorw("exiting with error", append([]interface{}{"handler", req.URL.Path, "statusCode", code, "clientMessage", message, "error", err}, fields...)...)
```
mattmoor, line 45: delete?  I see the function below, but I'm not sure where you are calling it.

mattmoor (commented):
Generally +1, but I didn't review it super thoroughly.

The dance we have to do for the weird content types some of the APIs use is unfortunate, but hopefully we can switch the main consumers to the GRPC path.

cc @nsmith5 @tstromberg

fulcio.proto
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * Returns the current version information for this Fulcio instance
+     */
+    rpc Version(google.protobuf.Empty) returns (VersionResponse){
+        option (google.api.http) = {
+          get: "/api/v1/version"
+        };
+    }
+}
+
```
haydentherapper, line 51: There's some stylistic changes I'd recommend - https://google.aip.dev/general describes a style for APIs, and if we follow it closely, the GRPC tooling should work nicely.

fulcio.proto
```diff
@@ -0,0 +1,86 @@
...
+message CertificateRequest {
+    string OIDCIdentityToken = 1; // The OIDC token that identifies the caller
```
haydentherapper, line 53: We should prefer underscore case, `oidc_identity_token`, `public_key`, etc

fulcio.proto
```diff
@@ -0,0 +1,86 @@
...
+    PublicKey publicKey      = 2; // The public key to be stored in the requested certificate
+    bytes signedEmailAddress = 3; // Proof that the client possesses the private key
```
haydentherapper, line 55: Can we use something else besides email address? Maybe `signed_subject`? Because for cases like SPIFFE, it's not an email

fulcio.proto
```diff
@@ -0,0 +1,86 @@
...
+}
+
+message PublicKey {
+    string algorithm = 1; // The cryptographic algorithm to use with the key material
```
haydentherapper, line 59: Should this be an enum? Otherwise, we need to list which types of algorithms are supported.

fulcio.proto
```diff
@@ -0,0 +1,86 @@
...
+     * The signed certificate timestamp is a promise for including the certificate in
+     * a certificate transparency log. It can be "stapled" to verify the inclusion of
+     * a certificate in log in an offline fashion.
+     */
+    bytes signedCertificateTimestamp = 1;
+    /*
+     * The full certificate chain, serialized with the leaf certificate first followed
+     * by all subsequent certificates (finishing with the root certificate).
+     */
+    string certificate = 2;
```
haydentherapper, line 74: Specify encoding of certificates?

fulcio.proto
```diff
@@ -0,0 +1,86 @@
...
+    bytes content    = 2; // DER or PEM encoded public key
+}
+
+message CertificateResponse {
+    /*
+     * The signed certificate timestamp is a promise for including the certificate in
+     * a certificate transparency log. It can be "stapled" to verify the inclusion of
+     * a certificate in log in an offline fashion.
+     */
+    bytes signedCertificateTimestamp = 1;
```
haydentherapper, line 69: nit: `signed_certificate_timestamp`

fulcio.proto
```diff
@@ -0,0 +1,86 @@
...
+message PublicKey {
+    string algorithm = 1; // The cryptographic algorithm to use with the key material
+    bytes content    = 2; // DER or PEM encoded public key
+}
+
+message CertificateResponse {
+    /*
+     * The signed certificate timestamp is a promise for including the certificate in
+     * a certificate transparency log. It can be "stapled" to verify the inclusion of
+     * a certificate in log in an offline fashion.
```
haydentherapper, line 67: nit: ... in the log ...

fulcio.proto
```diff
@@ -0,0 +1,86 @@
...
+    bytes signedCertificateTimestamp = 1;
+    /*
+     * The full certificate chain, serialized with the leaf certificate first followed
+     * by all subsequent certificates (finishing with the root certificate).
+     */
+    string certificate = 2;
+}
+
+message RootCertificateResponse {
+    string certificate = 1; // The root certificate for this Fulcio instance
```
haydentherapper, line 78: Not sure if this is something to deal with here, but currently, RootCert for CA Service-backed instances might return multiple root certificates. I think we should standardize the expectation - Either make this repeated, or only allow one CA per CA Service pool.

haydentherapper, line 78: encoding?

haydentherapper:
Left a bunch of API-specific comments. The main consumer annoyance I predict is around consuming a certificate chain. I think we should try to resolve that here with a repeated field.

fulcio.proto
mattmoor: In some of these cases, we may be stuck due to api compat.  I totally agree that we should shift towards a better API field for this, but perhaps we should tackle that separately?

haydentherapper: That's fine if we wanna tackle these in a later PR. An option could be using a one_of to wrap these two options (and same with my comment about certificate vs certificate_chain below). I believe adding a one_of later is not a breaking change, we should confirm that.

fulcio.proto
bobcallaway: I could resurrect the old `openapi.yaml` doc and make sure it is up to date

pkg/api/error.go
bobcallaway: yes, this will get deleted when I rip out the rest of the previous HTTP-centric code path

fulcio.proto
mattmoor: Oh, I was just talking about in the `rpc` comment.  Pretty sure you can generate swagger docs from these as well, although if I've done it it's been years...

cmd/app/http.go
```diff
@@ -0,0 +1,85 @@
...
+
+	if legacyGRPCServer != nil {
+		if err := legacy_gw.RegisterCAHandlerFromEndpoint(ctx, mux, legacyGRPCServer.grpcServerEndpoint, opts); err != nil {
+			log.Logger.Fatal(err)
+		}
+	}
+
+	// Limit request size
+	handler := api.WithMaxBytes(mux, maxMsgSize)
+	httpServerEndpoint := fmt.Sprintf("%s:%s", viper.GetString("http-host"), viper.GetString("http-port"))
```
nsmith5, line 64: Nit: Can we move the viper calls out and pass in the server endpoint as an arg? The more localized these viper calls are the better I think. Its makes it a little easier to reason about the configuration surface when it happens in one spot and data is passed down to components

config/ctfe/root.pem
```diff
@@ -1,13 +1,13 @@
...
+MTEwMDcxMzU2NTlaFw0zMTEwMDUxMzU2NThaMCoxFTATBgNVBAoTDHNpZ3N0b3Jl
+LmRldjERMA8GA1UEAxMIc2lnc3RvcmUwdjAQBgcqhkjOPQIBBgUrgQQAIgNiAAT7
+XeFT4rb3PQGwS4IajtLk3/OlnpgangaBclYpsYBr5i+4ynB07ceb3LP0OIOZdxex
+X69c5iVuyJRQ+Hz05yi+UF3uBWAlHpiS5sh0+H2GHE7SXrk1EC5m1Tr19L9gg92j
+YzBhMA4GA1UdDwEB/wQEAwIBBjAPBgNVHRMBAf8EBTADAQH/MB0GA1UdDgQWBBRY
+wB5fkUWlZql6zJChkyLQKsXF+jAfBgNVHSMEGDAWgBRYwB5fkUWlZql6zJChkyLQ
+KsXF+jAKBggqhkjOPQQDAwNpADBmAjEAj1nHeXZp+13NWBNa+EDsDP8G1WWg1tCM
+WP/WHPqpaVo0jhsweNFZgSs0eE7wYI4qAjEA2WB9ot98sIkoF3vZYdd3/VtWB5b9
+TNMea7Ix/stJ5TfcLLeABLE4BNJOsQ4vnBHJ
 -----END CERTIFICATE-----
```
nsmith5, line 13: Question: what is this change about? Just some test data or are we intending to change the root bundle in prod ctlog?

fulcio.proto
```diff
@@ -0,0 +1,120 @@
...
+     * Returns information about the current state of the transparency log
+     */
+    rpc CreateSigningCertificate(CreateSigningCertificateRequest) returns (SigningCertificate){
+        option (google.api.http) = {
+          post: "/api/v2/signingCert"
+          body: "*"
+        };
+    }
+    /**
+     * Returns the bundle of certificates that can be used to validate the signed certificate timestamp
```
nsmith5, line 39: This trust bundle also can also validate signatures from issued code-signing certs right? Can we add that to the comment here if so? It feels like an important aspect of what the trust bundle can be used for

fulcio.proto
```diff
@@ -0,0 +1,120 @@
...
+     */
+    Credentials credentials    = 1 [(google.api.field_behavior) = REQUIRED];
+    /*
+     * The public key to be stored in the requested certificate
+     */
+    PublicKey public_key       = 2 [(google.api.field_behavior) = REQUIRED];
+    /*
+     * Proof that the client possesses the private key; must be verifiable by provided public key
+     */
+    bytes proof_of_possession  = 3 [(google.api.field_behavior) = REQUIRED];
```
nsmith5, line 60: I like this name change from the generic "challenge". Its much more clear. Perhaps we could even explicitly note here that its the signature of the `sub` claim from the OIDC token? Note to self: we need to update the some docs to match I think once this gets merged.

fulcio.proto
```diff
@@ -0,0 +1,120 @@
...
+
+message CertificateChain {
+    /*
+     * The PEM-encoded root certificate chain for this Fulcio instance
+     */
+    repeated string certificates = 1;
+}
+
+enum PublicKeyAlgorithm {
+    PUBLIC_KEY_ALGORITHM_UNSPECIFIED = 0;
```
nsmith5, line 116: Question: Do we actually support anything outside ECDSA key-pairs right now? I've seen places around sigstore project where its sort of hard coded to ECDSA and wasn't sure if there is actually flexibility there

pkg/ctl/ctl_logging.go
```diff
@@ -25,15 +28,16 @@ type ctlLoggingClient struct {
 	logger *zap.SugaredLogger
```
nsmith5, line 28: You're pulling the logger from the context now so I think this struct field can / should be dropped.

```suggestion
```

pkg/ctl/interface.go
```diff
@@ -15,8 +15,12 @@
...
-import "github.com/sigstore/fulcio/pkg/ca"
+import (
+	"context"
+
+	"github.com/sigstore/fulcio/pkg/ca"
+)
 
 type Client interface {
-	AddChain(csc *ca.CodeSigningCertificate) (*CertChainResponse, error)
+	AddChain(ctx context.Context, csc *ca.CodeSigningCertificate) (*CertChainResponse, error)
```
nsmith5, line 25: Nice, thanks for wiring up the context into this interface 🥇

nsmith5 (approved):
Super awesome work ❤️ Really excited for this to land and mega kudos for not breaking the v1 API in the process. Extra special high-fives to wiring up the OIDC token into the context and passing context through to the CTlog client!

Feedback is just nits and questions for my own learning mostly

fulcio.proto
haydentherapper: Actually the trust bundle can't verify the SCT, it can only verify the code-signing certs. The CT log key doesn't come from Fulcio (and probably shouldn't, we don't want to couple the two).

nsmith5: Right ok! That makes more sense. We should amend this comment entirely then!

fulcio.proto
nsmith5: Right right, is there even a need for this ever? What is the API even supposed to do if there is a mismatch? The alg type is always available in the public key anyways right?

config/ctfe/root.pem
bobcallaway: it was to update the public key to permit submissions to the docker-compose managed CTFE instance to match that of the googleca v1 root. I should probably get rid of it and hard code the use of it to be something with the ephemeralCA or fileCA impl instead of defaulting to googleca.

pkg/ctl/ctl_logging.go
bobcallaway: yes, but without it the tests in `ctl_logging_test.go` won't work because they introspect whether or not the log messages actually made it.

pkg/api/legacy_server.go
```diff
@@ -0,0 +1,124 @@
...
+		ProofOfPossession: request.SignedEmailAddress, //lint:ignore SA1019 this is valid because we're converting from v1beta to v1 API
+	}
+
+	v2Response, err := l.v2Server.CreateSigningCertificate(ctx, &v2Request)
+	if err != nil {
+		return nil, errors.Wrap(err, "legacy handler")
+	}
+
+	// we need to return a HTTP 201 Created response code to be backward compliant
+	if err = grpc.SetHeader(ctx, metadata.Pairs("x-http-code", "201")); err != nil {
```
haydentherapper, line 84: nit: Make constants for the headers (this and SCT)

pkg/api/legacy_server.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil, errors.Wrap(err, "legacy handler")
+	}
+
+	// the SCT for the certificate needs to be returned in a HTTP response header
+	sctString := base64.StdEncoding.EncodeToString(v2Response.SignedCertificateTimestamp)
```
haydentherapper, line 89: Should you check if `sctString` is empty before setting the header?

.github/workflows/verify-k8s.yml
```diff
@@ -164,7 +164,7 @@ jobs:
                 automountServiceAccountToken: false
                 containers:
                 - name: check-oidc
-                  image: gcr.io/projectsigstore/cosign:v1.4.1
+                  image: ttl.sh/11fe4c92-b8a5-4870-8d16-b386801bf9ae/cosign@sha256:29104eddb8e8c773f92aa90f6089fbb34447002c288d3f81a1d79d391a17b108
```
lukehinds, line 167: I figure you know ttl.sh gets wiped periodically?

cmd/app/http.go
```diff
@@ -0,0 +1,115 @@
...
+	if err := gw.RegisterCAHandlerFromEndpoint(ctx, mux, grpcServer.grpcServerEndpoint, opts); err != nil {
+		log.Logger.Fatal(err)
+	}
+
+	if legacyGRPCServer != nil {
+		endpoint := fmt.Sprintf("unix:%v", legacyGRPCServer.grpcServerEndpoint)
+		if err := legacy_gw.RegisterCAHandlerFromEndpoint(ctx, mux, endpoint, opts); err != nil {
+			log.Logger.Fatal(err)
+		}
+	}
```
lukehinds, line 62: What determines if `legacyGRPCServer` is set , does this bubble up from config?

.github/workflows/verify-k8s.yml
bobcallaway: yes, this was just a debugging change, will be resetting this

cmd/app/http.go
bobcallaway: it doesn't bubble up from config (yet), but if we wanted to deprecate and turn that off we could just set it to nil

cmd/app/serve.go
```diff
@@ -64,12 +66,31 @@ func newServeCmd() *cobra.Command {
 	cmd.Flags().String("fileca-key", "", "Path to CA encrypted private key")
 	cmd.Flags().String("fileca-key-passwd", "", "Password to decrypt CA private key")
 	cmd.Flags().Bool("fileca-watch", true, "Watch filesystem for updates")
-	cmd.Flags().String("host", "0.0.0.0", "The host on which to serve requests")
-	cmd.Flags().String("port", "8080", "The port on which to serve requests")
+	cmd.Flags().String("host", "0.0.0.0", "The host on which to serve requests for HTTP; --http-host is alias")
```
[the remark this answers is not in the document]
bobcallaway: two lines below, there is a separate listener on a different port for grpc requests VS http requests.

cmd/app/serve.go
```diff
@@ -160,64 +181,42 @@ func runServeCmd(cmd *cobra.Command, args []string) {
...
-
-				inner.ServeHTTP(rw, r.WithContext(ctx))
-			})
-		}
-		handler = withDependencies(handler)
+	grpcServer, err := createGRPCServer(cfg, ctClient, baseca)
+	if err != nil {
+		log.Logger.Fatal(err)
+	}
+	//TODO: remove this reflection call for use with grpc_cli
```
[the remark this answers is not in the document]
bobcallaway: it will be removed, it is added to enable `grpc_cli ls`

config/logid.sh
```diff
@@ -48,5 +48,8 @@ if ! [[ -s /etc/config/ct_server.cfg ]]; then
 else
 	echo " found."
 	configid=`cat /etc/config/ct_server.cfg|grep log_id|awk ' { print $2 } '`
-	echo "Exisiting configuration uses log ID $configid, exiting"
+	echo "Existing configuration uses log ID $configid, exiting"
 fi
+curl -s --retry-connrefused --retry 10 http://fulcio-server:5555/api/v1/rootCert -o /etc/config/root.pem
```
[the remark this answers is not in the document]
bobcallaway: we'll see what merges first 😛

fulcio.proto
```diff
@@ -0,0 +1,122 @@
...
+option java_multiple_files = true;
+option java_outer_classname = "FulcioProto";
+
+service CA {
+    /**
+     * Returns information about the current state of the transparency log
+     */
+    rpc CreateSigningCertificate(CreateSigningCertificateRequest) returns (SigningCertificate){
+        option (google.api.http) = {
+          post: "/api/v2/signingCert"
```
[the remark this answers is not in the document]
bobcallaway: we can either:
* change the package to be v2
* change the url to be `v1` but be a different endpoint than `/api/v1/signingCert`, e.g. `/api/v1/certificate`

I'd probably vote for v2 but curious what others think?

fulcio.proto
```diff
@@ -0,0 +1,122 @@
...
+import "google/protobuf/empty.proto";
+
+option go_package = "github.com/sigstore/fulcio/pkg/generated/protobuf";
+option java_package = "dev.sigstore.fulcio.v1";
+option java_multiple_files = true;
+option java_outer_classname = "FulcioProto";
+
+service CA {
+    /**
+     * Returns information about the current state of the transparency log
```
bobcallaway, line 30: need to update this comment

fulcio.proto
```diff
@@ -0,0 +1,122 @@
...
+
+// (-- api-linter: core::0142::time-field-type=disabled
+//     aip.dev/not-precedent: SCT is defined in RFC6962 and we keep the name consistent for easier understanding. --)
+message SigningCertificate {
+    /*
+     * The signed certificate timestamp is a promise for including the certificate in
+     * a certificate transparency log. It can be "stapled" to verify the inclusion of
+     * a certificate in the log in an offline fashion.
+     */
+    bytes signed_certificate_timestamp = 1;
```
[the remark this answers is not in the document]
bobcallaway: I think the ability to run without a CT log is really there for developer/debugging convenience, not a deployment option. @lukehinds thoughts?

pkg/api/legacy_server.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/pkg/errors"
+	fulciogrpc "github.com/sigstore/fulcio/pkg/generated/protobuf"
+	"github.com/sigstore/fulcio/pkg/generated/protobuf/legacy"
+	"google.golang.org/genproto/googleapis/api/httpbody"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+const (
+	PEMCertificateChain = "application/pem-certificate-chain"
```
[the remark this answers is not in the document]
bobcallaway: calling the new API version over HTTP will return `application/json` with a JSON object marshalled based on the protobuf response message definition

pkg/api/legacy_server.go
bobcallaway: good call, i'd (incorrectly) assumed it would always be there.

cmd/app/http.go
lukehinds: gotcha, I was thinking there might be a flag/config leveraged that would decide on what runs (restAPI or gPRC (or even both)).

bobcallaway: hmm, interesting thought. let me noodle on that for a bit.

config/logid.sh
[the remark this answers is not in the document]
haydentherapper: So we'll need to extract all roots from the list of certificate chains. Is there a good CLI to do that? openssl probably

cmd/app/serve.go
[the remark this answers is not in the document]
haydentherapper: I was wondering what handles routing, since both will be accessible from https://fulcio.sigstore.dev

pkg/ctl/ctl_logging.go
nsmith5: If you're pushing the logger into the context I think its better to go all the way and remove all references. Remove the logger from the struct here, remove it from the `WithLogging()` signature and in the test just check that `log.Logger` sees the correct changes when testing.

In general I think explicit dependency injection is better than context stuffing and global vars, but if we're going one way or another I think its better to go all the way if that makes sense.

cmd/app/serve.go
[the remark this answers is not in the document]
bobcallaway: Nginx should be able to handle it

bobcallaway:
I need to add a few tests and rebase it

haydentherapper:
From conversation about SCTs, we should consider a one of with either a certificate and chain (where the SCT is embedded in a certificate), or a certificate, chain and SCT (a detached SCT for CAs that don't support embedded SCTs). This is good for clients so that they don't need to parse an x509 cert first to know if the SCT is embedded.

dlorenc:
Looks like tests are passing now, what's left?

nsmith5 (approved):
❤️ Looks good to me!

fulcio.proto
```diff
@@ -0,0 +1,123 @@
...
+    rpc CreateSigningCertificate(CreateSigningCertificateRequest) returns (SigningCertificate){
+        option (google.api.http) = {
+          post: "/api/v2/signingCert"
+          body: "*"
+        };
+    }
+    /**
+     * Returns the bundle of certificates that can be used to validate code signing certificates issued by this Fulcio instance
+     */
+    rpc GetTrustBundle(google.protobuf.Empty) returns (TrustBundle){
```
haydentherapper, line 41: space after (TrustBundle) {

fulcio.proto
```diff
@@ -0,0 +1,123 @@
...
+option go_package = "github.com/sigstore/fulcio/pkg/generated/protobuf";
+option java_package = "dev.sigstore.fulcio.v2";
+option java_multiple_files = true;
+option java_outer_classname = "FulcioProto";
+
+service CA {
+    /**
+     * Returns an X509 certificate created by the Fulcio certificate authority for the given request parameters
+     */
+    rpc CreateSigningCertificate(CreateSigningCertificateRequest) returns (SigningCertificate){
```
haydentherapper, line 32: space after (SigningCertificate) {

fulcio.proto
```diff
@@ -0,0 +1,123 @@
...
+import "google/protobuf/empty.proto";
+
+option go_package = "github.com/sigstore/fulcio/pkg/generated/protobuf";
+option java_package = "dev.sigstore.fulcio.v2";
+option java_multiple_files = true;
+option java_outer_classname = "FulcioProto";
+
+service CA {
+    /**
+     * Returns an X509 certificate created by the Fulcio certificate authority for the given request parameters
```
haydentherapper, line 30: nit: X.509

fulcio.proto
```diff
@@ -0,0 +1,123 @@
...
+
+// (-- api-linter: core::0142::time-field-type=disabled
+//     aip.dev/not-precedent: SCT is defined in RFC6962 and we keep the name consistent for easier understanding. --)
+message SigningCertificate {
+    /*
+     * The signed certificate timestamp is a promise for including the certificate in
+     * a certificate transparency log. It can be "stapled" to verify the inclusion of
+     * a certificate in the log in an offline fashion.
+     */
+    bytes signed_certificate_timestamp = 1;
```
haydentherapper, line 93: Can we update the comment to say this is optional, as the SCT could also be embedded? Also IIRC we had chatted about a one-of, do we want to do that? Or just leave this as an optional field?

config/logid.sh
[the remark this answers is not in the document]
haydentherapper: ^ (Adding existing comment to review)

pkg/api/client.go
```diff
@@ -39,6 +39,25 @@ type RootResponse struct {
 	ChainPEM []byte
 }
 
+type Key struct {
```
haydentherapper, line 42: Should this client be renamed to legacy client?

pkg/api/grpc_server.go
```diff
@@ -0,0 +1,204 @@
...
+	if err := cryptoutils.ValidatePubKey(publicKey); err != nil {
+		return nil, handleFulcioGRPCError(ctx, codes.InvalidArgument, err, insecurePublicKey)
+	}
+
+	subject, err := challenges.ExtractSubject(ctx, principal, publicKey, request.ProofOfPossession)
+	if err != nil {
+		return nil, handleFulcioGRPCError(ctx, codes.InvalidArgument, err, invalidSignature)
+	}
+
+	var csc *certauth.CodeSigningCertificate
```
haydentherapper, line 100: Merge conflict incoming: This section will have to be updated based on https://github.com/sigstore/fulcio/pull/507/files#diff-41ae48039acc4fc3fe73112a8e3af14bad4ab8691f9edec07804e0fd9604bec4

pkg/api/grpc_server.go
```diff
@@ -0,0 +1,204 @@
...
+		Chains: []*fulciogrpc.CertificateChain{{
+			Certificates: []string{string(root)},
+		}},
+	}, nil
+}
+
+func extractIssuer(token string) (string, error) {
+	parts := strings.Split(token, ".")
+	if len(parts) < 2 {
+		return "", fmt.Errorf("oidc: malformed jwt, expected 3 parts got %d", len(parts))
```
haydentherapper, line 174: Should we enforce that parts == 3, not just >= 2?

bobcallaway: That was a copy/paste of code from `ca.go` - I agree with the proposed change though.

pkg/api/grpc_server.go
bobcallaway: he who merges first wins!

dlorenc: I think this one will go in first right? #507 needs to wait until Cosign 1.8

dlorenc:
I'm happy with this whenever you're all ready for the merge!

pkg/api/grpc_server.go
haydentherapper: We can submit #507 now, the intermediate CA won't be rolled out until we make the changes to the production configs.

dlorenc (approved):

dlorenc (approved):

dlorenc:
Merge when ready!

dlorenc (approved):

PR #490, opened 2022-03-29
Add validation of public keys to prevent certifying weak keys

haydentherapper:
This adds checks for RSA, ECDSA and ED25519 (no-op) keys. This
will prevent Fulcio from signing certificates for the following:
* DSA keys
* RSA keys that are smaller than 2048 or larger than 4096
* RSA keys with weak primes
* ECDSA keys that aren't P256, P384, or P521 (P224 or custom curves)

#### Ticket Link

Fixes

#### Release Note

```release-note
Added validation of public keys
```


bobcallaway (approved):
lgtm, we should probably make the release note a bit more verbose

Issue #489, opened 2022-03-28
Fulcio as an intermediate CA


haydentherapper:
[Design doc](https://docs.google.com/document/d/1QLTMgjX7n74PyNpp3SIxsH7vbak8k-EoylttR25gBXM/edit#heading=h.dgwm7wkem51s)

From the work to support embedded SCTs, coupled with a desire to be able to scale beyond 25 QPS, I've written up a doc on the work to reconfigure the production instance of Fulcio as an intermediate CA.

tl;dr - An in-memory signing key is generated on each server initialization. Fulcio requests an intermediate CA certificate from the managed GCP root, periodically refreshing it. This work involves adding a new CA implementation, which would be a combination of the ephemeral and file CAs.

This has many benefits:
* No outbound calls to create and sign certificates
* Ability to scale up QPS significantly
* Minimizes the risk of signing key leakage
* Automates the process of fetching intermediate CA certs (less operational overhead)

k4leung4:
I found a similar issue, https://github.com/sigstore/fulcio/issues/28, should we close that one?

PR #494, opened 2022-03-29
Add intermediate CA implementation with in-memory signer

haydentherapper:

#### Ticket Link

Ref #489

#### Release Note

```release-note

```


haydentherapper:
cc @bobcallaway 
Looking for comments on the usage of goroutines and mutexes. I've tested locally, this is working.

Not in love with the amount of reader locks. The goroutine communication seems to be in line with the recommended style.

cmd/app/serve.go
```diff
@@ -53,7 +54,7 @@ func newServeCmd() *cobra.Command {
 
 	cmd.Flags().StringVarP(&serveCmdConfigFilePath, "config", "c", "", "config file containing all settings")
 	cmd.Flags().String("log_type", "dev", "logger type to use (dev/prod)")
-	cmd.Flags().String("ca", "", "googleca | pkcs11ca | fileca | ephemeralca (for testing)")
+	cmd.Flags().String("ca", "", "googleca | pkcs11ca | fileca | intermediateca | ephemeralca (for testing)")
```
bobcallaway, line 57: given the dependency on google CA service for `intermediateca`, does it make sense to have the name reflect that somehow?

pkg/ca/intermediateca/intermediateca.go
```diff
@@ -0,0 +1,215 @@
...
+		if err != nil {
+			return nil, err
+		}
+		if len(certs) != 1 {
+			return nil, errors.New("unexpected number of certificates returned")
+		}
+		parsedCerts = append(parsedCerts, certs[0])
+	}
+	log.Logger.Info("Current CA certificate chain:")
+	log.Logger.Info(pemCerts)
```
bobcallaway, line 122: combine these two calls?

pkg/ca/intermediateca/intermediateca.go
```diff
@@ -0,0 +1,215 @@
...
+		ica.RLock()
+		currentCert := ica.certs[0]
+		ica.RUnlock()
+
+		// Refresh certificate 7 days before expiration
+		if time.Until(currentCert.NotAfter) < (time.Hour * 24 * 7) {
+			certs, err := fetchCACertificate(ctx, ica.parent, ica.client, ica.signer)
+			if err != nil {
+				// An intermittent error is acceptable
+				log.Logger.Error(err)
```
bobcallaway, line 139: make this log message more informative since it would "randomly" appear without any request context

pkg/ca/intermediateca/intermediateca.go
```diff
@@ -0,0 +1,215 @@
...
+	return &ica, nil
+}
+
+func (ica *intermediateCA) CreateCertificate(ctx context.Context, challenge *challenges.ChallengeResult) (*ca.CodeSigningCertificate, error) {
+	cert, err := x509ca.MakeX509(challenge)
+	if err != nil {
+		return nil, err
+	}
+
+	parentCA, privateKey := ica.getX509KeyPair()
```
bobcallaway, line 192: since the locking is done within this function call, wouldn't it be possible that after returning from line 192, `ica.certs` would be updated in the background before getting to line 199? 

that would end up in a situation where the value of `ica.certs` changes from what was used to create the CSC vs the chain reported back in the response message to the caller. 

why not just change line 214 to return all `ica.certs`, and then use that local copy instead of going back to a read lock and reference `ica.certs` in the call to `ca.CreateCSCFromDER`?

nsmith5:
If scaled horizontally, this would result in several different in memory signers and different responses on `/api/v1/rootCert` or /api/v2/trustBundle` (soon to come) right?

haydentherapper:
Yep, it would. Intermediates would change, but the root would be the same.

haydentherapper:
Question away!

The certificate chain doesn't/shouldn't get verified based on what's returned by Rekor. (Separately, I am actually working on support for adding the cert chain, but there's a larger question around how to search when you don't know if an entry contains just the cert chain or the cert). The verification key/cert/chain comes from either CLI input or the OCI object (expect in one case for verify-blob, discussed below).

The ephemeral intermediate CA certificate needs to be persisted alongside the leaf certificate. This is done currently in the OCI signature. For other use-cases, it's up to the client to figure out how to store the cert.

Rekor shouldn't be used as storage for certificates. For example, the way that verify-blob fetches a certificate given an artifact isn't ideal. Rekor is meant as a witness to an event, and to find that proof, you need to know the contents of the Rekor entry (the signature, artifact hash, and verification key). Searching means you will accept someone else's proof and verification key.

As for the comment about `api/v1/rootCert`, lemme think on this. Could build an aggregator that calls all instances of Fulcio to build a trust bundle maybe.

haydentherapper:
High level comment: Intermediate CA certs are meant to be untrusted and just used for chain building from a leaf to a root. It's always up to the client to store these in the same way they store leaf certificates.

nsmith5:
Ahhhh ok I think I understand whats up here. Limme check my understanding:

- When requesting a code signing certificate the client will grab the leaf _and_ the intermediate(s) (basically everything it needs to chain up that final Fulcio root certificate)
- The leaf and intermediate will be stored alongside the signature

Verifiers can then check that this chains back to Fulcio root cert because they have the intermediates in hand already and then can also check the signature because they have the leaf in hand.

I think this makes sense, but I can foresee some modifications to the API and perhaps clients to make it function as expected. I recall then when I did the `fileca` stuff it serves up the intermediate and root chain on `/api/v1/rootCert`. It assumes there is just the one intermedate shared by all replicas and I'm not sure that the clients grabbed that intermedate and stored it along side the signature. That said, I test a lot with `sign-blob` and not OCI signature so maybe that already works as expected there.

haydentherapper:
Cosign will not need any updates besides figuring out `verify-blob`, I've wrapped up the work for verifying intermediates. External clients may need changes, hence why I want to do this before 1.0. Even with a static intermediate CA certificate, clients may need to be updated if they're not handling the certificate chain returned by Fulcio.

Thanks for pointing out replicas, this is something I'll need to consider. I do quite like the idea of an ephemeral key, but without coordination between the replicas, this might not be feasible.

pkg/ca/intermediateca/intermediateca.go
haydentherapper: Fixed in another PR!

PR #496, opened 2022-03-31
Add intermediate CA implementation with KMS-backed signer

### LINKED ISSUE #489: Fulcio as an intermediate CA

haydentherapper:
This CA implementation will use an on-disk certificate chain and a
remote KMS signer to sign certificates. There is validation on server
startup that the provided chain matches the provided key.

I've also added a utility to generate the intermediate certificate by
calling GCP CA Service. This will be used to set up Fulcio.

This also refactors the code to add an intermediate CA struct that
implements the common methods. This makes it simple to add new
intermediate CA types, with each only needing to provide a method to
fetch a signer and certificate chain.

Updated sigstore/sigstore to pull in the latest change to compare public keys.

Tested with:
`go run pkg/ca/intermediateca/fetch_ca_cert/fetch_ca_cert.go --kms-key="gcpkms://projects/<project>/locations/us-central1/keyRings/test-key-ring/cryptoKeys/ca-key/versions/1" --gcp-ca-parent="projects/<project>/locations/us-west1/caPools/<pool>" --output="chain.crt.pem"`

`go run main.go serve --port 5555 --ca kmsca --ct-log-url="" --kms-key="gcpkms://projects/<project>/locations/us-central1/keyRings/test-key-ring/cryptoKeys/ca-key/versions/1" --cert-chain-path="chain.crt.pem"`

Made a call using a script based on the load testing tool, got back a certificate chain.

#### Ticket Link

Fixes #489

#### Release Note

```release-note
Added a KMS-backed intermediate CA implementation
```


haydentherapper:
This is a good compromise between the various solutions. While an ephemeral key and certificate would be great, as discussed in https://github.com/sigstore/fulcio/pull/494, this doesn't work well with replicated instances. Moving to a KMS-backed signer gives us the ability to limit and audit signing key access. KMS will also enable us to have far more QPS than CA Service, and controlling the signer will let us implement embedded SCTs (PR coming right up!).

With this design, we'll issue intermediates for the same lifetime as the root. I'd like to include the intermediate in the TUF targets to ease client discovery. Also this simplifies the revocation story - If the KMS signer is compromised, we simply remove the intermediate certificate from the TUF target metadata and add a new one.

haydentherapper:
cc @nsmith5 (Can't add you as a reviewer for some reason)

haydentherapper:
Will investigate failing test - It's not failing locally for me.

haydentherapper:
One last remark: I haven't added tests for `kmca.go` because this relies on sigstore/sigstore's KMS library, and there's no mocks or fakes for this.

haydentherapper:
Failing test resolved, I switched to using a directory generated by the test library.

dlorenc:
Nice work!

haydentherapper:
Added tests for `kmsca.go`. Temporarily added a `replace` to my fork of sigstore/sigstore until https://github.com/sigstore/sigstore/pull/361 is submitted.

cmd/app/serve.go
```diff
@@ -53,7 +54,7 @@ func newServeCmd() *cobra.Command {
 
 	cmd.Flags().StringVarP(&serveCmdConfigFilePath, "config", "c", "", "config file containing all settings")
 	cmd.Flags().String("log_type", "dev", "logger type to use (dev/prod)")
-	cmd.Flags().String("ca", "", "googleca | pkcs11ca | fileca | ephemeralca (for testing)")
+	cmd.Flags().String("ca", "", "googleca | pkcs11ca | fileca | kmsca | ephemeralca (for testing)")
```
haydentherapper, line 57: Change to `signer`

cmd/app/serve.go
```diff
@@ -64,6 +65,8 @@ func newServeCmd() *cobra.Command {
 	cmd.Flags().String("fileca-key", "", "Path to CA encrypted private key")
 	cmd.Flags().String("fileca-key-passwd", "", "Password to decrypt CA private key")
 	cmd.Flags().Bool("fileca-watch", true, "Watch filesystem for updates")
+	cmd.Flags().String("kms-key", "", "KMS key resource path")
```
haydentherapper, line 68: Change to `kms-resource`, and give example

cmd/app/serve.go
```diff
@@ -64,6 +65,8 @@ func newServeCmd() *cobra.Command {
...
+	cmd.Flags().String("cert-chain-path", "", "Path to PEM-encoded CA certificate chain")
```
haydentherapper, line 69: Prefix with `kms`

priyawadhwa:
This looks great!

cmd/app/serve.go
haydentherapper: @bobcallaway Changing this affects a lot of files, will follow up in a later PR to make this change, as I don't want to break the build.

haydentherapper:
PR should be ready to go!

pkg/ca/kmsca/kmsca.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/sigstore/fulcio/pkg/ca"
+	"github.com/sigstore/fulcio/pkg/ca/intermediateca"
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
+	"github.com/sigstore/sigstore/pkg/signature/kms"
+
+	// Register the provider-specific plugins
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/aws"
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/azure"
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/gcp"
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/hashivault"
```
nsmith5, line 34: Nit: I prefer having imports with side effects (e.g loading database drivers and plugins etc) in `package main`. You can see an example of this behaviour [here in `distribution/distribution`](https://github.com/distribution/distribution/blob/main/cmd/registry/main.go). This makes it really easy for folks to build a slimmed down version of the registry (or fulcio in our case) by just customizing that `package main` themselves. 

Example

```golang
package main

import (
	"github.com/distribution/distribution/v3/registry"

	_ "github.com/distribution/distribution/v3/registry/auth/token"          # Only token auth
	_ "github.com/distribution/distribution/v3/registry/storage/driver/gcs"  # Only GCS storage
)

func main() {
	registry.RootCmd.Execute()
}
```

pkg/test/cert_utils.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	cert, err := x509.ParseCertificate(certBytes)
+	if err != nil {
+		return nil, err
+	}
+	return cert, nil
+}
+
+func GenerateRootCa() (*x509.Certificate, *ecdsa.PrivateKey, error) {
```
nsmith5, line 63:
```suggestion
func GenerateRootCA() (*x509.Certificate, *ecdsa.PrivateKey, error) {
```

To match go conventions of consistent case for acronyms https://github.com/golang/go/wiki/CodeReviewComments#initialisms

nsmith5: Similar comment about the other methods in this file

nsmith5 (commented):
Awesome! Really like the refactor into `pkg/intermediateca`

pkg/ca/intermediateca/fetch_ca_cert/fetch_ca_cert.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/sigstore/sigstore/pkg/signature/kms"
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/aws"
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/azure"
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/gcp"
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/hashivault"
+)
+
+/*
+To run:
+go run pkg/ca/intermediateca/update/fetch_ca_cert.go \
```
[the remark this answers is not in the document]
haydentherapper: Thanks, done!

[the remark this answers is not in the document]
haydentherapper: Done

pkg/test/cert_utils.go
haydentherapper: Done

pkg/ca/kmsca/kmsca.go
haydentherapper: Interesting point! I feel like there's a larger thing here - If you want to create a slimmed down Fulcio, you should be able to remove all signer types that aren't in use. Given this is the KMS type, I'd prefer that we load in all supported KMS backends in the same file.

nsmith5: Open to making that file be in `package main`? I think I see what you're saying in that we should be able to turn on and off all of "kmsca" with an import drop or configuration but I think its nice to also be able to say e.g "I was _kmsca_ and specifically only support gcpkms:// keys" and that still isn't possible if these driver imports happen under /pkg

haydentherapper: The benefit of moving these out to `main` would be if someone wanted to use `kmsca` as a library, correct? I don't expect that to be very common, so I'd rather simply install all of the supported KMS backends so if someone does use `kmsca` without `main`, they won't run into issues with missing KMS backends.

nsmith5: The benefit is in being able to have fine grain control of the `kmsca` backends supported by dropping them in main instead of using `kmsca` as a library, but I see your point about using `kmsca` as a lib. I'm alright to have coarse control of the entire kmsca backend in `package main`

pkg/ca/kmsca/kmsca.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+type kmsCA struct {
+	intermediateca.IntermediateCA
+}
+
+func NewKmsCA(ctx context.Context, kmsKey, certPath string) (ca.CertificateAuthority, error) {
```
nsmith5, line 41: Nit: same thing as the "CA" versus "Ca" comment before:

```suggestion
func NewKMSCA(ctx context.Context, kmsKey, certPath string) (ca.CertificateAuthority, error) {
```

nsmith5 (approved):
Awesome work :D Just a nit about Kms versus KMS, but otherwise LGTM

pkg/ca/kmsca/kmsca.go
haydentherapper: Done

haydentherapper:
Thanks! PR should be good to go unless there's any other comments.

dlorenc (approved):

PR #502, opened 2022-04-01
Use provided HTTP client instead when fetching root cert

### LINKED ISSUE #501: pkg/api/client.RootCert should use the provided *http.Client
**Description**

`RootCert` calls `http.Get`, which uses the default client:

https://github.com/sigstore/fulcio/blob/56c15d68101f8aa3d6376bd8da9993a27edcd0cd/pkg/api/client.go#L138

Contrast with `SigningCert`, which uses the provided `*http.Client` if one was provided:

https://github.com/sigstore/fulcio/blob/56c15d68101f8aa3d6376bd8da9993a27edcd0cd/pkg/api/client.go#L97

This means that if `WithTimeout` or `WithUserAgent` are called, these will be ignored when fetching the root cert. If future client options are added, these would likely also be ignored.

If this is unintentional, I can send a PR to fix it.

imjasonh:
Signed-off-by: Jason Hall <jasonhall@redhat.com>

Fixes #501 

```release-note
Client request to fetch root cert honors provided timeout and useragent
```

@vaikas

### DIFF
--- a/pkg/api/client.go
+++ b/pkg/api/client.go
@@ -135,7 +135,11 @@ func (c *client) RootCert() (*RootResponse, error) {
 	endpoint := *c.baseURL
 	endpoint.Path = path.Join(endpoint.Path, rootCertPath)
 
-	resp, err := http.Get(endpoint.String())
+	req, err := http.NewRequest(http.MethodGet, endpoint.String(), nil)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


vaikas (approved):
Thank you!

dlorenc (approved):

Issue #310, opened 2022-01-04
Consider supporting multiple CTLogs


vaikas:
**Description**
Currently we support one CTLog where SCT is then returned from. It might be useful in the future to support multiple ones. Besides the obvious need to modify code, just curious if we may want to consider at least allowing the API here:
https://github.com/sigstore/fulcio/blob/main/pkg/api/client.go#L31

Also copied here in case links shift:
```
type CertificateResponse struct {
	CertPEM  []byte
	ChainPEM []byte
	SCT      []byte
}
```

To be able to return multiple SCT, and we could further restrict the current api to say that it will only return 0 atm.

Brought this up 2022-01-04 meeting and it was deemed that we may want to look into this, from a quick look didn't see an issue, so here it is :)

There also has been some further discussion in the past so mayhaps we can gather those thoughts here, or pointers to them in case folks want to chime in / follow along.

haydentherapper:
https://github.com/sigstore/fulcio/pull/507 will solve this since an SCT list will be embedded in certificates.

vaikas:
Nice!

haydentherapper:
Closing issue, as we now support embedding SCT lists in certificates.

PR #507, opened 2022-04-06
Embed SCTs in issued certificates

### LINKED ISSUE #42: Figure out how to handle the CT log proofs in clients
The simplest way seems to be the "precertificate" method, here: https://letsencrypt.org/2018/04/04/sct-encoding.html

We request two certs from the CA - a "precertificate" and the real one. The precertificate is "poisoned" with a special bit set: https://github.com/google/certificate-transparency-go/blob/v1.1.1/x509/x509.go#L1208

Then that is included in the log. When that is finished, the SCT (signed cert timestamp) is embedded in the final cert. The certs are identical, except the poison bit is removed and the SCT is added.

That way clients can verify the cert is in the log without making any requests.

haydentherapper:
#### Summary

This adds support for embedding SCTs in certificates instead of
returning a header with a detached SCTs. This is done by implementing an
SCT interface for a signer. For example, GCP CA Service will not
support embedded SCTs, but KMS will.

This heavily leverages the Go CT library. I've removed the custom
client in favor of the CT library client, which includes more
verification and retry logic. Note that there's a TODO to include the
public key of the CT log in Fulcio so that the SCT is checked before
returning a response.

A certificate is signed twice, which adds an extra remote call to KMS.
The first certificate is added to the CT log via AddPreChain instead of
AddChain.

The Cosign client will need to be updated to support embedded SCTs.

#### Ticket Link

Fixes #42, #310

#### Release Note

```release-note
Added support for embedded SCTs for intermediate CA implementation. The CA will not return the SCT detached in a header. The client must verify the SCT using the SCT embedded in the certificate.
```


haydentherapper:
If you want to take a look, please only review the second commit! Waiting until we submit #496.

dlorenc:
> The Cosign client will need to be updated to support embedded SCTs.

How will this rollout work? Can cosign support both for awhile?

haydentherapper:
https://github.com/sigstore/fulcio/issues/42#issuecomment-1089743356 - Cosign will need to be updated to support both. It'll conditionally check if the SCT is embedded or in a header, and verify based on that. Then we can rollout this change, but it'll be a breaking change for anyone who hasn't updated Cosign. It's not possible for the current version of Cosign to successfully verify a certificate with an embedded SCT.

dlorenc:
Cool! OK, so we can change cosign to support both, get a release out with a version bump (1.8?), then deploy this change.

dlorenc:
Do we have an issue or PR in flight to add support to cosign for verifying both?

haydentherapper:
https://github.com/sigstore/cosign/issues/1730 for the issue, https://github.com/sigstore/cosign/pull/1731 for the WIP PR - The logic is all there, just working on testing (generating SCTs in tests is fun!)

haydentherapper:
Any other comments, or can this be submitted?

pkg/api/ca.go
```diff
@@ -192,21 +193,55 @@ func (a *api) signingCert(w http.ResponseWriter, req *http.Request) {
...
+			handleFulcioAPIError(w, req, http.StatusInternalServerError, err, failedToEnterCertInCTL)
+			return
+		}
+		csc, err = sctCa.IssueFinalCertificate(ctx, precert, sct)
+		if err != nil {
+			// if the error was due to invalid input in the request, return HTTP 400
+			if _, ok := err.(certauth.ValidationError); ok {
+				handleFulcioAPIError(w, req, http.StatusBadRequest, err, err.Error())
+				return
+			}
```
bobcallaway, line 240: nit: not sure this check is necessary here but it is functionally correct

pkg/ca/intermediateca/intermediateca.go
```diff
@@ -37,6 +42,88 @@ type IntermediateCA struct {
...
+	cert, err := x509ca.MakeX509(challenge)
+	if err != nil {
+		return nil, err
+	}
+
+	certChain, privateKey := ica.getX509KeyPair()
+
+	// Append poison extension
+	cert.ExtraExtensions = append(cert.ExtraExtensions, pkix.Extension{
+		Id:       asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 11129, 2, 4, 3},
```
bobcallaway, line 55: can you use https://github.com/google/certificate-transparency-go/blob/58784bdeda2ec153f22c23ff1d6cfcc6f1c7f5a2/x509/x509.go#L2234 instead of putting the ints inline here?

pkg/ca/intermediateca/intermediateca.go
```diff
@@ -37,6 +42,88 @@ type IntermediateCA struct {
...
+	listBytes, err := cttls.Marshal(list)
+	if err != nil {
+		return pkix.Extension{}, err
+	}
+	extBytes, err := asn1.Marshal(listBytes)
+	if err != nil {
+		return pkix.Extension{}, err
+	}
+	return pkix.Extension{
+		Id:    asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 11129, 2, 4, 2},
```
bobcallaway, line 97: same as above, use const from ct-lib: https://github.com/google/certificate-transparency-go/blob/58784bdeda2ec153f22c23ff1d6cfcc6f1c7f5a2/x509/x509.go#L2236

bobcallaway, line 97: same as above, use const from ct-lib: https://github.com/google/certificate-transparency-go/blob/58784bdeda2ec153f22c23ff1d6cfcc6f1c7f5a2/x509/x509.go#L2236

pkg/ca/intermediateca/intermediateca.go
```diff
@@ -37,6 +42,88 @@ type IntermediateCA struct {
...
+		Value: extBytes,
+	}, nil
+}
+
+func (ica *IntermediateCA) IssueFinalCertificate(ctx context.Context, precert *ca.CodeSigningPreCertificate, sct *ct.SignedCertificateTimestamp) (*ca.CodeSigningCertificate, error) {
+	// remove poison extension from precertificate.
```
bobcallaway, line 103: did you look at using https://github.com/google/certificate-transparency-go/blob/58784bdeda2ec153f22c23ff1d6cfcc6f1c7f5a2/x509/x509.go#L1208 with `precert.PreCert.RawTBSCertificate`? there is a fair bit of logic in the ct-library but your code probably covers most of the same stuff.

bobcallaway, line 103: did you look at using https://github.com/google/certificate-transparency-go/blob/58784bdeda2ec153f22c23ff1d6cfcc6f1c7f5a2/x509/x509.go#L1208 with `precert.PreCert.RawTBSCertificate`? there is a fair bit of logic in the ct-library but your code probably covers most of the same stuff.

bobcallaway (commented):
generally LGTM, a couple questions inline

cmd/app/serve.go
```diff
@@ -181,10 +192,17 @@ func runServeCmd(cmd *cobra.Command, args []string) {
...
-	var ctClient ctl.Client
+	var ctClient *ctclient.LogClient
 	if logURL := viper.GetString("ct-log-url"); logURL != "" {
-		ctClient = ctl.New(logURL)
-		ctClient = ctl.WithLogging(ctClient, log.Logger)
+		ctClient, err = ctclient.New(logURL,
+			&http.Client{Timeout: 30 * time.Second},
+			jsonclient.Options{
+				Logger: logAdaptor{logger: log.Logger},
+				// TODO: Add public key from CT Log for verification.
```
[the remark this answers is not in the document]
haydentherapper: No, this just provides a way of double-checking that the returned SCT is valid. It's not critical. I wanted to figure out the best way to pipe the CT log key to Fulcio, and was planning to in a later PR.

pkg/api/ca.go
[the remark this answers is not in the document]
haydentherapper: Thanks, I see this is just returned from `MakeX509`, so it only impacts the pre certificate issuance - Removing

pkg/ca/intermediateca/intermediateca.go
haydentherapper: Unforunately ct-go forked asn1 and x509 packages, so this `asn1` package does not match CT's `asn1` package. We could switch over to using the CT version of x509 and asn1, but it seems simpler to just copy the OIDs over.

pkg/ca/intermediateca/intermediateca.go
haydentherapper: Comment above

pkg/ca/intermediateca/intermediateca.go
haydentherapper: I had looked at it but I thought that it was cleaner to clearly show the differences between the pre and final certificate. This also clearly shows that extension order must be preserved.

pkg/ca/intermediateca/intermediateca.go
haydentherapper: Moving the OIDs into constants.

bobcallaway (approved):

haydentherapper:
We can deploy the main branch to prod, this only affects the intermediate CA implementation, which is not what’s rolled out in prod. The CA Service CA will continue to return detached SCTs.

haydentherapper:
I wanted to switch once we cut 1.8. Though if you prefer, I’m happy to do testing in staging first.

dlorenc:
This can merge now then @bobcallaway can rebase :)

Issue #43, opened 2021-03-15
Log rotation


dlorenc:
Figure out our strategy for log rotation. Ct logs get cut off and rotated each year.

lukehinds:
@dlorenc this sounds more relevant to the public instance, shall I migrate this issue over to public good instance?

haydentherapper:
cc @priyawadhwa @bobcallaway Something we may need to consider before GA.

haydentherapper:
Going to move conversation over to https://github.com/sigstore/fulcio/issues/589 and close as dup

PR #514, opened 2022-04-13
Add documentation for CT log

### LINKED ISSUE #510: Log sharding strategy for Fulcio
**Description**

We need to figure out the log sharding strategy. The simplest approach is to follow what other public CAs do, a new log every year accessible from a unique URL. Let's name shards using the year - ctfe.sigstore.dev/2022, ctfe.sigstore.dev/2023, etc. The current log is accessible at ctfe.sigstore.dev/test, so it'd be good to exercise a sharding before GA.

Each year, the TUF targets will be updated by appending the new log key. Cosign already supports verifying an SCT using a set of public keys. The log URL is not referenced anywhere in code, so only log monitors would need to be made aware of a new log shard URL.

haydentherapper:
Includes information about embedded SCTs and the log
sharding strategy.

#### Ticket Link

Fixes #510, #43

#### Release Note

```release-note

```

### DIFF
--- a/docs/ctlog.md
+++ b/docs/ctlog.md
@@ -1,0 +1,103 @@
+# Certificate Transparency Log Information
+
+## Summary
+
+Fulcio maintains a certificate transparency (CT) log, writing all issued certificates to the log.
+Users of Sigstore can verify via cryptographic proof that certificates are included in the log
+along with monitoring the log for inconsistencies. 
+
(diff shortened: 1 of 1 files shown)


haydentherapper:
Added details about log sharding - Let's create a new log each year, and keep the same signing key (unless there's a compromise at which point we'd freeze the log and rotate the key).

PR #526, opened 2022-04-15
Fix null pointer crash and incorrect error statuses

### LINKED ISSUE #525: gRPC legacy server does not return correct error statuses
**Description**

When the gRPC server returns an error to the legacy server, the errors are wrapped and returned, such as [here](https://github.com/sigstore/fulcio/blob/main/pkg/api/legacy_server.go#L82)

This results in the error status always being a 500, with the JSON error response code being '2' for unknown error. For example, `{"code":2,"message":"legacy handler: rpc error: code = InvalidArgument desc = The public key supplied in the request could not be parsed","details":[]}`

Looking into this now...

cc @bobcallaway 

fyi @cpanato @dlorenc, probably shouldn't cut a new release til we resolve this.

haydentherapper:
The gRPC and legacy servers did not check if the public
key was provided before accesses it, causing a null pointer exception if
no public key was provided in the request.

The errors from the gRPC server to the legacy server were being wrapped,
and the original error status was being lost.

#### Ticket Link

Fixes #525

#### Release Note

```release-note
Fixed null pointer crash and incorrect error statuses from gRPC server
```


bobcallaway:
Good catch; should we also add required field options in `fulcio_legacy.proto`?

haydentherapper:
Good idea, added!

dlorenc (approved):

PR #527, opened 2022-04-15
Add CSR support for key delivery and proof of possession 

haydentherapper:
#### Summary

[](https://github.com/haydentherapper)Many ecosystems and libraries have support for generating CSRs, but do
not have easy-to-use utilities for generating encoded public keys. A CSR
provides a simple way to send an encoded public key, along with a proof
of possession since CSRs are self-signed.

Existing clients do not have to change their behavior, as we will
continue to support providing a public key and signed challenge.

#### Ticket Link

Fixes

#### Release Note

```release-note

```


haydentherapper:
There are three options for the API changes:

1) (What's here) Add a field to the HTTP and gRPC API. All other fields must become optional.
2) Add a oneof to the gRPC layer, combining the PublicKey and Challenge fields into a single message, and adding the CSR as the other oneof field. The benefit is the gRPC API layer is well-defined. The downside is the HTTP layer does not support oneofs and we don't want to break the HTTP API, so we must still do (1) and make those fields optional.
3) Don't add fields to the API. Support the CSR being set in PublicKey.Content, parsing it server-side. Make the challenge field optional. The downside with this approach is now the public key field supports 3 different encodings.

Currently leaning towards (2), it's a cleaner gRPC API.

znewman01:
Looking forward to seeing this come through!

+1 to (2), I think the plan of "make gRPC API nice" is the right one and HTTP will always be a bit hacky regardless of this choice. I think (3) is a bad idea because the number of possible encodings may continue to grow.

haydentherapper:
FYI @di

fulcio.proto
```diff
@@ -49,16 +49,21 @@ message CreateSigningCertificateRequest {
...
-     *
-     * This is a currently a signature over the `sub` claim from the OIDC identity token
-     */
-    bytes proof_of_possession  = 3 [(google.api.field_behavior) = REQUIRED];
+    oneof key {
+        /*
+         * The public key to be stored in the requested certificate along with a signed
+         * challenge as proof of possession of the private key.
+         */
+        PublicKeyRequest public_key_request = 2 [(google.api.field_behavior) = REQUIRED];
```
znewman01, line 57: What is the format of this key?

Will we ever want to accept multiple formats of public key (we probably do)? Right now the implementation just tries to decode in a couple of different ways IIRC, but if we add more formats we probably want to specify that in the API. Should we separate out those code paths now? I think I'd rather have users say "this is PEM" or "this is DER" than just trying to parse that out ourselves.

I know this isn't an issue at all right now, but I could imagine it becoming one soon.

haydentherapper: Yep, currently we support PEM and DER encoded keys. My concern is a client might not be aware of the differences between PEM or DER encodings. Even that is a little ambiguous, we support a PKIX DER encoded public key or a PEM encoded version of a PKIX key (Not for example a PKCS1 encoding, typically just used for RSA). I'd rather the API be "Just give us a public key, however you've generated it" and let the server handle different formats. It also allows us to support more formats without requiring an update to the API.

This was one reason I thought about supporting the CSR as a "public key" type, but I think being able to annotate all fields as required is more beneficial.

I'm adding a comment to the API to specify the supported key formats.

haydentherapper: Updated

znewman01: > I don't think this list will continue to grow.

Famous last words :)

What if we wanted to, say, issue SSH certs? [Smallstep](https://smallstep.com/blog/use-ssh-certificates) runs an SSO-based CA for SSH certificates, which I could imagine being something we want to bring to Fulcio (spoiler: I have some need for this). The SSH key formats (there are at least 2, wire and the stuff that goes in your `authorized_keys` file) look quite different from the above.

I suppose we could make another field in the `oneof` for that, or we could extend the "try every key format" if-ladder. I suppose both of those are compatible the code in this PR—if you agree then feel free to resolve this.

haydentherapper: SSH is an entirely different beast in my opinion, and I think it'd be best to tackle it as a separate gRPC service. I've in the past tried to add an SSH CA to an X.509 based CA API, and the design gets muddy quickly. For one, the PKI is different - no chains, just a root issuing a leaf. There's support for different public key types too.

haydentherapper: We'd also need support for SSH cert extensions to restrict how the cert is used, etc. If this was something you were interested in, let's chat more, I'd be happy to sketch out an API for SSH.

dlorenc (approved):

PR #549, opened 2022-04-29
Fix key usage for issued certificates

### LINKED ISSUE #550: Incorrect key usage set for certificates generated by service
**Description**

Certificates are currently issued with the KeyCertSign key usage ([code](https://github.com/sigstore/fulcio/blob/0b972e84bfecaeaa5b75b269ba5268db47304fbd/pkg/ca/x509ca/common.go#L58)). This is incorrect - Certificates that have this key usage must also assert the `isCA` bit, as per RFC5280. KeyCertSign means that the certificate's public key is used to verify certificates whereas DigitalSignature means that the public key is used to verify signatures, which is what Fulcio's issued certificates are used for. 

This does not affect production, the certs fetched from CA Service had the correct key usage set.

haydentherapper:
The key usage should be Digital Signature, not Key Cert Sign. The latter
requires that isCA also be asserted. The certificate's key is only used
to verify signatures, not certificates.

Note that this does not affect the GCP CA.

#### Ticket Link

Fixes #550

#### Release Note

```release-note
Fixed key usage for issued certificates not from GCP CA Service
```

### DIFF
--- a/pkg/api/grpc_server_test.go
+++ b/pkg/api/grpc_server_test.go
@@ -1225,8 +1225,8 @@ func verifyResponse(resp *protobuf.SigningCertificate, eca *ephemeralca.Ephemera
 	if len(leafCert.SubjectKeyId) != 20 {
 		t.Fatalf("expected certificate subject key ID to be of length 20 bytes, got %d", len(leafCert.SubjectKeyId))
 	}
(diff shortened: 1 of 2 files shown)


haydentherapper:
@cpanato - We're going to want to cut a release with this fix before rolling out the new CA to production.

dlorenc (approved):

Issue #275, opened 2021-12-12
Improve `OIDCIssuer` abstraction


nsmith5:
**Description**

To add an new class of oidc issuer (eg. spiffe, github workflow...) to fulcio requires touching a lot of different areas of the code base. Here are all the locations I can think of

- [In this switch you need to check the issuer type and call the right challenge](https://github.com/sigstore/fulcio/blob/main/pkg/api/ca.go#L220-L231)
- [You need to add a new issuer type to the config package](https://github.com/sigstore/fulcio/blob/main/pkg/config/config.go#L155-L160)
- [You need to add a new challenge in the challenges package](https://github.com/sigstore/fulcio/blob/main/pkg/challenges/challenges.go#L140-L168)
- [You need to map your challenge response into x509 fields (URI vs. Email vs customer extensions)](https://github.com/sigstore/fulcio/blob/main/pkg/ca/x509ca/common.go#L54-L76)

I feel like these OIDC issuer types would be easier to test and probably easier to write if we make a interface for them all to obey and make the implementation details of each one abstracted from the rest of the code base.

Not sure what that interface would be. Has anyone gave this idea some thought already? It seems like at a high level, the issuers need to

- Match against the issuer from a token (i.e to check "is this the right issuer for this token?")
- Map its claims into x509 details
- Validate a token? (Not 100% on this one as it kinda of seems like we use the same OIDC machinery to do this for all providers really..)

mattmoor:
I definitely agree this is overdue for a refactor.  This is leftover from when there were 2-3 options, and just keeps growing.

I'd be happy to give feedback / discuss if you have ideas!

nsmith5:
Sweet, I'll give it a little thought and share a WIP PR or something if I brew up an idea.

nsmith5:
Drafted a proposal for this change here https://docs.google.com/document/d/1PKAou7wEmEob4VOLq6rFlx5maOM5hkJ3XffZizfwdFw/edit?usp=sharing. Looking for some eyeballs before diving in to implementing

haydentherapper:
@nsmith5 Do you have an estimate as to the amount of work left for this refactor? Context is determining when to make a new release of Fulcio.

nsmith5:
Yup remaining work is as follows:

- [ ] Make `FulcioConfig` obey the Issuer interface
- [ ] Stick that in an IssuerPool and replace the current `FulcioConfig` usage
[...]

PR #558, opened 2022-05-05
Add new `Issuer` and `Principal` abstractions

nsmith5:
#### Summary

Adds identity package with `Issuer` and `Principal` abstractions as well as an `IssuerPool` implementation that matches a token against a pool is issuers to authenticate. This work is part of a larger refactor of our identity issuer code as described [here](https://docs.google.com/document/d/1PKAou7wEmEob4VOLq6rFlx5maOM5hkJ3XffZizfwdFw/edit). 

Future work includes:

- [ ] Reimplementing our existing issuer abstractions to fit this interface including
- [ ] Changing our server to use the issuer pool instead of the current implementation
- [ ] Removing all the dead code once the new abstraction is in place

#### Ticket Link

Partially addresses #275 along with the other future works describe above

#### Release Note

```release-note
NONE
```

### DIFF
--- a/pkg/identity/issuer.go
+++ b/pkg/identity/issuer.go
@@ -1,0 +1,26 @@
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package identity
+
+import "context"
+
+type Issuer interface {
+	// Match checks if this issuer can authenticate tokens from a given issuer URL
+	Match(ctx context.Context, url string) bool
+
+	// Authenticate ID token and return Principal on success. The ID token's signature
+	// is verified in the call -- invalid signature must result in an error.
+	Authenticate(ctx context.Context, token string) (Principal, error)
+}
--- a/pkg/identity/issuerpool.go
+++ b/pkg/identity/issuerpool.go
@@ -1,0 +1,59 @@
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package identity
+
+import (
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+type IssuerPool []Issuer
+
+func (p IssuerPool) Authenticate(ctx context.Context, token string) (Principal, error) {
+	url, err := extractIssuerURL(token)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, issuer := range p {
+		if issuer.Match(ctx, url) {
+			return issuer.Authenticate(ctx, token)
+		}
+	}
+	return nil, fmt.Errorf("failed to match issuer URL %s from token with any configured providers", url)
+}
+
+func extractIssuerURL(token string) (string, error) {
+	parts := strings.Split(token, ".")
+	if len(parts) != 3 {
+		return "", fmt.Errorf("oidc: malformed jwt, expected 3 parts got %d", len(parts))
+	}
+
+	raw, err := base64.RawURLEncoding.DecodeString(parts[1])
+	if err != nil {
+		return "", fmt.Errorf("oidc: malformed jwt payload: %w", err)
+	}
+
+	var payload struct {
+		Issuer string `json:"iss"`
+	}
+	if err := json.Unmarshal(raw, &payload); err != nil {
+		return "", fmt.Errorf("oidc: failed to unmarshal claims: %w", err)
+	}
+	return payload.Issuer, nil
+}
--- a/pkg/identity/issuerpool_test.go
+++ b/pkg/identity/issuerpool_test.go
@@ -1,0 +1,213 @@
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package identity
+
+import (
+	"context"
+	"crypto/x509"
+	"errors"
+	"testing"
+)
+
+type testPrincipal struct {
+	name string
+}
+
+func (p testPrincipal) Name(ctx context.Context) string {
+	return p.name
+}
+
+func (p testPrincipal) Embed(ctx context.Context, cert *x509.Certificate) error {
+	return nil
+}
+
+type testIssuer struct {
+	match func(context.Context, string) bool
+	auth  func(context.Context, string) (Principal, error)
+}
+
+func (i testIssuer) Match(ctx context.Context, url string) bool {
+	return i.match(ctx, url)
+}
+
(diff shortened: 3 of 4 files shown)


pkg/identity/issuer.go
```diff
@@ -0,0 +1,25 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package identity
+
+import "context"
+
+type Issuer interface {
+	// Can this Issuer authenticate the given issuer URL?
+	Match(ctx context.Context, url string) bool
```
haydentherapper, line 21: Should this return an error or bool?

pkg/identity/issuer.go
```diff
@@ -0,0 +1,25 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package identity
+
+import "context"
+
+type Issuer interface {
+	// Can this Issuer authenticate the given issuer URL?
```
haydentherapper, line 20: nit: comments should be phrased as statements instead of questions

pkg/identity/principal.go
```diff
@@ -0,0 +1,29 @@
...
+
+package identity
+
+import (
+	"context"
+	"crypto/x509"
+)
+
+type Principal interface {
+	// URI, email etc of principal (usually matches `sub` from ID token)
```
haydentherapper, line 23: Nit: I think there’s enough cases where it doesn’t match the sub that I’d probably leave that out (for email, for kubernetes, for GitHub). Maybe point to existing documentation we have around this?

pkg/identity/issuer.go
```diff
@@ -0,0 +1,25 @@
...
+	Match(ctx context.Context, url string) bool
+
+	// Authenticate ID token and return Principal on success
+	Authenticate(ctx context.Context, token string) (Principal, error)
```
haydentherapper, line 24: To clarify, does this just extract the principal, or also do signature verification of the token? If both, we should mention that in the comment. Though I think verification of the token should happen early on in the request flow.

pkg/identity/issuerpool.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, issuer := range p {
+		if issuer.Match(ctx, url) {
+			return issuer.Authenticate(ctx, token)
+		}
+	}
+	return nil, fmt.Errorf("Failed to match issuer URL %s from token with any configured providers", url)
```
haydentherapper, line 38: nit: errors should start with a lowercase

dlorenc (approved):

pkg/identity/issuerpool.go
```diff
@@ -0,0 +1,59 @@
...
+		return "", fmt.Errorf("oidc: malformed jwt, expected 3 parts got %d", len(parts))
+	}
+
+	raw, err := base64.RawURLEncoding.DecodeString(parts[1])
+	if err != nil {
+		return "", fmt.Errorf("oidc: malformed jwt payload: %w", err)
+	}
+
+	var payload struct {
+		Issuer string `json:"iss"`
```
haydentherapper, line 53: How will this handle the issuer claim field in the config, where a custom issuer field is specified? Can we reuse the code that’s handling that in here?

haydentherapper: https://github.com/sigstore/fulcio/blob/f6016fda1ec6f1bd550699c7b8f53c2057ada174/pkg/oauthflow/oidc.go#L38

pkg/identity/principal.go
```diff
@@ -0,0 +1,29 @@
...
+	Name(ctx context.Context) string
+
+	// Embed all SubjectNameAlt and custom x509 extension information into
+	// certificate.
+	Embed(ctx context.Context, cert x509.Certificate) (x509.Certificate, error)
```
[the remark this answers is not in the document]
nsmith5: Is there an advantage to a pointer? I prefer a pure function over mutation where possible

pkg/identity/issuer.go
nsmith5: Bool seems more reasonable to me as its not an error if the URL doesn't match

pkg/identity/issuer.go
nsmith5: extraction and signature verification. Basically the call to `IssuerPool.Authenticate()` will replace this line right here https://github.com/sigstore/fulcio/blob/a08e6a73669b12255fd5dab6f7851f327dfdc56e/pkg/api/grpc_server.go#L73

pkg/identity/issuerpool.go
nsmith5: Mapping the custom issuer path actually happens _post_ authentication when embedding the that customer issuer path into the issuer extension in the certificate. In the concrete example of using a Dex token, this issuer pool needs to check the issuer URL matches the _Dex_ URL here and pass the token back the _Dex_ issuer instance for authentication. The implementation of that Dex specific issuer will take the upstream issuer claim (e.g in `$.federated.connection` or what ever that claim is) and map it into the correct x509 extension (our Fulcio Issuer OID)

nsmith5: So: importantly we need to use the actual `iss` claim in the issuer pool to match up with the correct issuer implementation. This will be a little more clear when I implement all those issuers again in the next PR.

pkg/identity/issuer.go
nsmith5: The comment does say the token is authenticated (so the signature is checked) and a `Principal` is returned. I can clarify that authenticating an ID token means checking the signature if that makes it more clear?

pkg/identity/issuer.go
haydentherapper: SG, just wanted to check that it didn't also parse the URL or do anything that could return an error.

pkg/identity/principal.go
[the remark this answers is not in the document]
haydentherapper: I prefer it because it signals that the initial certificate will get mutated. Otherwise, the test needs to demonstrate that nothing else was mutated but the fields passed in.
Also for performance - Certs can be large, so avoiding copying is preferred.

Unfortunately there's no central style guide for Go, so either is acceptable, but fwiw, when working with certs, I almost always see them passed by pointer.

pkg/identity/issuer.go
haydentherapper: SG on replacing that line. That'd be great to clarify that this verifies the signature too.

pkg/identity/issuerpool.go
haydentherapper: Thanks for the clarification! SG

pkg/identity/issuer.go
```diff
@@ -0,0 +1,25 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package identity
+
+import "context"
+
+type Issuer interface {
+	// Matche checks if this issuer can authenticate tokens from a given issuer URL
```
haydentherapper, line 20: nit: Match

pkg/identity/issuerpool_test.go
```diff
@@ -0,0 +1,213 @@
...
+		// Example principals
+		alice = dummyPrincipal{`alice`}
+		bob   = dummyPrincipal{`bob`}
+
+		// Example issuers
+		bobIfExampleCom = dummyIssuer{
+			match: func(_ context.Context, url string) bool {
+				return url == `example.com`
+			},
+			auth: func(_ context.Context, tok string) (Principal, error) {
```
haydentherapper, line 60: nit: unused variable `tok` can be replaced with `_`

pkg/identity/issuerpool_test.go
```diff
@@ -0,0 +1,213 @@
...
+		`empty pool should never authenticate`: {
+			Pool:    IssuerPool{},
+			Token:   exampleToken,
+			WantErr: true,
+		},
+		`match then reject all pool should never authenticate`: {
+			Pool:    IssuerPool{matchThenRejectAll},
+			Token:   exampleToken,
+			WantErr: true,
+		},
```
haydentherapper, line 140: Love the descriptive tests!

haydentherapper (approved):
Looks great! Just a couple nits and the other open comment on pointer vs val

pkg/identity/issuer.go
nsmith5: Nope no parsing or anything necessary. Even in the meta providers its just a regex or string match kinda thing so no errors there either. It just matches or doesn't match

pkg/identity/principal.go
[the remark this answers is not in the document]
nsmith5: Alright I'll switch the signature to a mutating one

pkg/identity/principal.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"context"
+	"crypto/x509"
+)
+
+type Principal interface {
+	// URI, email etc of principal
+	Name(ctx context.Context) string
+
+	// Embed all SubjectNameAlt and custom x509 extension information into
```
haydentherapper, line 26: nit: SubjectAltName

haydentherapper (approved):

haydentherapper (approved):

nsmith5:
Huh I can't merge. Maybe because I rerequested Dan's review? I don't have permissions to dismiss that unfortunately

haydentherapper:
Merged, it needs write permissions.

PR #562, opened 2022-05-09
Add client options testing

nsmith5:
#### Summary

Adds some tests for the `WithTimeout` and `WithUserAgent` options for the client. While this legacy client will be deprecated soon, these were some easy tests to write so why not? Not sure if we're keeping library API compat in the deprecation to GRPC, but if so perhaps these can be good tests to make sure we reimplement this when cutting over to the new client.

#### Release Note

```release-note
NONE
```

### DIFF
--- a/pkg/api/client_test.go
+++ b/pkg/api/client_test.go
@@ -1,0 +1,62 @@
+//
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


haydentherapper (approved):
gRPC generates its own clients, so I think the plan will be to just remove the API client entirely.

nsmith5:
Okie, We can drop this if GRPC cut over is happening soon. I was just looking around the repo for easy test coverage to add honestly 😆

pkg/api/client_test.go
```diff
@@ -0,0 +1,62 @@
+//
+// Copyright 2021 The Sigstore Authors.
```
cpanato, line 2:
```suggestion
// Copyright 2022 The Sigstore Authors.
```

nsmith5: Ah right we should update the boilerplate everywhere and the boilerplate test to match...

cpanato: no, this is added when you create the file first time. 

if is a new file you add the current year

PR #570, opened 2022-05-11
Consume `identity.Principal` in CA abstraction

nsmith5:
#### Summary

As part of #275, we'd like to move away from the `challenges.ChallengeResult` data structure and use `identity.Principal` and `identity.Issuer` instead. This PR changes the `ca.CertificateAuthority` to consume `identity.Principal` and a public key instead of `challenges.ChallengeResult`.

#### Ticket Link

Related to #275

#### Release Note

```release-note
NONE
```


nsmith5:
@znewman01 for refactoring mentorship eyeballs 🙏🏻

pkg/challenges/challenges.go
```diff
@@ -69,6 +70,78 @@ type ChallengeResult struct {
 	AdditionalInfo map[AdditionalInfo]string
 }
```
haydentherapper, line 71: nit: Worth adding a comment that says this implements `identity.Principal`?

pkg/api/grpc_server.go
```diff
@@ -115,7 +115,7 @@ func (g *grpcCAServer) CreateSigningCertificate(ctx context.Context, request *fu
 	// For CAs that do not support embedded SCTs or if the CT log is not configured
 	if sctCa, ok := g.ca.(certauth.EmbeddedSCTCA); !ok || g.ct == nil {
 		// currently configured CA doesn't support pre-certificate flow required to embed SCT in final certificate
-		csc, err = g.ca.CreateCertificate(ctx, subject)
+		csc, err = g.ca.CreateCertificate(ctx, subject, publicKey)
```
haydentherapper, line 118: I assume later in the refactor things will get shuffled around such that `ChallengeResult` doesn't contain a public key (since it's now unused)?

pkg/challenges/challenges.go
```diff
@@ -69,6 +70,78 @@ type ChallengeResult struct {
...
+	}
+
+	exts := x509ca.Extensions{
+		Issuer: cr.Issuer,
+	}
+	if cr.TypeVal == GithubWorkflowValue {
+		var ok bool
+		exts.GithubWorkflowTrigger, ok = cr.AdditionalInfo[GithubWorkflowTrigger]
+		if !ok {
+			return errors.New("x509ca: github workflow missing trigger claim")
```
haydentherapper, line 116: can you drop "x509ca" from each of these error messages? they don't have any value

pkg/challenges/challenges.go
```diff
@@ -69,6 +70,78 @@ type ChallengeResult struct {
 	AdditionalInfo map[AdditionalInfo]string
 }
 
+func (cr *ChallengeResult) Name(context.Context) string {
+	return cr.Value
```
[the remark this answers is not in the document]
haydentherapper: Proof of possession is checked in `challenges.ExtractSubject`, correct?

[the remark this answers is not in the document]
haydentherapper: `preChallengeResult.Subject` is meant to be an internal struct just used for ExtractSubject, so `CheckSignature(publicKey, challenge, result.Subject)` where `result` is a `preChallengeResult` seems correct to me.

pkg/api/grpc_server.go
nsmith5: Yup! Just removed it now

pkg/challenges/challenges.go
nsmith5: Pushing a recent change that makes that a lot more clear I think

pkg/challenges/challenges.go
```diff
@@ -61,21 +63,87 @@ const (
 )
 
 type ChallengeResult struct {
```
haydentherapper, line 65: can this now be made private?

pkg/challenges/challenges.go
haydentherapper: Yep, I think this looks clear!

pkg/challenges/challenges.go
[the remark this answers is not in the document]
nsmith5: The end state should look something like this I think:

```golang
// Inside of our api.CreateSigningCertificate handler

// ip is an identity.IssuerPool and we authenticate and extract the principal first
principal, err := ip.Authenticate(ctx, rawIDToken)
if err != nil {
   // handle err
}

if csr != nil {
  err = csr.CheckSignature()
  if err != nil {
    // handle errpr
  }
} else {
  // Proof of possession path
  err = challenges.CheckSignature(publicKey, proofOfPosession, principal.Name(ctx))
  if err != nil {
    // handle err
  }
}
```

So basically: `ExtractSubject` will go away and we'll be using `CheckSignature` directly. Once this current change is in I can actually do that part so that there is a clear difference between parsing id token -> Principal and checking proof of possession.

pkg/challenges/challenges.go
```diff
@@ -61,21 +63,87 @@ const (
...
 	// Extra information from the token that can be added to extensions.
 	AdditionalInfo map[AdditionalInfo]string
+
+	// subject or email from the id token. This must be the thing
+	// signed in the proof of possession!
+	subject string
+}
+
+func (cr *ChallengeResult) Name(context.Context) string {
+	return cr.subject
```
haydentherapper, line 78: I think this is an issue. Since subject and Value aren't the same, we're changing what's included in the certificate.

Why don't we stick with this being `cr.Value`, and stick with returning `ChallengeResult` for these internal methods so that you have access to `subject`?

haydentherapper: I'd kinda hope a test breaks too.

pkg/challenges/challenges.go
nsmith5: Yup! Want me to do that now? (This is slowly turning into the private implementation of an uber `Issuer` that we can break into seperate implementation)

pkg/challenges/challenges.go
haydentherapper: Sorry, looked over the code again, I see we're not using `Name` where I thought. I guess that's another question, what do we need `Name` for?

pkg/challenges/challenges.go
haydentherapper: I'd like to avoid making this too much larger, it's hard to follow each of these changes and make sure nothing is breaking. If it's just making it private, I think that's fine, but if it touches a bunch of other files, let's hold off.

pkg/challenges/challenges.go
haydentherapper: Yep, I agree that it didn't. I'm a little unclear about what Name is meant for, besides the proof of possession check. Is there a clearer function name than `Name` to distinguish between what's currently called `Value` (what goes into the certificate) and `subject` (The OIDC token subject/email)?

haydentherapper: If there isn't another planned use for `Name`, maybe we remove it? YAGNI unless it's going to be used elsewhere.

pkg/challenges/challenges.go
nsmith5: Yeah sorry this one is big. I'll just make it private in a follow up

pkg/challenges/challenges.go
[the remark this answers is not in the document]
nsmith5: Sorry, might have pre-maturely resolved this convo. Feels related to the other thread below as well. Is the intent to use `Name()` to deal the proof of possession checking for a principal clear above?

pkg/challenges/challenges.go
haydentherapper: Can we add a comment to `Value`, something like - "The identity that will be used as the subject alternative name of the certificate"

SG about changing in the next iteration.

pkg/challenges/challenges.go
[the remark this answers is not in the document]
haydentherapper: Yep, makes sense, this question just arose from this being an intermediate state. Sounds good to push that logic up to the handler.

pkg/challenges/challenges.go
nsmith5: Yup absolutely. I'll clarify intended use on ChallengeResult right now and clarify the usage on the interface of Principal in an iteration

haydentherapper (approved):
Looks good!

pkg/ca/googleca/v1/googleca.go
```diff
@@ -179,7 +180,7 @@ func (c *CertAuthorityService) CreateCertificate(ctx context.Context, subj *chal
 		return nil, ca.ValidationError(err)
 	}
 
-	logger.Infof("requesting cert from %s for %v", c.parent, subj.Value)
+	logger.Infof("requesting cert from %s for %v", c.parent, principal.Name(ctx))
```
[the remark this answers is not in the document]
nsmith5: Oh ew. Yeah I didn't even notice that when I was touching this line. I'll submit a new PR to nerf it

haydentherapper (approved):
Leaving open for a bit if @znewman01 has any comments

znewman01:
> Leaving open for a bit if @znewman01 has any comments

LGTM, thanks!

PR #579, opened 2022-05-12
identity: improve the documentation for Principal.Name()

nsmith5:
#### Summary
 
Clarifies that Name _must_ match the proof of possession.

@haydentherapper let me know if you can thing of a better name for the method. Happy to add that in here. Thought I'd at least improve the docs while I remembered to.

### DIFF
--- a/pkg/identity/principal.go
+++ b/pkg/identity/principal.go
@@ -20,7 +20,8 @@ import (
 )
 
 type Principal interface {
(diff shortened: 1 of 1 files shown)


haydentherapper (approved):
Thanks for the added documentation. I think this is good for now - Naming is hard :)

Ideally I'd like to call this `Subject`, but it's not accurate in terms of OIDC.

PR #580, opened 2022-05-12
Refactor challenge verification

nsmith5:
#### Summary

The challenge verification logic is sort of mixed up with the OIDC token parsing at the moment. This work makes two changes to this logic to make it a little easier to understand (there is one commit per change so you can read one at a time if desired)

- Break up the OIDC parsing logic and proof of possession verification in `ExtractSubject` into two seperate things
- Group all the logic for verifying a CSR together and all the logic for verify a signed subject proof of possession together. Make a really clear branch in logic between these two completely different challenges.

### DIFF
--- a/pkg/api/grpc_server.go
+++ b/pkg/api/grpc_server.go
@@ -70,9 +70,17 @@ func (g *grpcCAServer) CreateSigningCertificate(ctx context.Context, request *fu
 		}
 	}
 
-	principal, err := authorize(ctx, token)
+	// Authenticate OIDC ID token by checking signature
+	idtoken, err := authorize(ctx, token)
 	if err != nil {
 		return nil, handleFulcioGRPCError(ctx, codes.Unauthenticated, err, invalidCredentials)
+	}
+	// Parse authenticated ID token into principal
+	// TODO:(nsmith5) replace this and authorize call above with
+	// just identity.IssuerPool.Authenticate()
+	principal, err := challenges.PrincipalFromIDToken(ctx, idtoken)
+	if err != nil {
+		return nil, handleFulcioGRPCError(ctx, codes.InvalidArgument, err, invalidIdentityToken)
 	}
 
 	// optionally parse CSR
@@ -103,10 +111,15 @@ func (g *grpcCAServer) CreateSigningCertificate(ctx context.Context, request *fu
 		return nil, handleFulcioGRPCError(ctx, codes.InvalidArgument, err, insecurePublicKey)
 	}
 
-	// verify challenge
-	subject, err := challenges.ExtractSubject(ctx, principal, publicKey, csr, proofOfPossession)
-	if err != nil {
-		return nil, handleFulcioGRPCError(ctx, codes.InvalidArgument, err, invalidSignature)
+	// verify the challenge
+	if csr != nil {
+		if err := csr.CheckSignature(); err != nil {
+			return nil, handleFulcioGRPCError(ctx, codes.InvalidArgument, err, invalidSignature)
+		}
+	} else {
+		if err := challenges.CheckSignature(publicKey, proofOfPossession, principal.Name(ctx)); err != nil {
+			return nil, handleFulcioGRPCError(ctx, codes.InvalidArgument, err, invalidSignature)
+		}
 	}
 
 	var csc *certauth.CodeSigningCertificate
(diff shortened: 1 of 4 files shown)


haydentherapper:
I’ve got to prep for a presentation, so I’ll take a look at this first thing Monday.

pkg/api/grpc_server.go
```diff
@@ -70,7 +70,15 @@ func (g *grpcCAServer) CreateSigningCertificate(ctx context.Context, request *fu
...
+	idtoken, err := authorize(ctx, token)
+	if err != nil {
+		return nil, handleFulcioGRPCError(ctx, codes.Unauthenticated, err, invalidCredentials)
+	}
+	// Parse authenticated ID token into principal
+	// TODO:(nsmith5) replace this and authorize call above with
+	// just identity.IssuerPool.Authenticate()
+	principal, err := challenges.PrincipalFromIDToken(ctx, idtoken)
 	if err != nil {
 		return nil, handleFulcioGRPCError(ctx, codes.Unauthenticated, err, invalidCredentials)
```
haydentherapper, line 83: To maintain the same status code, we'll want to return `InvalidArgument` here - The verification of the token signature should have already happened in `authorize`.

Another error message would also be good, something like: `invalidIdentityToken = "There was an error processing the identity token."`

pkg/api/grpc_server.go
```diff
@@ -103,10 +111,16 @@ func (g *grpcCAServer) CreateSigningCertificate(ctx context.Context, request *fu
...
 	}
 
-	// verify challenge
-	subject, err := challenges.ExtractSubject(ctx, principal, publicKey, csr, proofOfPossession)
-	if err != nil {
-		return nil, handleFulcioGRPCError(ctx, codes.InvalidArgument, err, invalidSignature)
+	// verify the challenge
+	if csr != nil {
+		err = csr.CheckSignature()
+		if err != nil {
```
haydentherapper, line 104: nit: inline err initialization to match what's below

pkg/api/grpc_server.go
```diff
@@ -83,41 +83,51 @@ func (g *grpcCAServer) CreateSigningCertificate(ctx context.Context, request *fu
...
 	}
 
-	// optionally parse CSR
-	var csr *x509.CertificateRequest
+	var publicKey crypto.PublicKey
+	// Verify caller is in possession of their private key and extract
+	// public key from request.
 	if len(request.GetCertificateSigningRequest()) > 0 {
-		csr, err = cryptoutils.ParseCSR(request.GetCertificateSigningRequest())
+		// Option 1: If they gave us a CSR we can verify that to check
```
haydentherapper, line 90: nit: I don't think the comment is adding anything meaningful here.

pkg/api/grpc_server.go
```diff
@@ -83,41 +83,51 @@ func (g *grpcCAServer) CreateSigningCertificate(ctx context.Context, request *fu
...
-
-	// verify the challenge
-	if csr != nil {
+		// Check signature of CSR
 		err = csr.CheckSignature()
 		if err != nil {
 			return nil, handleFulcioGRPCError(ctx, codes.InvalidArgument, err, invalidSignature)
 		}
 	} else {
+		// Option 2: If no CSR, we need to check that they signed something with theirprivate key
```
haydentherapper, line 108: nit: Comments are clearer as imperative, "Check the signature for proof of possession of a private key"

pkg/api/grpc_server.go
nsmith5: Ah ok gotcha! Thanks!

pkg/api/grpc_server.go
```diff
@@ -70,43 +70,65 @@ func (g *grpcCAServer) CreateSigningCertificate(ctx context.Context, request *fu
...
 
-	// validate public key, checking for weak key parameters
-	if err := cryptoutils.ValidatePubKey(publicKey); err != nil {
-		return nil, handleFulcioGRPCError(ctx, codes.InvalidArgument, err, insecurePublicKey)
-	}
+		if err := csr.CheckSignature(); err != nil {
+			return nil, handleFulcioGRPCError(ctx, codes.InvalidArgument, err, invalidSignature)
+		}
+	} else {
+		// Option 2: Check the signature for proof of posession of a private key
```
haydentherapper, line 106: Possession, causing the lint failure

PR #590, opened 2022-05-17
Move domain validation checks for URI/Username to service startup

haydentherapper:
These checks were done once per request. These instead can be static
checks done on service startup. This enforces that the Username and URI
configs appropriately set the SubjectDomain and IssuerUrl fields, with
no scheme just for the SubjectDomain field for Username.

#### Ticket Link

Fixes

#### Release Note

```release-note

```


nsmith5 (approved):
😍 Heck yes. Failing fast on bad config is a way better operator experience than logging errors. Thanks for moving all the URI related validation early ❤️

bobcallaway (approved):
lgtm

PR #599, opened 2022-05-20
Move github principal to its own package

nsmith5:
Remove github workflow logic from challenge result and put it in its own
package.


pkg/identity/github/principal.go
```diff
@@ -0,0 +1,109 @@
...
+		ref:        claims.Ref,
+	}, nil
+}
+
+func (w workflowPrincipal) Name(ctx context.Context) string {
+	return w.subject
+}
+
+func (w workflowPrincipal) Embed(ctx context.Context, cert *x509.Certificate) error {
+	// Set workflow URL to SubjectNameAlt on certificate
```
haydentherapper, line 88: nit: SubjectAltName

pkg/identity/github/principal.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/coreos/go-oidc/v3/oidc"
+	"github.com/sigstore/fulcio/pkg/ca/x509ca"
+	"github.com/sigstore/fulcio/pkg/identity"
+)
+
+type workflowPrincipal struct {
+	subject    string
+	issuer     string
+	url        string
```
haydentherapper, line 31: Can you note that the `url` field is what will be used for the subject of the certificate, and that other claims from the the token?

pkg/identity/github/principal.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (w workflowPrincipal) Embed(ctx context.Context, cert *x509.Certificate) error {
+	// Set workflow URL to SubjectNameAlt on certificate
+	parsed, err := url.Parse(w.url)
+	if err != nil {
+		return err
+	}
+	cert.URIs = []*url.URL{parsed}
+
+	// Embed additional information into customer extensions
```
haydentherapper, line 95: nit: custom

pkg/identity/github/principal_test.go
```diff
@@ -0,0 +1,280 @@
...
+				"event_name": "push",
+				"exp":        0,
+				"iss":        "https://token.actions.githubusercontent.com",
+				"ref":        "refs/heads/main",
+				"repository": "sigstore/fulcio",
+				"sha":        "bf96275471e83ff04ce5c8eb515c04a75d43f854",
+				"sub":        "repo:sigstore/fulcio:ref:refs/heads/main",
+				"workflow":   "foo",
+			},
+			WantErr: true,
```
haydentherapper, line 76: What do you think about also checking the contents of the error message to check that the missing field is mentioned in the error message? This would be helpful to make sure that the test is constructed with the correct missing field.

pkg/identity/github/principal_test.go
```diff
@@ -0,0 +1,280 @@
...
+					t.Error(err)
+				}
+				return
+			}
+			if err == nil && test.WantErr {
+				t.Error("expected error")
+				return
+			}
+
+			gotName := untyped.Name(context.TODO())
```
haydentherapper, line 174: nit: Should this be moved to its own test, since it's testing a different function?

haydentherapper (changes requested):

haydentherapper (approved):

pkg/challenges/challenges.go
```diff
@@ -243,35 +202,6 @@ func kubernetes(ctx context.Context, principal *oidc.IDToken) (identity.Principa
...
-	if err != nil {
-		return nil, err
-	}
-
-	cfg, ok := config.FromContext(ctx).GetIssuer(principal.Issuer)
-	if !ok {
-		return nil, errors.New("invalid configuration for OIDC ID Token issuer")
-	}
-
-	issuer, err := oauthflow.IssuerFromIDToken(principal, cfg.IssuerClaim)
```
haydentherapper, line 261: Sorry, just saw this - Don't we still need this? `IssuerFromIDToken` pulls the issuer either from the issuer field or from a field that's specified in the configuration?

Actually, I think the bug might be here - https://github.com/sigstore/fulcio/blob/e1fe2f85ba4dc3277f1c4585e9d3049f72246ee3/pkg/challenges/challenges.go#L413
If you want to frontload fetching the correct issuer, you'll need to do that in `PrincipalFromIDToken`

haydentherapper (changes requested):

pkg/challenges/challenges.go
haydentherapper: Created https://github.com/sigstore/fulcio/pull/605 to test `IssuerClaim`

nsmith5: Ah yes I didn't noticed I'd make a change to the business logic here. I've done the same change in SPIFFE PR and justified its removal here https://github.com/sigstore/fulcio/pull/604#issuecomment-1134998659.

The change is ignoring `FulcioConfig.IssuerClaim` for github actions workflows. This is likely for the best as I can only see this claim being abused / misconfigured for a Github workflow. The claim is only helpful in the case of a federating identity provider like Dex for which the issuer is the federated identity provider (e.g `oauth2.sigstore.dev`) and we'd prefer to place the upstream provider (e.g `https://accounts.google.com`) in the certificate. 

I don't think we should support this form of federating identity provider for github workflow principals. We can add some logic in the config parsing to panic should such an invalid configuration is provided.

If you wish I can break up this change in business logic from the refactor

haydentherapper: Makes sense - I think this is something we would only support if we either control the OIDC provider or if we have a list of configured federated issuers too, but the latter isn't something we need now.

+1, can you split this off into its own PR? Makes it easier if we need to rollback or if we want to look at the history of changes.

nsmith5: Ok sounds good. I'll front-load a change to drop this from Github workflows, SPIFFE and Kubernetes type principals. Should it be dropped from username and URI principals? I don't really understand how those ones are used or will be used right now

haydentherapper: Yea, let's drop it from those too, since there's not an immediate need.

nsmith5: Opened #606. Will rebase on that once merged

haydentherapper (approved):

PR #606, opened 2022-05-23
Restict issuer claim mapping to email issuers

nsmith5:
#### Summary

Using a claim other than "iss" to set the issuer on a certificate only
makes sense for federating identity issuers like Dex. This restricts the
issuers configued to this setting to email type issuers.

cc @haydentherapper 

#### Release Note

```release-note
* Restricted usage of "IssuerClaim" to configure which claim is used to set the issuer extension to email type issuers.
```


haydentherapper (approved):

bobcallaway (approved):
code is fine, feels like a pretty minor issue IMO

PR #616, opened 2022-05-26
Split pkg/server from pkg/api

mtrmac:
Signed-off-by: Miloslav Trmač <mitr@redhat.com>

#### Summary
Split pkg/server from pkg/api

#### Ticket Link
N/A

#### Release Note
```release-note
External users of pkg/api that refer to the server code now need to use pkg/server.
```

---

API clients using `pkg/api.NewClient`, like `github.com/sigstore/cosign`, currently end up depending on all the implementation details of the servers, unnecessarily increasing their binary size and supply-chain attack surface.

Instead, move the server code to a new pkg/server (assuming there are no external users of that functionality), and leave pkg/api as a package only providing a client (as it is used in practice, and documented in various places in this repo).

The move includes `pkg/api/version.go`, because it is only useful with a Go build that specifies the various values at build time, which is unlikely to be the case for any third-party importer of that package.

This shaves over 500 kB from a simple user of `pkg/api.NewClient`.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -32,7 +32,7 @@ issues:
     - errcheck
     - gosec
   # the following section is due to the legacy API being deprecated
-  - path: pkg/api/legacy_server.go
+  - path: pkg/server/legacy_server.go
     linters:
     - staticcheck
     text: SA1019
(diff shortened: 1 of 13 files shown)


dlorenc (approved):

haydentherapper:
LGTM, also worth noting that we'll be recommending the use of the V2 API, and we'll be deprecating the legacy client.

PR #619, opened 2022-05-26
Move kubernetes principal to package

nsmith5:


haydentherapper (dismissed):

haydentherapper (approved):

pkg/identity/kubernetes/principal.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil, err
+	}
+	return principal{
+		subject: token.Subject,
+		issuer:  token.Issuer,
+		uri:     k8sURI,
+	}, nil
+}
+
+func (p principal) Name(context.Context) string {
```
haydentherapper, line 51: Sorry, just noticed this - This (and below) should be a pointer receiver, to avoid the copy of the principal. Even though the receiver is not mutated, it saves a copy.

haydentherapper (changes requested):

pkg/identity/kubernetes/principal.go
nsmith5: I think that communicating to the reader that the values are immutable is more important that any performance overhead here.

That said, if performance is a concern I've benchmarked both here: https://github.com/nsmith5/receiverbench. The allocations are the same in each case and the speed looks identitical if not slightly faster for the value receiver. For such a simply example its makes sense to me that the compiler would be able to optimize away any difference between these two.

haydentherapper: It’s likely a small difference because the struct is a small size. At scale, handling 1000s of QPS, this impact can become more significant with more allocations.

There’s no official style guide, the best I’ve found is https://github.com/golang/go/wiki/CodeReviewComments#receiver-type. 
The guidance here is prefer a pointer unless there’s a good reason not to, and be consistent. Generally across our code base, I see pointer receivers preferred. 

I agree that there’s a benefit to clearly showing it’s immutable, but I think it’s also pretty clear from the function that the method doesn’t mutate the struct fields, given they’re small functions.

haydentherapper: I don't feel incredibly strongly about this, I'd just prefer consistency

haydentherapper: Thought more, this is fine, it’s consistent within the principal package

haydentherapper (approved):

PR #634, opened 2022-06-02
Add NewClientFromString func.

wlynch:
#### Summary

This is a small wrapper around the existing NewClient, which allows
callers to pass in a string rather than a parsed url.URL.

This is meant to be a replacement for
https://github.com/sigstore/cosign/blob/9ef6b207218572b3257a5b4251418d75569baaae/cmd/cosign/cli/fulcio/fulcio.go#L164-L171
so that callers don't need to implement their own / depend on cosign.

Signed-off-by: Billy Lynch <billy@chainguard.dev>

#### Release Note

```release-note
Added new NewClientFromString function for creating Fulcio clients from raw strings.
```

### DIFF
--- a/pkg/api/client.go
+++ b/pkg/api/client.go
@@ -89,6 +89,15 @@ func NewClient(url *url.URL, opts ...ClientOption) LegacyClient {
 	}
 }
 
+// NewClientFromString creates a new Fulcio API client talking to the provided URL string.
+func NewClientFromString(rawURL string, opts ...ClientOption) (LegacyClient, error) {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return nil, err
+	}
+	return NewClient(u, opts...), nil
+}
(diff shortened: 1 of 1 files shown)


imjasonh (approved):

haydentherapper:
I'd prefer if we can instead use the new V2 API. We're switching over to calling the V2 API over gRPC in https://github.com/sigstore/cosign/pull/1762, I'd prefer we do the same everywhere in the Sigstore ecosystem. If gRPC can't be used for some reason and we prefer to use HTTP, then we can write a V2 client. I'd prefer that we not make any more changes around the V1 client.

wlynch:
SGTM! I'll probably hold off on adding grpc support in gitsign until https://github.com/sigstore/cosign/pull/1762 lands. I can add this as an internal func to gitsign in the meantime.

PR #643, opened 2022-06-10
Add interface for certs/signer fetching to remove mutex

haydentherapper:
For all but the file CA implementation, locking when retrieving the
certs/signer is not necessary. This refactors the intermediate CA struct
by embedding an interface to fetch the certs/signer. Each CA type can
use either the basic structure that simply returns certs/signer, or one
that has a mutex to protect access to the variables.

This also fixes a bug with the mutex. File CA used a different lock than
the intermediate CA struct, so locking while writing wouldn't have
actually protected against concurrent reads. This fixes it by moving the
mutex to the certs/signer struct, and file CA reaches into that struct
to lock while writing.

Also small test fix - go 1.18 checks for duplicate extensions in certs
now, so I had to regenerate a test certificate.

#### Ticket Link

Fixes

#### Release Note

```release-note

```


haydentherapper:
cc @loosebazooka

bobcallaway (approved):
lgtm

PR #645, opened 2022-06-14
Add Tink signing backend

### LINKED ISSUE #646: Add Tink signing backend for local signing
**Description**

A Tink keyset will be encrypted with a remote KMS key, and on CA startup, the keyset will be decrypted and loaded into memory. Keys can be created and encrypted using https://github.com/google/tink/blob/master/docs/TINKEY.md.

There are two main benefits from this work: We can leverage this instead of KMSCA if we need to support a higher QPS, and Tink keysets use strong secure defaults. Keysets can be encrypted with AESGCM, do not rely on a KDF, cannot be brute-forced, and access to the key can be audited through cloud audit logs.

haydentherapper:
#### Summary

This adds support for using encrypted Tink keysets to load a signer.
There are two main benefits from this work: We can leverage this instead
of KMS if we need to support a higher QPS, and Tink keysets use strong
secure defaults. Keysets can be encrypted with AESGCM, do not rely on a
KDF, cannot be brute-forced, and access to the key can be audited
through cloud audit logs.

Tink does not provide a method to extract the signing key from the
keyset intentionally, so I wrote a helper library to reach into the key
handle proto to construct a crypto.Signer.

#### Ticket Link

Fixes #646

#### Release Note

```release-note
Added Tink signing backend for signing certificates with a local encrypted Tink keyset
```

### DIFF
--- a/pkg/ca/tinkca/signer.go
+++ b/pkg/ca/tinkca/signer.go
@@ -1,0 +1,130 @@
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tinkca
+
+import (
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/ed25519"
+	"errors"
+	"fmt"
+	"math/big"
+
+	signatureSubtle "github.com/google/tink/go/signature/subtle"
+	"github.com/google/tink/go/subtle"
+
+	"github.com/golang/protobuf/proto" //lint:ignore SA1019 needed for unmarshalling
+	"github.com/google/tink/go/insecurecleartextkeyset"
+	"github.com/google/tink/go/keyset"
+	commonpb "github.com/google/tink/go/proto/common_go_proto"
+	ecdsapb "github.com/google/tink/go/proto/ecdsa_go_proto"
+	ed25519pb "github.com/google/tink/go/proto/ed25519_go_proto"
+	tinkpb "github.com/google/tink/go/proto/tink_go_proto"
+)
+
+var (
+	ecdsaSignerKeyVersion   = 0
+	ecdsaSignerTypeURL      = "type.googleapis.com/google.crypto.tink.EcdsaPrivateKey"
+	ed25519SignerKeyVersion = 0
+	ed25519SignerTypeURL    = "type.googleapis.com/google.crypto.tink.Ed25519PrivateKey"
+)
+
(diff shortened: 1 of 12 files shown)


haydentherapper:
Yep! It's similar to the file CA implementation, but instead of a password-protected signing key, it's a KMS-encrypted signing key. All signing would happen with an in-memory key. There's only one call to KMS, just to decrypt the key (unlike the current implementation which signs all certs with a KMS key)

Here's the end to end flow:
* One-time, create a Tink keyset to be used for signing, and encrypt it with a KMS key. Also create a CA certificate using the Tink keyset's public key
* Configure the CA with the encrypted Tink keyset, the CA chain, and the KMS key
* On startup, the CA will use the KMS key to decrypt the Tink keyset. Once decrypted, it can be used to sign certificates

haydentherapper:
Yea, if the key leaked from memory somehow, there's nothing that can be done, but at rest, if the encrypted key is compromised, it should be infeasible to decrypt it without the KMS key/service account credentials.

haydentherapper:
Tested with:

```
./tinkey create-keyset --key-template ECDSA_P384 --out enc-keyset.cfg --master-key-uri gcp-kms://projects/project/locations/us-west1/keyRings/test-ring-local/cryptoKeys/test-key

(reverse-i-search)`go run ': go run cmd/fetch_ca_cert/fetch_ca_cert.go   --tink-kms-resource="gcp-kms://projects/project/locations/us-west1/keyRings/test-ring-local/cryptoKeys/test-key" --tink-keyset-path="../tinkey/enc-keyset.cfg" --gcp-ca-parent="projects/project/locations/us-west1/caPools/fulcio-test" --output="chain.crt.pem"
```

edit docker-compose to add necessary flags

test with curl
```

**Comment by haydentherapper** [comment]
@bobcallaway @dlorenc Bumping PR, thanks! Just a note, the only untested function is `tinkca.GetPrimaryKey` since it fetches a key from a remote KMS and the library doesn't support fakes.

File: pkg/ca/tinkca/signer.go [hunk 1]
```diff
@@ -0,0 +1,130 @@
...
+	if k == nil {
+		return nil, errors.New("no enabled key found in keyset")
+	}
+
+	switch k.GetTypeUrl() {
+	case ecdsaSignerTypeURL:
+		// https://github.com/google/tink/blob/9753ffddd4d04aa56e0605ff4a0db46f2fb80529/go/signature/ecdsa_signer_key_manager.go#L48
+		privKey := new(ecdsapb.EcdsaPrivateKey)
+		if err := proto.Unmarshal(k.GetValue(), privKey); err != nil {
+			return nil, fmt.Errorf("error unmarshalling ecdsa private key: %v", err)
```
bobcallaway:
```suggestion
			return nil, fmt.Errorf("error unmarshalling ecdsa private key: %w", err)
```

pkg/ca/tinkca/signer.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+		if err := validateEcdsaPrivKey(privKey); err != nil {
+			return nil, fmt.Errorf("error validating ecdsa private key: %v", err)
```
bobcallaway:
```suggestion
			return nil, fmt.Errorf("error validating ecdsa private key: %w", err)
```

bobcallaway (approved):
lgtm, thanks!

bobcallaway (approved):

PR #670, opened 2022-06-30
Add CORS support to HTTP endpoint

bobcallaway:
This is blocking interactive use of the swagger tool against our production instance.

During this process, I noticed we were still passing some Grpc information in HTTP response headers so this also squelches those.


cmd/app/http.go
```diff
@@ -67,6 +68,9 @@ func createHTTPServer(ctx context.Context, serverEndpoint string, grpcServer, le
 	handler = promhttp.InstrumentHandlerDuration(server.MetricLatency, handler)
 	handler = promhttp.InstrumentHandlerCounter(server.RequestsCount, handler)
 
+	// enable CORS
+	handler = cors.Default().Handler(handler)
```
haydentherapper, line 72: Can you add a comment that the default sets up CORS to accept all domains?

haydentherapper (commented):
Error in test `parse "passthrough:///:\x00\x00\x00\x00\x00\x00\x00": net/url: invalid control character in URL`

cmd/app/http_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	viper.Set("grpc-host", "")
+	viper.Set("grpc-port", 0)
+	grpcServer, err := createGRPCServer(nil, nil, &TrivialCertificateAuthority{})
+	if err != nil {
+		t.Error(err)
+	}
+	grpcServer.startTCPListener()
+	// loop until server starts listening in separate goroutine
+	for {
```
haydentherapper, line 46: nit: Would it be worth adding a timeout (https://gobyexample.com/timeouts) to avoid this test spinning for awhile? Not sure what the default timeout is for a test.

haydentherapper (approved):

haydentherapper (approved):

Issue #689, opened 2022-07-16
Return cert chain in json


lukehinds:
Right now we return the cert chain back with whitespace delimiters, it should instead be a json array. I am having to write some ugly arse regex to parse this out in rust.

haydentherapper:
gRPC solves this - https://github.com/sigstore/fulcio/blob/main/fulcio.proto#L195

lukehinds:
> The v2 API over gRPC solves this - https://github.com/sigstore/fulcio/blob/main/fulcio.proto#L195 For the v2 API over HTTP, this should also occur. We won't be making breaking changes to the v1 API.

cool. so I figure this is not deployed to the fulcio.sigstore.dev yet

haydentherapper:
It has been as of Fulcio 0.4, for example `curl https://fulcio.sigstore.dev/api/v2/trustBundle`

Cosign is using the v2 API currently, we switched over to gRPC.

haydentherapper:
Actually, it looks like there's a bug:

```
$ curl "https://fulcio.sigstore.dev/api/v2/trustBundle" | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1592  100  1592    0     0  10258      0 --:--:-- --:--:-- --:--:-- 10270
{
  "chains": [
    {
      "certificates": [
[...]

PR #690, opened 2022-07-18
ensure GetTrustBundle returns array of strings instead of a single string with newlines

bobcallaway:
Fixes: #689 


pkg/ca/ca.go
```diff
@@ -26,5 +27,5 @@ import (
 // fetching the CA trust bundle.
 type CertificateAuthority interface {
 	CreateCertificate(context.Context, identity.Principal, crypto.PublicKey) (*CodeSigningCertificate, error)
-	Root(ctx context.Context) ([]byte, error)
+	Root(ctx context.Context) ([]*x509.Certificate, error)
```
haydentherapper, line 30: To reflect the public interface, should we return a list of list of certificates?

haydentherapper, line 30: Do you want to rename `Root` to `TrustBundle`, given we're already changing the interface?

haydentherapper (changes requested):
+1 to changing the interface, this looks good! Left a few comments

pkg/ca/googleca/v1/googleca.go
```diff
@@ -136,32 +136,40 @@ func Req(parent string, pemBytes []byte, cert *x509.Certificate) (*privatecapb.C
...
-			pems += strings.Join(c.PemCaCertificates, "")
-		}
-		c.cachedRoots, err = cryptoutils.LoadCertificatesFromPEM(strings.NewReader(pems))
-		if err != nil {
-			panic(fmt.Errorf("failed parsing PemCACertificates: %w", err))
-		}
+func (c *CertAuthorityService) TrustBundle(ctx context.Context) ([][]*x509.Certificate, error) {
+	// if we've already successfully fetched the CA info, just use the cached value
+	if c.cachedRoots != nil {
+		return [][]*x509.Certificate{c.cachedRoots}, nil
```
haydentherapper, line 142: `cachedRoots` should be changed to `[][]*x509.Certificate`

pkg/ca/googleca/v1/googleca.go
```diff
@@ -136,32 +136,40 @@ func Req(parent string, pemBytes []byte, cert *x509.Certificate) (*privatecapb.C
...
-		return c.cachedRoots, fmt.Errorf("error fetching root certificates")
+	for {
+		ca, done := cas.Next()
+		if done == iterator.Done {
+			break
+		} else if done != nil {
+			// if the iterator returns an issue for some reason, exit
+			return [][]*x509.Certificate{}, done
+		}
+		pems += strings.Join(ca.PemCaCertificates, "")
```
haydentherapper, line 158: A CA pool can have multiple CAs. This is concatenating all of the CA certs in a CA pool together, so we instead should have list of chains (each chain represented as a string), then for each string, `cryptoutils.LoadCertificatesFromPEM` to create []*x509.Certificate, then append that to a [][]*x509.Certificate.

pkg/ca/googleca/v1/googleca.go
```diff
@@ -136,32 +136,40 @@ func Req(parent string, pemBytes []byte, cert *x509.Certificate) (*privatecapb.C
...
-	return c.cachedRoots, nil
+	// if we fail to parse the PEM content, return an error
+	roots, err := cryptoutils.LoadCertificatesFromPEM(strings.NewReader(pems))
+	if err != nil {
+		return [][]*x509.Certificate{}, fmt.Errorf("failed parsing PemCACertificates response: %w", err)
+	}
+	if len(roots) == 0 {
+		return [][]*x509.Certificate{}, fmt.Errorf("error fetching root certificates")
+	}
+	c.cachedRootsOnce.Do(func() {
```
haydentherapper, line 168: Does this lock execution for other threads? As in, will other threads block until `Once` is done, so `c.cachedRoots` will always be populated by the end of this? I assume yes, but just checking.

haydentherapper (changes requested):

pkg/ca/googleca/v1/googleca.go
haydentherapper: Looks like yep, there's a comment with this case - https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/sync/once.go;l=49-55

haydentherapper (approved):

PR #703, opened 2022-07-22
Create certificate specification

### LINKED ISSUE #698: Create a specification for issued certificates
**Description**

Requirements:
* One SAN
* IDP OID is populated

Optional
* GitHub OIDs are populated

(WIP list)

haydentherapper:
This outlines what MUST and SHOULD and MAY and SHOULD NOT
and MUST NOT be included with CA and end-entity certificates.

Fixes #698

### DIFF
--- a/docs/certificate-specification.md
+++ b/docs/certificate-specification.md
@@ -1,6 +1,11 @@
 # Certificate Specification
 
 This document includes the requirements for root, intermediate, and issued certificates.
+This document applies to all instances of Fulcio, including the production instance and
+all private instances using the service defined in this repository.
+
+The key words "MUST", "MUST NOT", "SHOULD", "SHOULD NOT", and "MAY" in this document are
+to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
 
 ## Root Certificate
 
@@ -29,7 +34,7 @@ A root certificate MAY:
 * Specify a Basic Constraints path length constraint to prevent additional CA certificates
   from being issued beneath the root
 * Specify an Authority Key Identifier. If specified, it MUST be the same as the Subject Key Identifier
-* Specify other values in the Subject.
+* Specify other values in the Subject
 
 ## Intermediate Certificate
 
@@ -62,11 +67,16 @@ An intermediate certificate SHOULD NOT:
 
 * Use a different signature scheme (ECDSA vs RSA) than its parent certificate, as some clients do not support this
 
+An intermediate certificate MAY:
+
+* Be optional. An end-entity certificate MAY be issued from a root certificate or an intermediate certificate.
+  Clients MUST be able to verify a chain with any number of intermediate certificates.
+
 ## Issued Certificate
 
 An issued certificate MUST:
 
-* Specify a Subject Alternative Name as a critical extension. It MUST be populated by either:
+* Specify exactly one Subject Alternative Name, as a critical extension. It MUST be populated by either:
    * An email
    * A URI
 * Specify an Issuer equal to the parent certificate's Subject
@@ -83,13 +93,12 @@ An issued certificate MUST:
    * RSA of key size 2048 to 4096 (inclusive) with size % 8 = 0, E = 65537, and containing no weak primes
    * ED25519
 * Specify the OpenID Connect identity token issuer with OID `1.3.6.1.4.1.57264.1.1`
-* Append a precertificate to a Certificate Transparency log, where the precertificate MUST be signed by the certificate authority
-  and MUST include a poison extension with OID `1.3.6.1.4.1.11129.2.4.3`
-* Specify the Signed Certificate Timestamp (SCT) from the Certificate Transparency log with OID `1.3.6.1.4.1.11129.2.4.2`
+* Be appended to a Certificate Transparency log. Clients MUST NOT trust certificates that do not present
+  either a proof of inclusion or a Signed Certificate Timestamp (SCT)
 
 An issued certificate MUST NOT:
 
-* Specify a Subject
+* Specify a nonempty Subject
 * Specify multiple Subject Alternative Name values
 * Specify other Key Usages besides Digital Signature
 * Specify other Extended Key Usages besides Code Signing
@@ -98,6 +107,9 @@ An issued certificate SHOULD:
 
 * Use an ephemeral key. A client MAY request a certificate with a long-lived key, but a client MUST
   adequately secure the key material
+* Append a precertificate to a Certificate Transparency log, where the precertificate MUST be signed by the certificate authority
+  and MUST include a poison extension with OID `1.3.6.1.4.1.11129.2.4.3`
+* Specify the Signed Certificate Timestamp (SCT) from the Certificate Transparency log with OID `1.3.6.1.4.1.11129.2.4.2`
 
 An issued certificate SHOULD NOT:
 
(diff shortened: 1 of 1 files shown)


haydentherapper:
cc @woodruffw

docs/certificate-specification.md
```diff
@@ -0,0 +1,112 @@
+# Certificate Specification
+
+This document includes the requirements for root, intermediate, and issued certificates.
```
bobcallaway, line 3: Are you intending these restrictions for all instances of Fulcio, or just the instance operated by the sigstore community?

haydentherapper: I was thinking for the public instance, but this could be tweaked slightly for all instances. The main differences would be an optional intermediate, and an SCT not always being embedded. I’d also want to exclude ephemeral, since that’s for testing.

is there a preference?

woodruffw: I have a strong preference towards the issued x509 certificate restrictions being required for all instances -- the soundness of our verification semantics for depends on there being exactly one SAN, for example, and other instances should conform to that constraint.

woodruffw: (But I agree re: intermediates and detached SCTs.)

docs/certificate-specification.md
```diff
@@ -0,0 +1,112 @@
...
+
+An intermediate certificate SHOULD NOT:
+
+* Use a different signature scheme (ECDSA vs RSA) than its parent certificate, as some clients do not support this
+
+## Issued Certificate
+
+An issued certificate MUST:
+
+* Specify a Subject Alternative Name as a critical extension. It MUST be populated by either:
```
woodruffw, line 69: Just to make this as explicit as possible (it's also listed under `MUST NOT`, but I figured a little bit of redundancy doesn't hurt):

```suggestion
* Specify exactly one Subject Alternative Name, as a critical extension. It MUST be populated by either:
```

docs/certificate-specification.md
```diff
@@ -0,0 +1,112 @@
...
+   * RSA of key size 2048 to 4096 (inclusive) with size % 8 = 0, E = 65537, and containing no weak primes
+   * ED25519
+* Specify the OpenID Connect identity token issuer with OID `1.3.6.1.4.1.57264.1.1`
+* Append a precertificate to a Certificate Transparency log, where the precertificate MUST be signed by the certificate authority
+  and MUST include a poison extension with OID `1.3.6.1.4.1.11129.2.4.3`
+* Specify the Signed Certificate Timestamp (SCT) from the Certificate Transparency log with OID `1.3.6.1.4.1.11129.2.4.2`
+
+An issued certificate MUST NOT:
+
+* Specify a Subject
```
woodruffw, line 92: For consistency with the above ("Specify an empty Subject"):

```suggestion
* Specify a nonempty Subject
```

docs/certificate-specification.md
```diff
@@ -0,0 +1,112 @@
+# Certificate Specification
+
```
woodruffw, line 2: Tiny nitpick: might want to reference https://www.ietf.org/rfc/rfc2119.txt for the semantics of `MUST`, `SHOULD`, etc.

docs/certificate-specification.md
haydentherapper: Agreed - I’ll make these updates.

haydentherapper: Updated!

haydentherapper: Updated!

docs/certificate-specification.md
```diff
@@ -0,0 +1,125 @@
...
+## Root Certificate
+
+A root certificate MUST:
+
+* Specify a Subject with a common name and organization
+* Specify an Issuer with the same values as the Subject
+* Specify Key Usages for Certificate Sign and CRL Sign
+* Specify Basic Constraints to `CA:TRUE`
+* Specify a unique, positive, 160 bit serial number
```
woodruffw, line 18: Maybe too nitpicky, but can we offer some normative `MAY` guidance for generating this serial?

docs/certificate-specification.md
```diff
@@ -0,0 +1,125 @@
...
+* Specify a Subject Key Identifier
+* Specify an Authority Key Identifier equal to the parent certificate's Subject Key Identifier 
+* Specify an empty Subject
+* Be compliant with [RFC5280](https://datatracker.ietf.org/doc/html/rfc5280)
+* Specify a public key that is either:
+   * ECDSA NIST P-256, NIST P-384, or NIST P-521
+   * RSA of key size 2048 to 4096 (inclusive) with size % 8 = 0, E = 65537, and containing no weak primes
+   * ED25519
+* Specify the OpenID Connect identity token issuer with OID `1.3.6.1.4.1.57264.1.1`
```
woodruffw, line 95: Another tiny nitpick, but it might be worth linking these OIDs to their sections in https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md

docs/certificate-specification.md
```diff
@@ -0,0 +1,125 @@
...
+* Use an ephemeral key. A client MAY request a certificate with a long-lived key, but a client MUST
+  adequately secure the key material
+* Append a precertificate to a Certificate Transparency log, where the precertificate MUST be signed by the certificate authority
+  and MUST include a poison extension with OID `1.3.6.1.4.1.11129.2.4.3`
+* Specify the Signed Certificate Timestamp (SCT) from the Certificate Transparency log with OID `1.3.6.1.4.1.11129.2.4.2`
+
+An issued certificate SHOULD NOT:
+
+* Use a different public key scheme (ECDSA vs RSA) than its parent certificate, as some clients do not support this
+* Specify a public key that is stronger than its parent certificate
```
woodruffw, line 117: I might be missing it, but this is a little unintuitive to me: why should the ephemeral signing certificate be limited in key strength to no more than its issuer?

docs/certificate-specification.md
haydentherapper: Do you mean how to generate it, as in what's the best way to generate a random 160 bit number, or do you mean to relax this requirement? For the latter, this is coming from RFC5280, I just called it out specifically since it's a less well-known part of 5280.

woodruffw: The former 🙂 -- I wanted to understand if we're asking for the number to be *random* (~approximating no numeric relationship between certificates) or merely *unique* (never repeated, but perhaps a simple increasing integer).

If it's meant to be like 5280 (https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.2), then maybe qualify with "according to the same rules as RFC 5280 4.1.2.2"?

znewman01 (dismissed):
I echo @woodruffw's comments  but otherwise LGTM

haydentherapper:
Thanks y’all! Sorry, will circle back to finishing this later this week.

docs/certificate-specification.md
haydentherapper: Random is recommended, IIRC there have been some issues in the past with monotonically increasing values. I'll note this, and add the qualification.

docs/certificate-specification.md
haydentherapper: Added

docs/certificate-specification.md
haydentherapper: The thought is that as keys fail due to new weaknesses, the issued certificates should be compromised before the parent certificate. If a parent's key is weaker and it is compromised, then it can simply issue new children. If a child's key is weaker, then there's no impact on the parent if it's compromised.

woodruffw: Makes sense!

woodruffw (approved):

dlorenc (approved):

PR #732, opened 2022-08-10
fix example to explicitly set port for gRPC call

bobcallaway:

### DIFF
--- a/examples/request-certificate/main.go
+++ b/examples/request-certificate/main.go
@@ -43,7 +43,7 @@ var (
 )
 
 // Some of this is just ripped from cosign
-func GetCert(signer *signature.RSAPKCS1v15SignerVerifier, fc fulciopb.CAClient, oidcIssuer string, oidcClientID string) (*fulciopb.SigningCertificate, error) {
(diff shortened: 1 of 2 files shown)


examples/request-certificate/main.go
```diff
@@ -85,10 +85,13 @@ func NewClient(fulcioURL string) (fulciopb.CAClient, error) {
...
+	hostWithPort := fmt.Sprintf("%s:80", fulcioServer.Host)
 	if fulcioServer.Scheme == "https" {
 		dialOpt = grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{}))
+		hostWithPort = fmt.Sprintf("%s:443", fulcioServer.Host)
 	}
-	conn, err := grpc.Dial(fulcioServer.Host, dialOpt)
+
+	conn, err := grpc.Dial(hostWithPort, dialOpt)
 	if err != nil {
 		return nil, err
```
haydentherapper, line 96: While you're updating this, could you also change the default signer to ECDSA? If people copy this, I'd prefer that over RSA.

PR #735, opened 2022-08-11
address Potential Slowloris Attack because ReadHeaderTimeout is not configured in the http.Server

cpanato:
#### Summary
- address Potential Slowloris Attack because ReadHeaderTimeout is not configured in the http.Server
- update golangci-lint to 1.48.0

#### Release Note
- address Potential Slowloris Attack because ReadHeaderTimeout is not configured in the http.Server

### DIFF
--- a/cmd/app/serve.go
+++ b/cmd/app/serve.go
@@ -80,6 +80,7 @@ func newServeCmd() *cobra.Command {
 	cmd.Flags().String("grpc-host", "0.0.0.0", "The host on which to serve requests for GRPC")
 	cmd.Flags().String("grpc-port", "8081", "The port on which to serve requests for GRPC")
 	cmd.Flags().String("metrics-port", "2112", "The port on which to serve prometheus metrics endpoint")
+	cmd.Flags().Duration("read-header-timeout", 10*time.Second, "The time allowed to read the headers of the requests in seconds")
 
 	// convert "http-host" flag to "host" and "http-port" flag to be "port"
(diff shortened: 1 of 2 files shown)


haydentherapper:
@cpanato Is there an equivalent setting for the gRPC handler?

cpanato:
> @cpanato Is there an equivalent setting for the gRPC handler?

i'm not aware, but we can check

cmd/app/serve.go
```diff
@@ -80,6 +80,7 @@ func newServeCmd() *cobra.Command {
 	cmd.Flags().String("grpc-host", "0.0.0.0", "The host on which to serve requests for GRPC")
 	cmd.Flags().String("grpc-port", "8081", "The port on which to serve requests for GRPC")
 	cmd.Flags().String("metrics-port", "2112", "The port on which to serve prometheus metrics endpoint")
+	cmd.Flags().Duration("read-header-timeout", 60*time.Second, "The time allowed to read the headers of the requests in seconds")
```
haydentherapper, line 83: Is this too long, should it be ~10 seconds or less (not much in the headers besides an auth token)? I assume we have request timeouts that would trigger before this, though I'm not sure where those are configured (k8s?).

cpanato: nginx uses 60s by default
i will reduce that to 10

cpanato: done

we need to check what is the value we are setting on the nginx side (`client_header_timeout`)

haydentherapper (approved):
thanks!

Issue #716, opened 2022-08-01
Enforce that identities are of an expected format


haydentherapper:
**Description**

Dex - Look like email, K8S - Look like a service account, Username - Doesn't look like an email, etc

Context: https://docs.google.com/document/d/1o8_bXIygufgiohJGlmBzqF4_BnXCTfgh4ILgJFJxYRs/edit?resourcekey=0-YEar3v67uoT31kj83dCVvA#heading=h.oiw6nn1ucgaq

znewman01:
I'll attempt to summarize the (very long) Google Doc comment thread linked in "Context" above.

Fulcio issues certs with the identity of the issuer in the Subject Alternative Name field. In X.509, the value in this field is a *union type* (see the definition of "GeneralName" in [RFC 5280, 4.2.1.6](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6)) indicating whether the value is, for instance, an email (`rfc822Name`), a URI (`uniformResourceIdentifier`), or an IP (`iPAddress`).

If we have clients provide a simple "identity" field, along with an "issuer," do they need to indicate the type (for instance, writing `uri:spiffe://domain/...` or `email:foo@bar.com`)? This is pretty cumbersome.

If users *don't* indicate the type, we have to worry about whether this might confuse a user into accepting a value that they otherwise wouldn't. There are two ways this could happen:

1. "Disjunction confusion:" if Fulcio issues certificates from the same issuer with different SAN types but the same string value (e.g., `uri:foo` and `email:foo`), then the client would accept *either* if they specify `foo` as the identity.

2. "Type confusion:" can an attacker generate a cert with a SAN of one type and get the victim to verify it with another type implicitly in mind?

*We don't have to worry about "disjunction confusion!"* Fulcio already [restricts each issuer](https://github.com/sigstore/fulcio/blob/ab2538a129ff1c069fdb414a7bde06ee47a7f2f8/config/config.jsn#L6) to only one ["issuer type"](https://github.com/sigstore/fulcio/blob/main/docs/oidc.md), which in turn will only ever produce one SAN type.

*That leaves "type confusion."* Such attacks have two preconditions: the attacker can fool the CA into emitting a SAN whose value is "wrong" for the type (e.g. is an email instead of a URI), and that value must be controllable to a value the victim thinks they're verifying.
    
[...]

PR #802, opened 2022-09-26
Change username format, enforce identity format 

haydentherapper:
#### Summary

This updates the username type to avoid the username subject format
looking like an email. Fulcio will now specify the subject in the
OtherName SAN, and the format will use a ! instead of @.

This required some custom ASN.1 marshalling and unmarshalling, since
crypto/x509 does not support the OtherName SAN.

This also adds enforcement that email subjects match a basic email
regex format, and that other types do not look like emails.

Fixes https://github.com/sigstore/fulcio/issues/716

#### Release Note

Changed username identity format to username!Domain, username now specified in the OtherName SAN


haydentherapper:
cc @woodruffw - Thanks for the suggestion to use OtherName! You also gave me an excuse to dig a bit more into ASN.1 encoding, which may or may not have been productive. :)

pkg/identity/email/principal.go
```diff
@@ -40,6 +42,11 @@ func PrincipalFromIDToken(ctx context.Context, token *oidc.IDToken) (identity.Pr
 		return nil, errors.New("email_verified claim was false")
 	}
 
+	emailRegex := regexp.MustCompile(`^.+@.+\..+$`)
```
bobcallaway, line 45: Can we use something like https://github.com/asaskevich/govalidator/blob/f21760c49a8d602d863493de796926d2a5c1138d/validator.go#L67 instead? 

the regex they use is a bit more complicated (and potentially more thorough?) - https://github.com/asaskevich/govalidator/blob/f21760c49a8d602d863493de796926d2a5c1138d/patterns.go#L7

bobcallaway: if not, then can we move this to a const or done in `init()` so we aren't compiling the regex on every call?

pkg/identity/uri/principal.go
```diff
@@ -40,6 +41,11 @@ func PrincipalFromIDToken(ctx context.Context, token *oidc.IDToken) (identity.Pr
 		return nil, errors.New("invalid configuration for OIDC ID Token issuer")
 	}
 
+	emailRegex := regexp.MustCompile(`^.+@.+\..+$`)
+	if emailRegex.MatchString(uriWithSubject) {
+		return nil, fmt.Errorf("uri subject should not be an email address")
+	}
+
```
bobcallaway, line 48: can you reuse whatever the solution is in `pkg/identity/email` here rather than re-writing it?

pkg/identity/username/principal.go
```diff
@@ -28,29 +30,34 @@ import (
...
 
-	if strings.Contains(username, "@") {
-		return nil, errors.New("username cannot contain @ character")
+	if strings.Contains(username, "!") {
+		return nil, errors.New("username cannot contain ! character")
+	}
+
+	emailRegex := regexp.MustCompile(`^.+@.+\..+$`)
+	if emailRegex.MatchString(username) {
+		return nil, fmt.Errorf("uri subject should not be an email address")
```
bobcallaway, line 47: same as before, can we re-use?

pkg/identity/email/principal.go
haydentherapper: done

woodruffw:
Exciting! I'll update the relevant bits on `sigstore-python` once this lands 🙂

bobcallaway (approved):

haydentherapper:
Found issue, crypto/x509 doesn't handle a critical SAN extension with a GeneralName like OtherName (not DNS/IP/email/URI). Either needs to be non-critical or figure out if we can fix this in the library

haydentherapper:
AI: Look at certificate-transparency-go

haydentherapper:
Found a way to handle this! You can remove unhandled critical extensions before verifying, there's a `cert.UnhandledCriticalExtensions` array of OIDs populated by `x509.ParseCertificate`. Just remove the SAN OID before verifying and we're good to go. This should be ready to merge now, and I'll get the Cosign update in.

I've filed a proposal in Go to add support for OtherName too.

znewman01:
@haydentherapper just one comment: a "username" *would* cause an existing Cosign client to choke, but you found an easy fix and we can roll that out to clients soon. In the meantime, we don't actually have username issuers configured in the public instance so it's not a problem in practice, and we just need to wait. Is that right?

The other fixes (e.g., verifying email format) won't break anybody because we've haven't seen any SANs that don't match so we don't need to worry there.

haydentherapper:
Yes, correct. No impact to production because we don't have any username issuers currently. This could impact someone who's deployed their own instance of Fulcio, but I am doubtful anyone is using the username identity right now, since it's not thoroughly documented. Also it would only impact them once they've updated to the latest Fulcio, so I'll make a note in the release notes to update to the latest Cosign, and we'll not cut a new release until the Cosign fix is out.

hslatman:
@haydentherapper do you have an example certificate that includes the `OtherName` you can share? I'm adding support for printing the Sigstore OIDs to [smallstep/certinfo](https://github.com/smallstep/certinfo) in this PR: https://github.com/smallstep/certinfo/pull/18 and would like to test the output of this new OID too.

haydentherapper:
Sweet, awesome to see this being added! This cert was generated from the tests:

```
-----BEGIN CERTIFICATE-----
MIICtDCCAlqgAwIBAgIUT3Bh3mWmJp+2wTQUxj19R1vYVKcwCgYIKoZIzj0EAwIw
aDEMMAoGA1UEBhMDVVNBMQswCQYDVQQIEwJXQTERMA8GA1UEBxMIS2lya2xhbmQx
FTATBgNVBAkTDDc2NyA2dGggU3QgUzEOMAwGA1UEERMFOTgwMzMxETAPBgNVBAoT
CHNpZ3N0b3JlMB4XDTIyMDkyOTE4MzcyMVoXDTIyMDkyOTE4NDcyMVowADBZMBMG
ByqGSM49AgEGCCqGSM49AwEHA0IABJUlrKgma8vvHGsin0Lns8ilkZIGfbFoNhp/
e2HPdLPwmy1IfigG2FRmMjw7DymnC3ce9o1k0zkcCuhc0Pn6PnOjggFIMIIBRDAO
BgNVHQ8BAf8EBAMCB4AwEwYDVR0lBAwwCgYIKwYBBQUHAwMwHQYDVR0OBBYEFDst
jNbwcRDr33q5Tt0kUjPLrG0sMB8GA1UdIwQYMBaAFN8zJRpf/idR3ooX82oyJ5BC
UFGjMCsGA1UdEQEB/wQhMB+gHQYKKwYBBAGDvzABB6APDA1mb28hMTI3LjAuMC4x
MCQGCisGAQQBg78wAQEEFmh0dHA6Ly8xMjcuMC4wLjE6NDU2NjEwgYkGCisGAQQB
1nkCBAIEewR5AHcAdQAodhoYkCf77zzQ1hoBjXawUFcpx6dBG8y99gT0XUJhUwAA
AAAAAAU5AAAEAwBGMEQCIAhzbUnllt0pkfDnAvEI/4yEESxXn6d+8bJqQ4i/oU68
AiBZgxlvo/D//+6I0ztBODY5/U6qYrUZt4a4SX1Uk3y/xzAKBggqhkjOPQQDAgNI
ADBFAiEAogo/upgvfrjroS72BcIrPznne7NC6mOt0jDVFDfFgswCIArUlhr07ftG
KCl3yVaw8c5Ck5p5eRSNY/RqZ0YRC7Ir
-----END CERTIFICATE-----
```

hslatman:
It works, @haydentherapper!

```console
   ...
                keyid:DF:33:25:1A:5F:FE:27:51:DE:8A:17:F3:6A:32:27:90:42:50:51:A3
                Fulcio Identity: foo!127.0.0.1
            Fulcio OIDC Issuer:
                http://127.0.0.1:45661
            RFC6962 Certificate Transparency SCT:
   ...
```

What do you think would be the best name for these? I've used `Fulcio` now, because that's the direct parent OID for all, but `Sigstore` is probably better known and may need to be given some visibility in the output?

haydentherapper:
I think `Sigstore`. Fulcio is more or less an implementation detail of Sigstore, and we could imagine another identity service that supports issuing certificates with the same format.

Issue #840, opened 2022-10-18
intermittent test failure in GRPC


bobcallaway:
**Description**

As seen in https://github.com/sigstore/fulcio/actions/runs/3276150029/jobs/5391859692, we occasionally see test failures due to a URL parsing issue:

```
=== RUN   TestHTTPCORSSupport
2022-[10](https://github.com/sigstore/fulcio/actions/runs/3276150029/jobs/5391859692#step:7:11)-18T19:02:20.812Z	INFO	app/grpc.go:105	listening on grpc at :::33327
2022-10-18T19:02:20.812Z	FATAL	app/http.go:56	parse "passthrough:///:0\x00\x00\x00\x00\x00\x00": net/url: invalid control character in URL
github.com/sigstore/fulcio/cmd/app.createHTTPServer
	/home/runner/work/fulcio/fulcio/cmd/app/http.go:56
github.com/sigstore/fulcio/cmd/app.setupHTTPServer
	/home/runner/work/fulcio/fulcio/cmd/app/http_test.go:61
github.com/sigstore/fulcio/cmd/app.TestHTTPCORSSupport
	/home/runner/work/fulcio/fulcio/cmd/app/http_test.go:71
testing.tRunner
	/opt/hostedtoolcache/go/1.19.2/x64/src/testing/testing.go:[14](https://github.com/sigstore/fulcio/actions/runs/3276150029/jobs/5391859692#step:7:15)46
FAIL	github.com/sigstore/fulcio/cmd/app	0.025s
```

**Version**

HEAD

bobcallaway:
was able to recreate locally:

```
=== RUN   TestHTTPDoesntLeakGRPCHeaders
2022/10/21 08:29:17 INFO: [core] [Server #7] Server created
2022-10-21T08:29:17.623-0400    INFO    app/grpc.go:105 listening on grpc at :::38983
2022/10/21 08:29:17 INFO: [core] [Channel #8] Channel created
2022/10/21 08:29:17 INFO: [core] [Channel #8] original dial target is: ":0\x00\x00\x00\x00\x00\x00"
2022/10/21 08:29:17 INFO: [core] [Server #7 ListenSocket #9] ListenSocket created
2022/10/21 08:29:17 INFO: [core] [Channel #8] dial target ":0\x00\x00\x00\x00\x00\x00" parse failed: parse ":0\x00\x00\x00\x00\x00\x00": net/url: invalid control character in URL
[...]

PR #849, opened 2022-10-21
fix race condition in unit tests

bobcallaway:
`grpcServerEndpoint` could be read at any point depending on goroutine scheduling. Sometimes that read occurred (appearing intermittently in CI testing) after memory for a string was allocated but before the string value was populated with non-null bytes, leading to issues as reported in #840. 

This reduces the scope of what is run within the go routine when launching the grpc server.

Fixes: #840


haydentherapper (dismissed):
Good find!!

haydentherapper (approved):

PR #859, opened 2022-10-26
client: add retryable client with default

asraa:
Signed-off-by: Asra Ali <asraa@google.com>

This adds a retryable http client that has a default retry for use in other libraries. See https://github.com/sigstore/cosign/issues/2198

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,7 @@ require (
 	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.12.0
+	github.com/hashicorp/go-retryablehttp v0.7.1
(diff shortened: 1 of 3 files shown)


haydentherapper (changes requested):
This is for the legacy client, which I don't want to support.

The hope was most would use auto-generated gRPC clients going forward. I'd also be fine with creating a new client for HTTP connection, and that one can use retriable calls.

haydentherapper:
Correct, there's an open PR (https://github.com/sigstore/cosign/pull/1762) to switch off the legacy client to gRPC.

> Do you mean post-grpc? Or was there something else that marked this as legacy?

The current "v2" API supports either gRPC or HTTP. We don't need to maintain a gRPC client. For HTTP, we could maintain one. Legacy was the "v1" API that doesn't support gRPC

asraa:
> For HTTP, we could maintain one. Legacy was the "v1" API that doesn't support gRPC

OOOH got it! Thanks: we can add it in that PR, no need to support an HTTP one

Issue #754, opened 2022-08-23
Standardizing CI OIDC token claims


haydentherapper:
## Goal

Create a standard set of claims that should be present in OIDC tokens from CI systems such as GitHub Actions, Cirrus CI, GitLab, Circle CI, etc.

## Background

As noted in the NPM RFC for integrating with Sigstore, and as documented in other tickets (https://github.com/sigstore/fulcio/issues/243, https://github.com/sigstore/fulcio/issues/591, https://github.com/sigstore/fulcio/issues/748), there is interest in support for other CI systems. It is technically possible to implement support for each, but it will require code duplication and work for onboarding every CI platform. It would be ideal if all OIDC tokens from all CI systems had a standard set of claims to represent identity, so that onboarding would simply be updating configuration.

## Current state

All of the above platforms either are working on or currently produce OIDC tokens for CI workflows. Fulcio currently only accepts CI tokens from GitHub Actions, and has hardcoded the GitHub specific claim values and produces a code signing certificate with GitHub specific OID values.

### Currently expected claims ([GitHub ref](https://docs.github.com/en/enterprise-cloud@latest/actions/deployment/security-hardening-your-deployments/using-openid-connect-with-reusable-workflows))
* `job_workflow_ref`
* `sha`
* `event_name`
* `repository`
* `workflow`
* `ref`
* `aud` (which must be set to `sigstore`)
* `exp`

`sha`, `event_name`, `repository`, `workflow`, and `ref` are included in issued certificates in custom OIDs - https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md.

## Required claims

The token should include standard OIDC claims like:
* `aud` (which must be customizable and set to `sigstore`)
* `sub`
* `iss`
* `exp`
* `iat`
* `nbf`

We should include the claims specified in "Currently expected claims".

There was conversation in https://github.com/sigstore/fulcio/issues/624 about including the run ID (`run_id`), run count (`run_number`) and attempt count (`run_attempt`). We should decide if these should be required for Fulcio certificates.

Another useful claim may be `actor`, who triggered the CI run.

Any claim values must be immutable. For example, user IDs should be used instead of usernames, and repository IDs should be used instead of repository names, to prevent resurrection attacks.

cc @asraa @laurentsimon @znewman01 @fkorotkov @feelepxyz, what would you like to see in a token and do you have recommendations on claim names?

wlynch:
[...]

Issue #397, opened 2022-02-08
Designing list of requirements for new IDPs for public good deployment


haydentherapper:
**Description**

In today's community meeting, we discussed two ways of supporting new IDPs:
* Integrating with Dex
* Adding the IDP directly to Fulcio's OIDC issuer config list

We need to determine what are the requirements for a new IDP before inclusion.

Some initial thoughts:
* As pointed out in the meeting, Dex allows us to restrict the lifetime of issued tokens. Lifetime isn't included in the hosted openid-configuration, so this may have to just be an audit and not enforced (beyond rejecting tokens that are over a certain lifetime, but at that point, the token's already been sent over the wire so it's not worth rejecting the requesting in my opinion). 
* Also brought up in the meeting, we should require that IDPs prevent identity subject reuse.
* Key rotation policy
* Signing key storage policy
* Uptime requirements
* A hosted `.well-known/openid-configuration` (which should be a given for OIDC, but it's worth having this explicitly checked as part of onboarding a new IDP)
* Minimum set of supported claims (issuer, subject, audience, issued at, expiration). I propose that we don't enforce that all tokens have precise claim name (`iss`, `sub`, `aud`, for some examples), but that these values are represented in some claim on the token. For example, the subject could be in `sub` or `email` or maybe some other claim.
* An OIDC provider must challenge the email address
* A configurable audience (`aud`) for the token, setting the audience to `sigstore`

What else would we like to see in in IDPs?

trevrosen:
On a related note, I'm curious how the Sigstore team is planning to expand the list of the [cert's SAN OIDs](https://github.com/sigstore/fulcio/blob/c74e2cfb763dd32def5dc921ff49f579fa262d96/docs/oid-info.md). 

I'm interested in use cases involving multiple integrated build systems which also function as an IdP (like GitHub Actions does). Will the OID list expand to cover other systems directly or are there plans to abstract build-related claims?

evankanderson:
Out of curiousity:

> An OIDC provider must challenge the email address

What's the requirement for this?

Also, I'm wondering about:

> Minimum set of supported claims (issuer, subject, audience, issued at, expiration). I propose that we don't enforce that all tokens have precise claim name (iss, sub, aud, for some examples), but that these values are represented in some claim on the token. For example, the subject could be in sub or email or maybe some other claim.

[...]

PR #890, opened 2022-11-16
Add Buildkite OIDC to Fulcio

### LINKED ISSUE #889: Add Buildkite to federation
**Description**

Hi! I'm Sam, I'm an engineer at [Buildkite](https://buildkite.com). We've been working on providing OIDC tokens within Buildkite pipelines, for CI/CD workloads. I'd like to be able to use Buildkite OIDC tokens to sign containers produced by my builds.

We have a command which can be run within a build to provision an OIDC token with a custom audience, asserting the current job and agent identity:

https://github.com/buildkite/agent/pull/1827

Combining this with cosign should theoretically allow [OIDC signing](https://docs.sigstore.dev/cosign/openid_signing). But a pipeline like this:

```yaml
steps:
- command: |
    docker build -t sj26/test:$BUILDKITE_BUILD_NUMBER .
    COSIGN_EXPERIMENTAL=1 cosign sign --identity-token="$(buildkite-agent oidc request-token --audience sigstore)" sj26/test:$BUILDKITE_BUILD_NUMBER
```

currently produces this error:

```
COSIGN_EXPERIMENTAL=1 cosign sign --identity-token="$(...)" sj26/test:16
Generating ephemeral keys...
Retrieving signed certificate...
 
        Note that there may be personally identifiable information associated with this signed artifact.
        This may include the email address associated with the account with which you authenticate.
        This information will be used for signing this artifact and will be stored in public transparency logs and cannot be removed later.
Error: signing [sj26/test:161]: getting signer: getting key from Fulcio: retrieving cert: POST https://fulcio.sigstore.dev/api/v1/signingCert returned 401 Unauthorized: "{\"code\":16,\"message\":\"There was an error processing the credentials for this request\",\"details\":[]}"
main.go:62: error during command execution: signing [sj26/test:161]: getting signer: getting key from Fulcio: retrieving cert: POST https://fulcio.sigstore.dev/api/v1/signingCert returned 401 Unauthorized: "{\"code\":16,\"message\":\"There was an error processing the credentials for this request\",\"details\":[]}"
```

I believe that means we need to be added to the federation config here:

https://github.com/sigstore/fulcio/tree/main/federation

I'm currently opening a PR, but required an issue open first. So, here it is! 🙏

sj26:
Add Buildkite to the OIDC connect federation configuration. Fixes #889.

#### Summary
I'd like to be able to use Buildkite OIDC tokens to sign containers produced by my builds. There are more details in #889.

Our OIDC token support is still under development, but it's possible to try it out now by signing up for Buildkite and running a pipeline like:

```
steps:
  - command: |
      docker build -t sj26/test:$BUILDKITE_BUILD_NUMBER .
      COSIGN_EXPERIMENTAL=1 \
        cosign sign \
          --identity-token="$$(buildkite-agent oidc request-token --audience sigstore)" \
          "sj26/test:$${BUILDKITE_BUILD_NUMBER}"
```

I wasn't sure about the "type", so I followed the example of "github-workflow" — our tokens represent one "buildkite-job" which is a process being run on a host somewhere as part of a build pipeline.

#### Release Note
- Add Buildkite OIDC to Fulcio

#### Documentation
I couldn't find specific docs for each OIDC provider, so I don't think any updates are required.

### DIFF
--- a/federation/agent.buildkite.com/config.yaml
+++ b/federation/agent.buildkite.com/config.yaml
@@ -1,4 +1,4 @@
-# Copyright 2022 The Sigstore Authors
+# Copyright 2023 The Sigstore Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/federation/agent.buildkite.com/config.yaml
+++ b/federation/agent.buildkite.com/config.yaml
@@ -1,0 +1,18 @@
+# Copyright 2022 The Sigstore Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+url: https://agent.buildkite.com
+contact: support@buildkite.com
+description: "Buildkite Agent OIDC tokens for job identity"
+type: "buildkite-job"
--- a/pkg/identity/buildkite/principal.go
+++ b/pkg/identity/buildkite/principal.go
@@ -1,0 +1,98 @@
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package buildkite
+
+import (
+	"context"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"net/url"
+
+	"github.com/coreos/go-oidc/v3/oidc"
+	"github.com/sigstore/fulcio/pkg/certificate"
+	"github.com/sigstore/fulcio/pkg/identity"
+)
+
+type jobPrincipal struct {
+	// Subject matches the 'sub' claim from the OIDC ID token this is what is
+	// signed as proof of possession for Buildkite job identities
+	subject string
+
+	// OIDC Issuer URL. Matches 'iss' claim from ID token. The real issuer URL is
+	// https://agent.buildkite.com/.well-known/openid-configution
+	issuer string
+
+	// The full URL to the job. This will be the set as SubjectAlternativeName URI in
+	// the final certificate.
+	url string
+}
+
+func JobPrincipalFromIDToken(ctx context.Context, token *oidc.IDToken) (identity.Principal, error) {
+	var claims struct {
+		OrganizationSlug string `json:"organization_slug"`
+		PipelineSlug     string `json:"pipeline_slug"`
+		BuildNumber      int    `json:"build_number"`
+		JobID            string `json:"job_id"`
+	}
+	if err := token.Claims(&claims); err != nil {
+		return nil, err
+	}
+
+	if claims.OrganizationSlug == "" {
+		return nil, errors.New("missing organization_slug claim in ID token")
+	}
+
+	if claims.PipelineSlug == "" {
+		return nil, errors.New("missing pipeline_slug claim in ID token")
+	}
+
+	if claims.BuildNumber == 0 {
+		return nil, errors.New("missing build_number claim in ID token")
+	}
+
+	if claims.JobID == "" {
+		return nil, errors.New("missing job_id claim in ID token")
+	}
+
+	return &jobPrincipal{
+		subject: token.Subject,
+		issuer:  token.Issuer,
+		url:     fmt.Sprintf("https://buildkite.com/%s/%s/builds/%d#%s", claims.OrganizationSlug, claims.PipelineSlug, claims.BuildNumber, claims.JobID),
+	}, nil
+}
+
+func (p jobPrincipal) Name(ctx context.Context) string {
+	return p.subject
+}
+
+func (p jobPrincipal) Embed(ctx context.Context, cert *x509.Certificate) error {
+	// Set job URL to SubjectAlternativeName on certificate
+	parsed, err := url.Parse(p.url)
+	if err != nil {
+		return err
+	}
+	cert.URIs = []*url.URL{parsed}
+
+	// Embed additional information into custom extensions
+	cert.ExtraExtensions, err = certificate.Extensions{
+		Issuer: p.issuer,
+	}.Render()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/pkg/identity/buildkite/principal.go
+++ b/pkg/identity/buildkite/principal.go
@@ -35,8 +35,8 @@ type jobPrincipal struct {
 	// https://agent.buildkite.com/.well-known/openid-configuration
 	issuer string
 
-	// The full URL to the job. This will be the set as SubjectAlternativeName URI in
-	// the final certificate.
+	// The URL of the pipeline, the container of many builds. This will be
+	// set as a human-friendly SubjectAlternativeName URI in the certificate.
 	url string
 }
 
@@ -44,8 +44,6 @@ func JobPrincipalFromIDToken(ctx context.Context, token *oidc.IDToken) (identity
 	var claims struct {
 		OrganizationSlug string `json:"organization_slug"`
 		PipelineSlug     string `json:"pipeline_slug"`
-		BuildNumber      int    `json:"build_number"`
-		JobID            string `json:"job_id"`
 	}
 	if err := token.Claims(&claims); err != nil {
 		return nil, err
@@ -59,18 +57,10 @@ func JobPrincipalFromIDToken(ctx context.Context, token *oidc.IDToken) (identity
 		return nil, errors.New("missing pipeline_slug claim in ID token")
 	}
 
-	if claims.BuildNumber == 0 {
-		return nil, errors.New("missing build_number claim in ID token")
-	}
-
-	if claims.JobID == "" {
-		return nil, errors.New("missing job_id claim in ID token")
-	}
-
 	return &jobPrincipal{
 		subject: token.Subject,
 		issuer:  token.Issuer,
-		url:     fmt.Sprintf("https://buildkite.com/%s/%s/builds/%d#%s", claims.OrganizationSlug, claims.PipelineSlug, claims.BuildNumber, claims.JobID),
+		url:     fmt.Sprintf("https://buildkite.com/%s/%s", claims.OrganizationSlug, claims.PipelineSlug),
 	}, nil
 }
 
@@ -79,7 +69,7 @@ func (p jobPrincipal) Name(ctx context.Context) string {
 }
 
 func (p jobPrincipal) Embed(ctx context.Context, cert *x509.Certificate) error {
-	// Set job URL to SubjectAlternativeName on certificate
+	// Set SubjectAlternativeName to the pipeline URL on the certificate
 	parsed, err := url.Parse(p.url)
 	if err != nil {
 		return err
--- a/pkg/identity/buildkite/principal.go
+++ b/pkg/identity/buildkite/principal.go
@@ -32,7 +32,7 @@ type jobPrincipal struct {
 	subject string
 
 	// OIDC Issuer URL. Matches 'iss' claim from ID token. The real issuer URL is
-	// https://agent.buildkite.com/.well-known/openid-configution
+	// https://agent.buildkite.com/.well-known/openid-configuration
 	issuer string
 
 	// The full URL to the job. This will be the set as SubjectAlternativeName URI in
--- a/pkg/identity/buildkite/principal.go
+++ b/pkg/identity/buildkite/principal.go
@@ -1,4 +1,4 @@
-// Copyright 2022 The Sigstore Authors.
+// Copyright 2023 The Sigstore Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/config/fulcio-config.yaml
+++ b/config/fulcio-config.yaml
@@ -22,6 +22,11 @@ data:
               "ClientID": "sigstore",
               "Type": "email"
             },
+            "https://agent.buildkite.com": {
+              "IssuerURL": "https://agent.buildkite.com",
+              "ClientID": "sigstore",
+              "Type": "buildkite-job"
+            },
             "https://allow.pub": {
               "IssuerURL": "https://allow.pub",
               "ClientID": "sigstore",
--- a/pkg/challenges/challenges.go
+++ b/pkg/challenges/challenges.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/sigstore/fulcio/pkg/config"
 	"github.com/sigstore/fulcio/pkg/identity"
+	"github.com/sigstore/fulcio/pkg/identity/buildkite"
 	"github.com/sigstore/fulcio/pkg/identity/email"
 	"github.com/sigstore/fulcio/pkg/identity/github"
 	"github.com/sigstore/fulcio/pkg/identity/kubernetes"
@@ -57,6 +58,8 @@ func PrincipalFromIDToken(ctx context.Context, tok *oidc.IDToken) (identity.Prin
 	var principal identity.Principal
 	var err error
 	switch iss.Type {
+	case config.IssuerTypeBuildkiteJob:
+		principal, err = buildkite.JobPrincipalFromIDToken(ctx, tok)
 	case config.IssuerTypeEmail:
 		principal, err = email.PrincipalFromIDToken(ctx, tok)
 	case config.IssuerTypeSpiffe:
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -208,6 +208,7 @@ func (fc *FulcioConfig) prepare() error {
 type IssuerType string
 
 const (
+	IssuerTypeBuildkiteJob   = "buildkite-job"
 	IssuerTypeEmail          = "email"
 	IssuerTypeGithubWorkflow = "github-workflow"
 	IssuerTypeKubernetes     = "kubernetes"
@@ -459,6 +460,8 @@ func validateAllowedDomain(subjectHostname, issuerHostname string) error {
 
 func issuerToChallengeClaim(issType IssuerType) string {
 	switch issType {
+	case IssuerTypeBuildkiteJob:
+		return "sub"
 	case IssuerTypeEmail:
 		return "email"
 	case IssuerTypeGithubWorkflow:
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -482,6 +482,9 @@ func Test_issuerToChallengeClaim(t *testing.T) {
 	if claim := issuerToChallengeClaim(IssuerTypeURI); claim != "sub" {
 		t.Fatalf("expected sub subject claim for URI issuer, got %s", claim)
 	}
+	if claim := issuerToChallengeClaim(IssuerTypeBuildkiteJob); claim != "sub" {
+		t.Fatalf("expected sub subject claim for Buildkite issuer, got %s", claim)
+	}
 	if claim := issuerToChallengeClaim(IssuerTypeGithubWorkflow); claim != "sub" {
 		t.Fatalf("expected sub subject claim for GitHub issuer, got %s", claim)
 	}
--- a/pkg/identity/buildkite/principal_test.go
+++ b/pkg/identity/buildkite/principal_test.go
@@ -41,10 +41,8 @@ func TestJobPrincipalFromIDToken(t *testing.T) {
 		`Valid token authenticates with correct claims`: {
 			Claims: map[string]interface{}{
 				"aud":               "sigstore",
-				"build_number":      123,
 				"exp":               0,
 				"iss":               "https://agent.buildkite.com",
-				"job_id":            "f81d4fae-7dec-11d0-a765-00a0c91e6bf6",
 				"organization_slug": "acme-inc",
 				"pipeline_slug":     "bash-example",
 				"sub":               "organization:acme-inc:pipeline:bash-example:ref:refs/heads/main:commit:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa:step:build",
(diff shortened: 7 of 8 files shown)


haydentherapper:
There’s a few options for onboarding. One is to conform to one of the existing types. Given you’re a CI platform, it’s likely the “GitHub” type is the closest. This would require just configuration change, but unless you support the same claims as GitHub Actions, we would need code changes. 
We’ve been thinking about how to standardize claims from CI platforms in https://github.com/sigstore/fulcio/issues/754.
The other option is to implement a new type specific to your platform, but we will need to also add custom OIDs too. I’d prefer to not do this for each CI platform though. 

I’m curious what your thoughts are on the attached issue.

haydentherapper:
We’d also want to run through the list of requirements for providers, https://github.com/sigstore/fulcio/issues/397. This is a WIP list, so some might not be applicable to CI. 

Let’s continue the conversation on the issue, I’ll copy this over.

sj26:
Yeah we don't support exactly the same claims as github, but it feels like there would be common ground.

We do include a unique immutable id (job_id, a uuid) which identifies a command and where it was run, and how it fits amongst a build (made of many jobs), pipeline and organization. Other provenance can be derived from that unique identifier, even if other symbolic attributes change over time. Perhaps that's the only hard requirement — there must be a unique immutable identifier for the runtime environment, and that identifier can be mapped to full provenance by the identity provider. I think this would be [the job_id on GitHub Actions](https://docs.github.com/en/rest/actions/workflow-jobs#get-a-job-for-a-workflow-run), for example.

I'm sure there are other commonly useful attributes, but they get pretty usage-specific pretty quickly.

mattmoor:
Once this lands, we should talk about adding ambient authentication to cosign for this so the `--identity-token` flag isn't needed.  That's done via the providers here: https://github.com/sigstore/cosign/tree/main/pkg/providers

haydentherapper:
Sweet, I’ll take a look tomorrow or Friday!

Could you take a look at the questions in https://github.com/sigstore/fulcio/issues/397 and answer them? In particular, confirming the well known endpoint, configurable audiences, secure key management of signing keys, uptime/SLO?

haydentherapper:
Also, what’s the best way to test this out?

sj26:
Thanks!

I'll address the specific questions you copied in here:

> confirming the well known endpoint

Our issuer is https://agent.buildkite.com, which is the private API endpoint used by the [Buildkite Agent](https://buildkite.com/docs/agent) (which runs jobs, and exchanges private credentials for OIDC identity tokens), and there are matching well-known endpoints here:

- https://agent.buildkite.com/.well-known/openid-configuration
- https://agent.buildkite.com/.well-known/jwks

> configurable audiences

The default audience is `https://buildkite.com/<organization>`, but this can be overridden when asking for a token.

In the original issue I included an example for the direct API which shows how to vary the audience:

```sh
curl -s -X POST -H "Authorization: Token $${BUILDKITE_AGENT_ACCESS_TOKEN}" \
  "$${BUILDKITE_AGENT_ENDPOINT:-https://agent.buildkite.com/v3}/jobs/$${BUILDKITE_JOB_ID}/oidc/tokens" \
  --data '{"audience":"sigstore"}'
```

but we expect folks to use a new subcommand of the buildkite agent (available within jobs) to ask for these tokens, which accepts an audience argument, like this:

```sh
buildkite-agent oidc request-token --audience sigstore
```

> secure key management of signing keys

Our keys are stored with other production secrets and treated similarly — encrypted at rest, encrypted in transit, only available within the application environment used for signing, etc.

The current cert has an expiry in 2 years, but I expect we'll rotate yearly. We're still refining these details for release.

> uptime/SLO?

We have a 99.95% uptime SLO.

sj26:
In terms of testing, I started a local ephemeral instance of fulcio with the buildkite provider configured using the instructions here:

https://github.com/sigstore/fulcio/blob/main/docs/setup.md

Then I ran a local buildkite agent on my mac:

https://buildkite.com/docs/agent/v3/macos

and had it run a pipeline like this:

```
- steps:
  - command: |
      docker push "sj26/test:$${BUILDKITE_BUILD_NUMBER}"
      docker tag sj26/test "sj26/test:$${BUILDKITE_BUILD_NUMBER}"
      BUILDKITE_OIDC_TOKEN="$$(curl -s -X POST -H "Authorization: Token $${BUILDKITE_AGENT_ACCESS_TOKEN}" "$${BUILDKITE_AGENT_ENDPOINT:-https://agent.buildkite.com/v3}/jobs/$${BUILDKITE_JOB_ID}/oidc/tokens" --data '{"audience":"sigstore"}' | jq -r .token)"
      COSIGN_EXPERIMENTAL=1 cosign sign --fulcio-url http://localhost:5555 --insecure-skip-verify --identity-token="$${BUILDKITE_OIDC_TOKEN}" "sj26/test:$${BUILDKITE_BUILD_NUMBER}"
```

again referencing the local fulcio usage example from:

https://github.com/sigstore/fulcio/blob/main/docs/setup.md#calling-the-fulcio-api

and .. it works:

image

This is a really rough get-it-working example, but it's enough to give me confidence that it's the right direction.

JAORMX (approved):

pkg/identity/buildkite/principal.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/sigstore/fulcio/pkg/identity"
+)
+
+type jobPrincipal struct {
+	// Subject matches the 'sub' claim from the OIDC ID token this is what is
+	// signed as proof of possession for Buildkite job identities
+	subject string
+
+	// OIDC Issuer URL. Matches 'iss' claim from ID token. The real issuer URL is
+	// https://agent.buildkite.com/.well-known/openid-configution
```
haydentherapper, line 35:
```suggestion
	// https://agent.buildkite.com/.well-known/openid-configuration
```

pkg/identity/buildkite/principal.go
```diff
@@ -0,0 +1,98 @@
...
+	issuer string
+
+	// The full URL to the job. This will be the set as SubjectAlternativeName URI in
+	// the final certificate.
+	url string
+}
+
+func JobPrincipalFromIDToken(ctx context.Context, token *oidc.IDToken) (identity.Principal, error) {
+	var claims struct {
```
haydentherapper, line 44: Do you have an example of a token? Curious to see what claims are available and what examples values look like.

pkg/identity/buildkite/principal.go
```diff
@@ -0,0 +1,98 @@
...
+		OrganizationSlug string `json:"organization_slug"`
+		PipelineSlug     string `json:"pipeline_slug"`
+		BuildNumber      int    `json:"build_number"`
```
haydentherapper, line 47: `build_number` is not one of the claims present in https://agent.buildkite.com/.well-known/openid-configuration - this might just be an error in the OIDC configuration

pkg/identity/buildkite/principal.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	if claims.JobID == "" {
+		return nil, errors.New("missing job_id claim in ID token")
+	}
+
+	return &jobPrincipal{
+		subject: token.Subject,
+		issuer:  token.Issuer,
+		url:     fmt.Sprintf("https://buildkite.com/%s/%s/builds/%d#%s", claims.OrganizationSlug, claims.PipelineSlug, claims.BuildNumber, claims.JobID),
```
haydentherapper, line 73: I'm interested in your thoughts on adding `build_branch` as part of the URL too. What we're looking for is a representation of a trust boundary. As a verifier, what do I want to know in order to determine if I trust this identity for a build? Going through the claims: 
* organization_slug - A verifier needs to know which org the repo is under (Do you have anything that represents repositories, or is that in organization_slug?)
* pipeline - I'm not familiar with the term, but I'll assume this is something like GH's workflow name, so you'd want to know what build pipeline ran (correct me if I'm wrong)
* build_branch - This seems important to me, to know if your workflow ran from a trusted branch. However, you could also argue this is provenance (@asraa, thoughts?)
* build_number/job_id - Does this need to be a part of the identity? Is it expected a verifier would know the build_number? Same question with job_id - These seem more like provenance than identity
* agent_id - I noticed this in the configuration - What does this represent?

cc @asraa who has thoughts about this for other CI platforms too

asraa, line 73: One comment I have is that I think that the build_number is extraneous: regardless of build number you could identify the build occurring, and the builder could always sign off on the build_number.

I otherwise think that using organization_slug/pipeline/job_id would be good, but that perhaps the build_branch should as well. this would make it align nicely with `job_workflow_ref` from GitHUb

haydentherapper: @sj26 Could you clarify the difference between pipeline and job id? I wonder if job_id is too fine grained, unless it's static for multiple runs of the workflow.

+1 on using build_branch as well. One open question is if we should also include build_commit somewhere in the certificate.

pkg/identity/buildkite/principal.go
sj26: Yeah I added it to the OIDC token but forgot the discovery endpoint, will update.

pkg/identity/buildkite/principal.go
sj26: Oh yeah! Here's an example from a test suite:

```json
{
  "iss": "http://agent.buildkite.localhost",
  "sub": "organization:acme-inc:pipeline:a38fb11dc0badd43e60bf1fbb1801f03d6309332:ref:refs/heads/main:commit:9f3182061f1e2cca4702c368cbc039b7dc9d4485:step:",
  "aud": "http://buildkite.localhost/acme-inc",
  "iat": 1669014898,
  "nbf": 1669014898,
  "exp": 1669015198,
  "organization_slug": "acme-inc",
  "pipeline_slug": "a38fb11dc0badd43e60bf1fbb1801f03d6309332",
  "build_number": 1,
  "build_branch": "main",
  "build_commit": "9f3182061f1e2cca4702c368cbc039b7dc9d4485",
  "step_key": null,
  "job_id": "0184990a-477b-4fa8-9968-496074483cee",
  "agent_id": "0184990a-4782-42b5-afc1-16715b10b8ff"
}
```

pkg/identity/buildkite/principal.go
sj26: Yeah, it feels important to identify the job and agent — the actual environment in which an artifact was produced — for provenance. But for verification you'd probably want to interrogate `organization_slug`, `pipeline_slug`, and _maybe_ `build_branch`. These are all available in the subject or subject alt, but are partial and so awkward. Having generic attributes for these would be great.

sj26: job_id would be too fine grained for what you're thinking of here I think.

There are cases where  you might not care about branch.

I think the ref in `job_workflow_ref` is about the source configuration, not the ref of the source code being built. Our equivalent is probably a pipeline, like `https://buildkite.com/buildkite/docs` for instance.

haydentherapper: Is BuildNumber correct here? This seems like provenance and not identity. Same question with JobID.

I think BuildBranch is part of the trust boundary, because you may have different protection levels for different branches. Used in conjunction with a commit hash (`build_commit`?), you have an immutable reference to a job configuration.

sj26: Pipeline, branch and commit are not enough for an immutable reference to job configuration because our configurations are dynamic and can be modified in ways not contained within the pipeline or repository, which is why I've referenced a particular job (which is itself part of a build, hence the build number — that's an actual URL where that job is accessible, and the final configuration and results can be interrogated).

If there were common branch and commit attributes/OIDs we could probably those as well,  for use cases where folks want to rely on branch and commit values? Or instead, if you think a reference to the concrete build and job are not desired in this case.

But shouldn't provenance be part of the signature as well? I've started there based on what seems to be available, and figured more attributes could be added as a common schema for CI providers emerges.

dlorenc:
This looks like a good first pass to me. What's missing to get it in?

haydentherapper:
Hey, sorry for the delay getting it reviewed. The remaining parts to review are:
* Finishing up the conversation on what needs to be a part of the subject - @asraa, did you have any other comments as to the format of the URL?
* I'd like one maintainer to check out this PR and test this is working as expected. @sj26 is there documentation for this feature?

pkg/identity/buildkite/principal.go
asraa: The `build_branch` is at least committed to the subject name. I think that even though it's "awkward" because it's present inside the whole string, I think it still suffices (we still do the same style parsing for GitHub refs based on the subject for that too)

asraa:
> Finishing up the conversation on what needs to be a part of the subject - @asraa, did you have any other comments as to the format of the URL?

The state right now seems like 
```
organization:project:pipeline:bash-example:ref:refs/heads/main:commit:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa:step:build
```

I think this is fine, although wordy, and justifiably because it's taking the sub claim anyway. In the GH world, the subjects also something fairly similar, identifying the reusable workflow like org/repo/workflow/path@ref.

sj26:
Totally agree @mattmoor, but in OIDC relying parties like AWS IAM there is only the ability to make assertions on the subject, not other claims or oid-equivalents, so all the details need to be crammed in there so that string matching is possible.

sj26:
> I'd like one maintainer to check out this PR and test this is working as expected. @sj26 is there documentation for this feature?

I pushed up docs for the buildkite agent oidc command line the other day here:
https://buildkite.com/docs/agent/v3/cli-oidc

The instructions I added earlier demonstrate how to use that command to produce an OIDC token and feed it into cosign with a local fulcio:
https://github.com/sigstore/fulcio/pull/890#issuecomment-1318230512

The only missing bits are that you need to sign up for a (free) buildkite account to create a pipeline:
https://buildkite.com

I'm happy to jump on a call to walk anybody through the details:
https://calendly.com/sam-buildkite/30min

mattmoor:
> but in OIDC relying parties like AWS IAM there is only the ability to make assertions on the subject, not other claims or oid-equivalents

Yeah, I'm aware.  FWIW, that presents similar challenges on the IAM trust relationship side to what I mention on the signature verification side.  However, I guess that's irrelevant since this is neither an email or URL, which are the two forms of ~identity that fulcio/cosign generally use in certificates today, so I guess the question I have is what we intend that sense of identity to be for BuildKite tokens?  Is it this (from your unit tests)?
```
https://buildkite.com/acme-inc/bash-example/builds/123#f81d4fae-7dec-11d0-a765-00a0c91e6bf6
```

I'm wondering what `cosign verify ` will show as `Subject:` once this lands, since that's what'll go into policies, and I think that should be stable, but it's hard to tell how often that UUID changes...  🤔 

As context, we have customers that are keen to use BuildKite with `sigstore/policy-controller`, so I'm just trying to get a sense for the shape of policies they'll need to write to verify signatures produced with BuildKite tokens.

sj26:
Hm, yeah, I imagine a consume of a container image signed by a github action wouldn't be using their subject format, but would want to do something like:

```
cosign verify sj26/test123 \
  --certificate-github-workflow-repository acme-inc/super-app \
  --certificate-github-workflow-ref refs/heads/main
```

so the equivalent would theoretically be something like:

```
cosign verify sj26/test123 \
  --certificate-buildkite-pipeline acme-inc/super-app \
  --certificate-git-ref refs/heads/main
```

The current OIDs don't allow encoding this sort of information. I guess that's where the conversation about standardising the OIDs populated by CI providers (https://github.com/sigstore/fulcio/issues/754) becomes interesting. Could there be a generic repository/workflow/pipeline identity attribute? Same for git ref and sha?

mattmoor:
My point is a little more subtle...  The "subject" used for actions (that shows up in `cosign verify` output) isn't actually the token's subject because this must be either an email address or URL shape and neither Github, not your token satisfies that.

For Github (Kubernetes does something similar), we synthesize a URL from the `job_workflow_ref`, so I am curious what Email/URL you actually plan to use?  Here's an example from a policy we use for an actions-produced image:
```hcl
issuer  = "https://token.actions.githubusercontent.com"
subject = "https://github.com/chainguard-images/images/.github/workflows/release.yaml@refs/heads/main"
```

sj26:
The subject from the oidc token is used, but there's a subject alt name which is the url to the specific job which would be unique every time a command is run:

```
https://buildkite.com/acme-inc/bash-example/builds/123#f81d4fae-7dec-11d0-a765-00a0c91e6bf6
```

Should this instead be the pipeline, the container of all builds and jobs which have or could run that pipeline?

```
https://buildkite.com/acme-inc/bash-example
```

Or would you have a way to match a subject prefix to do the same from the more-specific subject?

haydentherapper:
@mattmoor @sj26 Were there any more follow ups on SAN of the certificate?

mattmoor:
@sj26 I think that for the SAN that `cosign` shows as the "subject" to be useful for policy evaluation it should be something stable, and approachable by humans.  e.g. for Google it's the user's email address (vs. the large internal number Google uses for `sub`), for Github it's a derivative of the job_workflow_ref claim, for K8s it is based on the service account namespace and name.

I think that if the SAN is distinct with every pipeline execution, then verifying BuildKite signatures is going to end up be different than most other systems in downstream policy tools like `sigstore/policy-controller`, Kyverno, etc.

I'd be happy to hop on a call sometime next week when I'm back from Holiday if it'd be helpful.  I'm (normally) on the US west coast, and you can reach me at `mattmoor` on sigstore slack or grab some time via the same username on `calend.ly`.

sj26:
Thanks for the input, and sorry for the delay — I had a holiday!

I've updated the SAN to be the pipeline URL, so it will be consistent across builds and useful for writing policies. Branch/commit/etc can be introduced using whatever common attributes for CI become available over time.

haydentherapper:
@sj26 Thanks for your patience! I tested this out, it looks great! 

Steps to test:

* Create Buildkite account, set up local agent and pipeline
* Add step: `buildkite-agent oidc request-token --audience "sigstore"`, and copy output to `id_token`

Generate proof of possession:

* Generate key: `openssl ecparam -genkey -name secp384r1 > ec_private.pem && openssl ec -in ec_private.pem -pubout > ec_public.pem`
* `step crypto jwt inspect --insecure < id_token | jq -r .payload.sub | tr -d '\n' > sub`
* Self-sign: `openssl dgst -sha256 -sign ec_private.pem sub > sub.sig`
* Validate: `openssl dgst -sha256 -verify ec_public.pem -signature sub.sig sub`

For testing Fulcio:

* Run Fulcio locally, updating configuration to include buildkite OIDC provider: `go run main.go serve --port 5555 --ca ephemeralca --ct-log-url="" --config-path config/config.jsn`
* `curl -s http://localhost:5555/api/v1/signingCert -H "Authorization: Bearer $(cat id_token)" -H "Accept: application/pem-certificate-chain" -H "Content-Type: application/json" -o signingCertChain.pem --data-binary "{ \"publicKey\": { \"algorithm\": \"ecdsa\", \"content\": \"$(step crypto key format ec_public.pem|base64 -w0)\" }, \"signedEmailAddress\": \"$(base64 -w0 sub.sig)\" }"`
* Parse cert chain: `openssl crl2pkcs7 -nocrl -certfile signingCertChain.pem | openssl pkcs7 -print_certs -text -noout`

Output:

```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            15:29:18:07:37:66:75:99:94:8b:c3:9d:f7:79:b2:ab:d1:1c:cc:4f
        Signature Algorithm: ecdsa-with-SHA256
        Issuer: C=USA, ST=WA, L=Kirkland/street=767 6th St S/postalCode=98033, O=sigstore
        Validity
            Not Before: Jan 12 21:54:33 2023 GMT
            Not After : Jan 12 22:04:33 2023 GMT
        Subject:
        Subject Public Key Info:
            Public Key Algorithm: id-ecPublicKey
                Public-Key: (384 bit)
                pub:
                    04:5d:52:11:55:97:9e:d7:bf:a9:9c:e2:3e:bf:30:
                    44:3a:d1:6e:68:56:ed:10:1f:15:5d:33:cf:35:5f:
                    1c:37:d1:b6:80:8e:f6:41:fd:a9:0b:02:30:1d:4c:
                    f6:29:d6:7f:b0:f0:52:8e:e9:e7:10:b2:a4:10:8b:
                    9a:74:5c:98:92:cc:98:67:7a:88:38:46:1d:79:7d:
                    a9:4f:8e:74:40:40:08:a3:8c:f9:d5:88:9e:09:cf:
                    49:4f:21:3b:44:4e:01
                ASN1 OID: secp384r1
                NIST CURVE: P-384
        X509v3 extensions:
            X509v3 Key Usage: critical
                Digital Signature
            X509v3 Extended Key Usage:
                Code Signing
                33:8B:51:46:67:B0:41:C2:7A:5E:85:91:A3:15:8C:54:25:48:C5:78
                E1:1F:EC:27:36:63:FB:BB:AA:CB:BE:34:DE:FA:52:2E:23:7F:5D:F5
                URI:https://buildkite.com/test-fulcio/test
            1.3.6.1.4.1.57264.1.1:
                https://agent.buildkite.com
    Signature Algorithm: ecdsa-with-SHA256
    Signature Value:
        30:45:02:21:00:dd:30:2b:c1:d2:3f:0c:70:f5:32:3f:1c:3d:
        71:c5:71:9a:30:b1:72:ce:6a:19:08:4b:a4:90:ba:d5:0e:7e:
        bc:02:20:1b:c4:0f:09:4e:1d:09:5c:74:43:02:64:79:9c:aa:
        8a:01:51:b2:cd:9e:54:d6:a2:17:13:87:62:f5:65:cc:4f

Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            27:8c:b6:ad:f6:c3:3d:75:57:02:11:83:26:6c:75:56:1b:5e:db:33
        Signature Algorithm: ecdsa-with-SHA256
        Issuer: C=USA, ST=WA, L=Kirkland/street=767 6th St S/postalCode=98033, O=sigstore
        Validity
            Not Before: Jan 12 21:54:02 2023 GMT
            Not After : Jan 12 21:54:02 2033 GMT
        Subject: C=USA, ST=WA, L=Kirkland/street=767 6th St S/postalCode=98033, O=sigstore
        Subject Public Key Info:
            Public Key Algorithm: id-ecPublicKey
                Public-Key: (256 bit)
                pub:
                    04:64:88:50:80:b7:38:14:b8:0a:e2:8e:53:71:01:
                    1b:35:71:b4:ad:28:84:f9:1e:85:29:41:62:71:60:
                    2b:78:a0:63:b9:3c:4e:11:7f:eb:29:09:48:69:b1:
                    70:d7:08:34:74:1e:ee:a0:fd:f7:0a:e3:6d:5e:cd:
                    34:f4:cf:b9:29
                ASN1 OID: prime256v1
                NIST CURVE: P-256
        X509v3 extensions:
            X509v3 Key Usage: critical
                Certificate Sign, CRL Sign
            X509v3 Basic Constraints: critical
                CA:TRUE, pathlen:1
                E1:1F:EC:27:36:63:FB:BB:AA:CB:BE:34:DE:FA:52:2E:23:7F:5D:F5
    Signature Algorithm: ecdsa-with-SHA256
    Signature Value:
        30:45:02:20:56:a6:c5:87:ad:50:62:48:2d:b7:6f:06:3e:ce:
        62:49:02:5d:91:a4:50:1c:38:ec:01:00:7c:53:a5:eb:93:fe:
        02:21:00:d5:5c:6d:01:a7:94:01:30:d3:92:dc:26:19:81:ca:
        31:ba:b8:2f:af:5e:87:f3:bd:d4:ac:31:6b:b4:4f:60:fa
```

SAN is `URI:https://buildkite.com/test-fulcio/test`, the organization and pipeline slug, as expected.

haydentherapper:
I'll merge this in a day or so in case there's any more comments.

cc @asraa @bobcallaway @mattmoor if you have any more comments

haydentherapper (changes requested):
@sj26 Can you add a few tests to grpc_server_test, which tests e2e flows, and one other file?

* Add a test for the new issuer string - https://github.com/sigstore/fulcio/blob/main/pkg/config/config_test.go#L472
* Add a configuration test - https://github.com/sigstore/fulcio/blob/main/pkg/server/grpc_server_test.go#L174
* Add a test for issuing a certificate, like https://github.com/sigstore/fulcio/blob/main/pkg/server/grpc_server_test.go#L681-L812

sj26:
@haydentherapper thanks for taking another look, I've added those tests. While in there, I noticed the github api test was conflating subject and url so I made them match the principal tests.

haydentherapper (commented):
Thanks, this looks great! Just a linter failure

sj26:
@haydentherapper done, I think!

sj26:
There was a failing test hidden in the depths, so I did a rebase to clean up slightly and include a fix.

haydentherapper (dismissed):
Thanks for your work on this! Ill get this merged in and rolled out to staging this week, then you can test against it. URL is fulcio.sigstage.dev, and you’ll need to initialize your cosign environment with a different TUF root - https://github.com/sigstore/cosign/blob/main/KEYLESS.md#usage-1

federation/agent.buildkite.com/config.yaml
```diff
@@ -0,0 +1,18 @@
+# Copyright 2022 The Sigstore Authors
```
bobcallaway, line 1: nit: s/2022/2023 throughout for all new files

bobcallaway (commented):
one nit on copyright dates but otherwise LGTM

federation/agent.buildkite.com/config.yaml
haydentherapper: Are you sure, it still feels like 2022 to me...

sj26: It was 2022 when I wrote it 😅 but I'll update for merge

dlorenc (approved):

mattmoor:
@sj26 One thing I wanted to plug now that this has landed is that `cosign` supports "ambient credential detection" where it keys off of hints in the environment to detect the presence of certain credential providers.  e.g. this is how `cosign sign ` "just works" in Github Actions when `id-token: write` is specified.

Your PR unblocks `cosign sign --identity-token {tok} `, but if there are environmental factors to key off of, then we could add a buildkite provider here to tighten this up: https://github.com/sigstore/cosign/tree/main/pkg/providers

Happy to provide reviews/guidance, or elaborate on this further if you have questions.

PR #892, opened 2022-11-17
Expose client options for google ca

priyawadhwa:
I was playing around with some stuff and realized I needed to pass in some extra options. This just exposes the google options so that folks using this as a library can pass them in. 


haydentherapper (approved):
Thanks!

PR #897, opened 2022-11-21
Fix bugs in googleca and update flag description

priyawadhwa:
This fixes 2 bugs in the googleca:
1. We would always fail in `TrustBundle` because we were checking `if len(roots) == 0` instead of `if len(caCerts) == 0`, roots was always empty at this point
2. Allow specifying certificateAuthority in the `gcp_private_ca_parent ` flag

#### Release Note
Fix bugs in googleca and update flag description

### DIFF
--- a/cmd/app/serve.go
+++ b/cmd/app/serve.go
@@ -60,7 +60,7 @@ func newServeCmd() *cobra.Command {
 	cmd.Flags().String("log_type", "dev", "logger type to use (dev/prod)")
 	cmd.Flags().String("ca", "", "googleca | tinkca | pkcs11ca | fileca | kmsca | ephemeralca (for testing)")
 	cmd.Flags().String("aws-hsm-root-ca-path", "", "Path to root CA on disk (only used with AWS HSM)")
-	cmd.Flags().String("gcp_private_ca_parent", "", "private ca parent: /projects/<project>/locations/<location>/<name> (only used with --ca googleca)")
+	cmd.Flags().String("gcp_private_ca_parent", "", "private ca parent: projects/<project>/locations/<location>/caPools/<caPool>/certificateAuthorities/<certificateAuthority> (only used with --ca googleca)")
 	cmd.Flags().String("hsm-caroot-id", "", "HSM ID for Root CA (only used with --ca pkcs11ca)")
 	cmd.Flags().String("ct-log-url", "http://localhost:6962/test", "host and path (with log prefix at the end) to the ct log")
 	cmd.Flags().String("ct-log-public-key-path", "", "Path to a PEM-encoded public key of the CT log, used to verify SCTs")
--- a/cmd/app/serve.go
+++ b/cmd/app/serve.go
@@ -61,7 +61,7 @@ func newServeCmd() *cobra.Command {
 	cmd.Flags().String("ca", "", "googleca | tinkca | pkcs11ca | fileca | kmsca | ephemeralca (for testing)")
 	cmd.Flags().String("aws-hsm-root-ca-path", "", "Path to root CA on disk (only used with AWS HSM)")
 	cmd.Flags().String("gcp_private_ca_parent", "", "private ca parent: projects/<project>/locations/<location>/caPools/<caPool> (only used with --ca googleca)"+
-		"Optionally specify /certificateAuthorities/*, which will bypass CA pool load balancing.")
+		"Optionally specify /certificateAuthorities/<caID>, which will bypass CA pool load balancing.")
 	cmd.Flags().String("hsm-caroot-id", "", "HSM ID for Root CA (only used with --ca pkcs11ca)")
 	cmd.Flags().String("ct-log-url", "http://localhost:6962/test", "host and path (with log prefix at the end) to the ct log")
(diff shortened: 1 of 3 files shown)


cmd/app/serve.go
```diff
@@ -60,7 +60,7 @@ func newServeCmd() *cobra.Command {
 	cmd.Flags().String("log_type", "dev", "logger type to use (dev/prod)")
 	cmd.Flags().String("ca", "", "googleca | tinkca | pkcs11ca | fileca | kmsca | ephemeralca (for testing)")
 	cmd.Flags().String("aws-hsm-root-ca-path", "", "Path to root CA on disk (only used with AWS HSM)")
-	cmd.Flags().String("gcp_private_ca_parent", "", "private ca parent: /projects/<project>/locations/<location>/<name> (only used with --ca googleca)")
```
haydentherapper, line 63: Using the pool rather than the CA is an intentional decision, so that fulcio operators can increase QPS by simply adding more CAs to the pool. 

List is working as intended (besides the len(roots) bug) - For each CA in the pool, it grabs the CA chain, and appends it to a list of chains. 

The problem was that the v1 GetRootCertificate endpoint never properly supported lists of chains, so all certs would just be concatenated together. As long as you had just one CA per pool, this was fine. For v2, [TrustBundle](https://github.com/sigstore/fulcio/blob/main/fulcio.proto#L195) returns a list of list of certificates, so it works well with the pool concept.

haydentherapper (changes requested):

cmd/app/serve.go
priyawadhwa: Thanks for the background, I changed the flag description back to what it was. I also kept the support for specifying certificateAuthority, so users can choose either option now.

haydentherapper (changes requested):

pkg/ca/googleca/v1/googleca.go
```diff
@@ -52,10 +53,19 @@ func NewCertAuthorityService(ctx context.Context, parent string, opts ...option.
...
+	if !strings.Contains(parent, "certificateAuthorities") {
+		c.caPoolParent = parent
+	} else {
+		// parent should be in the form projects/*/locations/*/caPools/*/certificateAuthorities/*
+		// to create a cert, we only want projects/*/locations/*/caPools/*
+		caPoolParent := strings.Split(parent, "/certificateAuthorities")
+		c.caPoolParent = caPoolParent[0]
+		c.certAuthorityParent = parent
+	}
+	return &c, nil
```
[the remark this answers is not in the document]
priyawadhwa: thanks! added that in

cmd/app/serve.go
```diff
@@ -60,7 +60,8 @@ func newServeCmd() *cobra.Command {
...
+	cmd.Flags().String("gcp_private_ca_parent", "", "private ca parent: projects/<project>/locations/<location>/caPools/<caPool> (only used with --ca googleca)"+
+		"Optionally specify /certificateAuthorities/*, which will bypass CA pool load balancing.")
```
haydentherapper, line 64:
```suggestion
		"Optionally specify /certificateAuthorities/<caID>, which will bypass CA pool load balancing.")
```

haydentherapper: nit to match style of commebt

pkg/ca/googleca/v1/googleca.go
```diff
@@ -39,8 +39,10 @@ import (
 )
 
 type CertAuthorityService struct {
-	parent string
-	client *privateca.CertificateAuthorityClient
+	certAuthority       string
+	certAuthorityParent string
```
haydentherapper, line 43: I'd merge these two variables into one, `caPool`, since they represent the same resource. You can then check if `certAuthority` is set to decide whether or not to target a specific CA

haydentherapper (commented):
just a couple comments, looks good!

pkg/ca/googleca/v1/googleca.go
priyawadhwa: So we do need both of them later on, the full certAuthorityParent in the `GetCertificateAuthority` request and just the name of the CA `certAuthority` when actually making the request for `IssuingCertificateAuthorityId`.

i think it's more understandable to do all the validation/string splitting logic at the beginning and then have everything we need later on when actually getting CA's/making requests.

haydentherapper: Oh I see my confusion, it's around terminology. I thought two referred to the same resource. I would use the terms from [aip.dev](https://google.aip.dev/122) to describe the resources and IDs here:
* certAuthority -> certAuthorityID (since it represents just an ID)
* certAuthorityParent -> certAuthorityResource (since it's the full resource string)
* caPoolParent -> caPoolResource

Does that sound good? "parent" was what was confusing, since a parent of the a resource is a different resource.

priyawadhwa: Sure, that's fine. I'd still keep "parent" personally to follow the google naming standards for a googleca, but am fine with changing it too. It's updated now.

haydentherapper (approved):

PR #915, opened 2022-12-11
Set odic email contacts / uris the same

lukehinds:

### DIFF
--- a/config/fulcio-config.yaml
+++ b/config/fulcio-config.yaml
@@ -34,11 +34,11 @@ data:
               "Type": "email",
               "IssuerClaim": "$.federated_claims.connector_id"
             },
-            "https://oidc.dlorenc.dev": {
-              "IssuerURL": "https://oidc.dlorenc.dev",
+            "https://oidc.sigstore.dev": {
+              "IssuerURL": "https://oidc.sigstore.dev",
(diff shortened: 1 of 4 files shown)


haydentherapper (changes requested):
I don’t believe this is a valid domain, unless we’ve spun one up that’s I’m not aware of.

I don’t know if Dan still is hosting this example either. could we just remove it?

lukehinds:
I don't believe they are meant to resolve anything, but at the time we did not have sigstore.dev, so figured now we do, may as well have them set to our domain, so anyone can figure out the originating service.

haydentherapper:
These are meant to be the production values, not just examples, so we should have a resolvable service hosted at this endpoint if we specify it. We copy this config over to production. I think we should just remove the dlorenc SPIFFE endpoint since no one should be using that anyways. 

The contact updates LGTM.

haydentherapper:
These are meant to be the production values, not just examples, so we should have a resolvable service hosted at this endpoint if we specify it. We copy this config over to production. I think we should just remove the dlorenc SPIFFE endpoint since no one should be using that anyways. 

The contact updates LGTM.

haydentherapper:
Closing in favor of https://github.com/sigstore/fulcio/pull/932

PR #916, opened 2022-12-11
Incorrect URL points to malicious site (in commented code)

lukehinds:
Nothing major, as its only a comment.

I happened to be looking for the well-known config of github, so searched around in fulcio. Found the one to */githubcontent.com/ , tried to curl and got a redirect to some not very friendly looking site.

This changes it to the correct site.


haydentherapper (approved):
Thanks!

PR #926, opened 2022-12-15
proto, pkg, docs: update CSR docstrings

### LINKED ISSUE #863: Ambiguous CSR behavior: Fulcio accepts CSRs with email subjects that aren't emails
I noticed this while doing a read-through of `sigstore-python`'s signing code: we currently construct a CSR like so:

```python
        oidc_identity = Identity(identity_token)

        # Build an X.509 Certificiate Signing Request
        builder = (
            x509.CertificateSigningRequestBuilder()
            .subject_name(
                x509.Name(
                    [
                        x509.NameAttribute(NameOID.EMAIL_ADDRESS, oidc_identity.proof),
                    ]
                )
            )
            .add_extension(
                x509.BasicConstraints(ca=False, path_length=None),
                critical=True,
            )
        )
        certificate_request = builder.sign(private_key, hashes.SHA256())
```

...where `oidc_identity.proof` is the value of either the `email` or `sub` claim, depending on the kind of OIDC JWT being used. 

For OIDC JWTs that attest to email identities, this makes sense: we're sending a CSR with a `SubjectName` that's an email address. It doesn't make sense for OIDC JWTs that attest to URLs, since those aren't email addresses.

However, it currently works for both: `sigstore-python` currently uses `NameOID.EMAIL_ADDRESS` unconditionally, and Fulcio has been happily minting certs even when the `SubjectName` is really a URL (like a GitHub Actions workflow reference).

This probably isn't exploitable or dangerous in any way, since the CSR is otherwise well-formed and accompanied by a valid OIDC JWT *and* proof of possession for the signing key. But it's a little bit surprising, and suggests that Fulcio's CSR validation is a little looser than it should be.

IMO, there are (at least) two acceptable alternative behaviors here:

* Ignore the `SubjectName` in the CSR entirely, and (after a deprecation period) have users set it to something well-known, like `sigstore-csr` instead of an email address with an email OID. Then, explicitly document the behavior of re-deriving the subject from the adjacent JWT.
* Continue to handle `SubjectName`, but enforce the correct OID type for the content of the subject. In other words: `SubjectName`s that are tagged as emails should look like emails, URLs should be `UNSTRUCTURED_NAME` or something similar, or maybe something custom.

woodruffw:
Closes #863.

Signed-off-by: William Woodruff <william@trailofbits.com>

cc @haydentherapper


haydentherapper (dismissed):

pkg/generated/protobuf/fulcio.pb.go
```diff
@@ -16,7 +16,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.28.1
-// 	protoc        v3.20.3
+// 	protoc        v3.21.11
```
haydentherapper, line 19: Ah, this is gonna be an issue. My recommendation is to either update https://github.com/sigstore/fulcio/blob/d1fa542e81caff3583337a990898bcd701f0aa7f/.github/workflows/main.yml#L41, or install 3.20.3. We need to use a containerized protoc, we have an issue tracking this

woodruffw: My `protoc` is from `brew`, so downgrading might be a little painful 😅 -- I'll bump the action if that's alright.

woodruffw: https://github.com/sigstore/fulcio/pull/926/commits/05bc547af17d3e69919a7702b5a1a84acdff35a7

haydentherapper: Hm, it seems to think that version doesn't exist.

woodruffw: Yeah...looks like manifestation of this issue: https://github.com/arduino/setup-protoc/issues/33

woodruffw: (It seems like protobuf changed its versioning to chop the `3.` prefix off, and this action doesn't know about that.)

woodruffw: I opened https://github.com/arduino/setup-protoc/pull/58 with a fix to that action, and I've temporarily updated this PR to pin my `trail-of-forks` fork instead. But I assume we'll want this to be upstreamed before merging 🙂

woodruffw:
The diff is still failing, for reasons that aren't clear to me...is it possible `protoc` behaves differently on macOS vs. Linux in terms of comment generation?

haydentherapper:
@woodruffw Are you using Go 1.19.4? Differences in Go versions can also cause issues

woodruffw:
> @woodruffw Are you using Go 1.19.4? Differences in Go versions can also cause issues

`go version go1.18.2 darwin/amd64` according to `go version`, so pretty old 😅. I'll update and re-build.

.github/workflows/main.yml
```diff
@@ -35,10 +35,10 @@ jobs:
       - uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
         with:
           go-version: ${{ env.GOVERSION }}
-      - uses: arduino/setup-protoc@64c0c85d18e984422218383b81c52f8b077404d3 # v1.1.2
+      - uses: trail-of-forks/setup-protoc@a97892a429d98fae78d26f40334ab7eb616d08b9
```
woodruffw, line 38: N.B. that this is still using my fork of the `setup-protoc` action, since they haven't merged my fixes yet.

haydentherapper: I'm good to use your fork until they do, thanks for getting this working!

haydentherapper (approved):

PR #930, opened 2022-12-19
googleca: close certificate authority client when done

hectorj2f:

#### Summary

We realized that `NewCertAuthorityService` was creating a certificate authority client but the returned client **was not closed**  when it was done being used. This is needed to clean up its underlying connections. I added a new interface function so any third party (consumer) could close the connection when they are done.

#### Release Note

Close certificate authority client when done.

### DIFF
--- a/cmd/app/serve.go
+++ b/cmd/app/serve.go
@@ -195,6 +195,7 @@ func runServeCmd(cmd *cobra.Command, args []string) {
 	switch viper.GetString("ca") {
 	case "googleca":
 		baseca, err = googlecav1.NewCertAuthorityService(cmd.Context(), viper.GetString("gcp_private_ca_parent"))
+		defer baseca.Close()
 	case "pkcs11ca":
 		params := pkcs11ca.Params{
 			ConfigPath: viper.GetString("pkcs11-config-path"),
--- a/cmd/app/serve.go
+++ b/cmd/app/serve.go
@@ -195,7 +195,6 @@ func runServeCmd(cmd *cobra.Command, args []string) {
 	switch viper.GetString("ca") {
 	case "googleca":
 		baseca, err = googlecav1.NewCertAuthorityService(cmd.Context(), viper.GetString("gcp_private_ca_parent"))
-		defer baseca.Close()
 	case "pkcs11ca":
 		params := pkcs11ca.Params{
 			ConfigPath: viper.GetString("pkcs11-config-path"),
@@ -235,6 +234,7 @@ func runServeCmd(cmd *cobra.Command, args []string) {
 	if err != nil {
 		log.Logger.Fatal(err)
 	}
+	defer baseca.Close()
 
 	var ctClient *ctclient.LogClient
 	if logURL := viper.GetString("ct-log-url"); logURL != "" {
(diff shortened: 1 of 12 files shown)


cmd/app/serve.go
```diff
@@ -195,6 +195,7 @@ func runServeCmd(cmd *cobra.Command, args []string) {
 	switch viper.GetString("ca") {
 	case "googleca":
 		baseca, err = googlecav1.NewCertAuthorityService(cmd.Context(), viper.GetString("gcp_private_ca_parent"))
+		defer baseca.Close()
```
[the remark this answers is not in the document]
hectorj2f: I believe I am not following you. What do you want before the switch ? Do you mean to add it here https://github.com/sigstore/fulcio/pull/930/files#diff-1627906dbce532aff8cc81f54790572e5ba04f3e568c88b6ed39a590642dafa4R141 ?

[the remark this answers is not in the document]
priyawadhwa: sorry it wasn't clear! maybe something like:

```golang
var baseca certauth.CertificateAuthority
switch viper.GetString("ca") {
 ... // all of the switch statement stuff
}
defer baseca.Close()
```

so that Close() is called regardless of which CA we use.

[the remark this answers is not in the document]
hectorj2f: Done, thanks :).

cmd/fetch_ca_cert/fetch_ca_cert.go
```diff
@@ -188,6 +188,7 @@ func main() {
 
 	client, err := privateca.NewCertificateAuthorityClient(context.Background())
 	if err != nil {
+		client.Close()
```
priyawadhwa, line 191: We probably want to close the client whether this function returns an error or not. Maybe put this as a defer statement above as well?

hectorj2f: Good catch!

haydentherapper:
Can we implement `Close()` in https://github.com/sigstore/fulcio/blob/main/pkg/ca/baseca/baseca.go, and then only reimplement it in `googleca`?

hectorj2f:
@haydentherapper @priyawadhwa I changed the implementation to follow the same approach we followed with `TrustBundle`.

cmd/fetch_ca_cert/fetch_ca_cert.go
```diff
@@ -188,19 +188,24 @@ func main() {
...
 		log.Fatal(err)
 	}
 	parsedCerts, err := fetchCACertificate(context.Background(), *gcpCaParent, *kmsKey, *tinkKeysetPath, *tinkKmsKey, client)
```
priyawadhwa, line 194: I think we can probably just put the `defer client.Close()` call here after the error check once and then remove all the other calls to `client.Close()`

pkg/ca/fileca/fileca.go
```diff
@@ -60,6 +60,10 @@ func NewFileCA(certPath, keyPath, keyPass string, watch bool) (ca.CertificateAut
 	return &fca, err
 }
 
+func (fca *fileCA) Close() error {
```
priyawadhwa, line 63: I think this can probably be removed since it'll inherit from baseca

priyawadhwa (commented):
This is looking good, just a couple more comments!

cmd/fetch_ca_cert/fetch_ca_cert.go
haydentherapper: IIRC the problem is that log.Fatal ends execution immediately so the deferral doesn't happen. I think the linter complains about the lack of closure in that case. I ran into this with closing files in other tests

hectorj2f: No, this is due to the nature of log.Fatal. It won't run the defer.

haydentherapper (approved):

PR #931, opened 2022-12-19
Add in --duplex flag to run HTTP and GRPC servers on the same port

priyawadhwa:
also adds in a unit test for this. we can optionally start trying this out in staging, and if it works well we can consider deprecating the previous two-port solution.

#### Release Note
Add in --duplex flag to run HTTP and GRPC servers on the same port

cc @vaikas


haydentherapper:
Thanks! Let's hold this until the new year, and we can cut a new release at the same time.

haydentherapper:
@priyawadhwa Is there any change in behavior between serving gRPC on its own port and duplex? Any differences for metrics?

priyawadhwa:
There shouldn't be!

cmd/app/serve.go
```diff
@@ -85,6 +100,7 @@ func newServeCmd() *cobra.Command {
 	cmd.Flags().String("grpc-port", "8081", "The port on which to serve requests for GRPC")
 	cmd.Flags().String("metrics-port", "2112", "The port on which to serve prometheus metrics endpoint")
 	cmd.Flags().Duration("read-header-timeout", 10*time.Second, "The time allowed to read the headers of the requests in seconds")
+	cmd.Flags().Bool("duplex", false, "experimental: serve HTTP and GRPC on the same port instead of on two separate ports")
```
vaikas, line 103: Should we also mark grpc-host/grpc-port deprecated in their flag descriptions? Then maybe later on we can use something like:
https://pkg.go.dev/go.etcd.io/etcd/pkg/flags#DeprecatedFlag

vaikas (commented):
Awesomesauce!

cmd/app/serve.go
priyawadhwa: Added! I also added in a warning to suggest to users that they switch to using the --duplex flag if it isn't set.

vaikas: Thanks!, that's a great idea. wonder if we should do something about the http-host / http-port aliases as well, since with duplex they are kind of wonky? Up to you, but now that I think about it, in the future they too should be deprecated, removed and only use host/port?

priyawadhwa: I think it makes sense to remove the aliases! They were added in back in April (https://github.com/sigstore/fulcio/pull/472) so I think it's been enough time since then. I guess whenever we actually remove grpc-port grpc-host we can remove http-port and http-host as well.

vaikas (approved):

haydentherapper:
Thanks y’all, I’ll test out these changes after the holidays!

cmd/app/serve.go
```diff
@@ -81,10 +96,11 @@ func newServeCmd() *cobra.Command {
 	cmd.Flags().String("tink-keyset-path", "", "Path to KMS-encrypted keyset for Tink-backed CA")
 	cmd.Flags().String("host", "0.0.0.0", "The host on which to serve requests for HTTP; --http-host is alias")
 	cmd.Flags().String("port", "8080", "The port on which to serve requests for HTTP; --http-port is alias")
-	cmd.Flags().String("grpc-host", "0.0.0.0", "The host on which to serve requests for GRPC")
-	cmd.Flags().String("grpc-port", "8081", "The port on which to serve requests for GRPC")
+	cmd.Flags().String("grpc-host", "0.0.0.0", "[DEPRECATED] The host on which to serve requests for GRPC")
```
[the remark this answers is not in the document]
priyawadhwa: If we eventually plan to remove them (which I think makes the most sense) then it's best to deprecate them sooner. IIUC we'll still maintain support for at least a few releases or a few months (whatever the policy comes up with suggests).

cmd/app/serve.go
```diff
@@ -81,10 +96,11 @@ func newServeCmd() *cobra.Command {
...
+	cmd.Flags().String("grpc-port", "8081", "[DEPRECATED] The port on which to serve requests for GRPC")
 	cmd.Flags().String("metrics-port", "2112", "The port on which to serve prometheus metrics endpoint")
 	cmd.Flags().Duration("read-header-timeout", 10*time.Second, "The time allowed to read the headers of the requests in seconds")
+	cmd.Flags().Bool("duplex", false, "experimental: serve HTTP and GRPC on the same port instead of on two separate ports")
```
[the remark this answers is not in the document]
priyawadhwa: Hmm it would work, but as a user I don't find it very intuitive. As a user I'd expect that setting `--port` and `--grpc-port` to the same value would cause an error.  I'd still prefer to deprecate the grpc flags and have one `--port` flag to expose everything on, mostly because I think longterm it'll be cleaner, easier for users to understand, and because I don't see a use case for GRPC-only functionality.

cmd/app/serve.go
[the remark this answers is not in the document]
haydentherapper: I don't think it's a great look to deprecate functionality right after we've introduced it as stable. I know it's a tiny part of the service, but I think we shouldn't begin deprecation of features right after a 1.0 release until we start thinking about a 2.0 release, and right now, I don't see any features in the pipeline to need a 2.0 release.

cmd/app/serve.go
```diff
@@ -81,10 +96,11 @@ func newServeCmd() *cobra.Command {
 	cmd.Flags().String("tink-keyset-path", "", "Path to KMS-encrypted keyset for Tink-backed CA")
 	cmd.Flags().String("host", "0.0.0.0", "The host on which to serve requests for HTTP; --http-host is alias")
 	cmd.Flags().String("port", "8080", "The port on which to serve requests for HTTP; --http-port is alias")
-	cmd.Flags().String("grpc-host", "0.0.0.0", "The host on which to serve requests for GRPC")
-	cmd.Flags().String("grpc-port", "8081", "The port on which to serve requests for GRPC")
+	cmd.Flags().String("grpc-host", "0.0.0.0", "[DEPRECATED] The host on which to serve requests for GRPC")
+	cmd.Flags().String("grpc-port", "8081", "[DEPRECATED] The port on which to serve requests for GRPC")
```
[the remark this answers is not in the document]
haydentherapper: It might affect those who run their own infrastructure.

cmd/app/serve.go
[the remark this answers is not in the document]
haydentherapper: I see the addition of this change as fixing the error of setting the ports to be the same, rather than adding new functionality. I also would prefer to not use `duplex` as a flag name since that's exposing implementation details.

cmd/app/serve.go
```diff
@@ -313,3 +349,78 @@ func checkServeCmdConfigFile() error {
 	}
 	return nil
 }
+
+func StartDuplexServer(ctx context.Context, cfg *config.FulcioConfig, ctClient *ctclient.LogClient, baseca ca.CertificateAuthority, port, metricsPort int) error {
```
haydentherapper, line 353: Is `host` configurable?

priyawadhwa:
Discussed with @haydentherapper and made the following changes:
- Removed the `--duplex` flag in favor of using duplex if `--grpc-port` and `--port` are set to the same value.
- Removed the deprecation warnings for now, we can add those back in later once we feel duplex is running as expected.

@haydentherapper this is RFAL.

cmd/app/serve.go
priyawadhwa: viper.GetString("port")

cmd/app/serve.go
```diff
@@ -255,6 +271,22 @@ func runServeCmd(cmd *cobra.Command, args []string) {
 		}
 	}
 
+	if viper.GetString("port") == viper.GetString("grpc-port") {
+		p := viper.GetString("port")
```
bobcallaway, line 275: I think `viper.GetInt("port")` works and saves the `Atoi` call

cmd/app/serve.go
```diff
@@ -255,6 +271,22 @@ func runServeCmd(cmd *cobra.Command, args []string) {
...
+		port, err := strconv.Atoi(p)
+		if err != nil {
+			log.Logger.Fatal("%s is not a valid port", p)
+		}
+		mp := viper.GetString("metrics-port")
```
bobcallaway, line 280: same here on `GetInt`

cmd/app/serve.go
```diff
@@ -255,6 +271,22 @@ func runServeCmd(cmd *cobra.Command, args []string) {
...
+		metricsPort, err := strconv.Atoi(mp)
+		if err != nil {
+			log.Logger.Fatal("%s is not a valid port", mp)
+		}
+		if err := StartDuplexServer(ctx, cfg, ctClient, baseca, port, metricsPort); err != nil {
+			log.Logger.Fatal(err)
+		}
```
bobcallaway, line 287: I think we're missing a `return` statement after this, since we wouldn't want to go on and restart everything again

cmd/app/serve.go
```diff
@@ -314,3 +346,79 @@ func checkServeCmdConfigFile() error {
...
+	return nil
+}
+
+// The fulcio HTTP legacy client requires a 201 response status code to pass
+// However, even though the legacy client sets the 201 code, GRPC will automatically
+// change it to 200, causing the cert request to fail
+// GRPC recommends controlling HTTP status codes with this response modifier
+// https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/customizing_your_gateway/#controlling-http-response-status-codes
+// This is required to use fulcio with duplex.
+func HTTPResponseModifier(ctx context.Context, w http.ResponseWriter, p proto.Message) error {
```
bobcallaway, line 405: could we re-use https://github.com/sigstore/fulcio/blob/bc08c5b641033bcd9f604e202c84c8c9fde502dd/cmd/app/http.go#L97 instead of this implementation? The existing one deals with returning the SCT in a response header (which was part of v1 IIUC), and also strips out all GRPC response headers (to hide the internal implementation detail).

cmd/app/serve.go
```diff
@@ -255,6 +271,22 @@ func runServeCmd(cmd *cobra.Command, args []string) {
 		}
 	}
 
+	if viper.GetString("port") == viper.GetString("grpc-port") {
```
bobcallaway, line 274: Should this also test that `host` and `grpc-host` are equal before proceeding?

cmd/app/serve.go
```diff
@@ -255,6 +271,22 @@ func runServeCmd(cmd *cobra.Command, args []string) {
...
+		port, err := strconv.Atoi(p)
+		if err != nil {
+			log.Logger.Fatal("%s is not a valid port", p)
+		}
+		mp := viper.GetString("metrics-port")
+		metricsPort, err := strconv.Atoi(mp)
+		if err != nil {
+			log.Logger.Fatal("%s is not a valid port", mp)
+		}
+		if err := StartDuplexServer(ctx, cfg, ctClient, baseca, port, metricsPort); err != nil {
```
bobcallaway, line 285: pass `host` down to `StartDuplexServer`

cmd/app/serve.go
```diff
@@ -314,3 +346,79 @@ func checkServeCmdConfigFile() error {
...
+	reg.MustRegister(grpcMetrics, server.MetricLatency, server.RequestsCount)
+	grpc_prometheus.Register(d.Server)
+
+	// Register prometheus handle.
+	d.RegisterListenAndServeMetrics(metricsPort, false)
+
+	logger.Info("Starting duplex server...")
+	if err := d.ListenAndServe(ctx); err != nil {
+		return fmt.Errorf("duplex server: %w", err)
+	}
```
bobcallaway, line 395:
```suggestion
        lis, err := net.Listen("tcp", fmt.Sprintf("%s:%d", host, port))
	if err != nil {
		return fmt.Errorf("creating listener: %w", err)
	}
	if err := d.Serve(ctx, lis); err != nil {
		return fmt.Errorf("duplex server: %w", err)
	}
```

cmd/app/serve.go
priyawadhwa: Thanks! I hadn't noticed that before, it should work as well. I've updated it to use this one!

haydentherapper:
cc @bobcallaway for another look

bobcallaway (dismissed):

haydentherapper (commented):
LGTM, just needs to be rebased for go.mod

cmd/app/serve.go
bobcallaway: it was added in arg list in https://github.com/sigstore/fulcio/pull/931/files#diff-1627906dbce532aff8cc81f54790572e5ba04f3e568c88b6ed39a590642dafa4R345

haydentherapper: Oh whoops that was a comment I wrote awhile ago. This looks good

priyawadhwa:
Rebased! could i please get another lgtm?

bobcallaway (approved):

Issue #624, opened 2022-05-31
Add Github workflow run information to the signing certificate


tonistiigi:
**Description**

Currently the certificates created via Github token add the following GH info fields:

```
            X509v3 Subject Alternative Name: critical
                URI:https://github.com/<user>/<repo>/.github/workflows/<workflow>.yml@refs/heads/<branch>
            1.3.6.1.4.1.57264.1.3:
                <commit>
            1.3.6.1.4.1.57264.1.6:
                refs/heads/<branch>
            1.3.6.1.4.1.57264.1.5:
                <user>/<repo>
            1.3.6.1.4.1.57264.1.4:
                image
            1.3.6.1.4.1.57264.1.2:
                push
            1.3.6.1.4.1.57264.1.1:
                https://token.actions.githubusercontent.com
```

Would be good if this info would also include the workflow run information. Based on https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#understanding-the-oidc-token the token includes the run ID, run count and attempt count.

Having this info available would make a more direct connection to where the actual process that did the signing ran and look up the build logs if they are available. Could imagine a case where untrusted party has managed to trigger a workflow run on their terms and then tries to make it look like a legitimate release/branch build.

haydentherapper:
@asraa Any thoughts on this? Do you have the context on the set of information we initially chose to include in issued certs for GitHub? 

Some of this information seems more like build provenance rather than identity. It also depends where we draw the line for what represents a GitHub Actions identity. If we add this additional information, each run will get its own certificate with different identity information in the OIDs, so we're changing the certificate identity from a per-workflow identity to a per-run identity.

asraa:
Hey @tonistiigi - thanks for the issue!

>  Could imagine a case where untrusted party has managed to trigger a workflow run on their terms and then tries to make it look like a legitimate release/branch build.

Yes, I totally see the concern here.

To reiterate on the Hayden's point on identity: I think the information we include in the cert is meant to pinpoint the Workflow itself as an identity, not the Run (the repo, commit hash and ref pinning the workflow content). An anlogy would somewhat be that GitHub logins pinpoint on GitHub username/emails, and does not include extensions like "log in time" to map to the actual actor given that usernames can be overtaken.

Do you have a specific use-case in mind? I work on a project that creates build provenance that signs over the specific WorkflowRun information with Fulcio-issued certs, that may be a similar path for what you are trying to do.

[...]

PR #945, opened 2023-01-04
Standardizing Fulcio Certificate Extensions

feelepxyz:
#### Summary

Proposal for standardizing Fulcio's Certificate Extensions to align with the discussion on [standardizing OIDC token claims](https://github.com/sigstore/fulcio/issues/754) across CI/CD systems (today GitHub Actions, in future Circle, GitLab, Buildkite etc).

The aim here is to find new platform agnostic extensions (as the current ones are GitHub specific) that make sense across the different CI/CD providers that we'd like to see supported in Fulcio.

Fixes https://github.com/sigstore/fulcio/issues/754
Fixes https://github.com/sigstore/fulcio/issues/955

Signed-off-by: Philip Harrison <philip@mailharrison.com>


docs/oid-info.md
```diff
@@ -23,35 +23,49 @@ The `.1` is added to the root OID for sigstore for all OIDs set by Fulcio.
...
+An immutable reference to the specific version of the top-level build
+instructions. Should include the digest type followed by the digest, e.g.
+`sha1:abc123`.
+
+### 1.3.6.1.4.1.57264.3.6 | Build Siger URI
+
+Reference to specific build instructions that are responsible for signing. Can be the same as Build Config URI. For example a reusable workflow in GitHub Actions or a Circle CI Orbs.
+
+
+### 1.3.6.1.4.1.57264.3.7 | Build Siger Digest
```
fkorotkov, line 58: Could you please clarify this a bit similar to Build Siger URI? Is it up for a provider to decide what this is? E.g. SHA of an external GitHub Action or a hash of a bash script that did signing?

docs/oid-info.md
```diff
@@ -23,35 +23,49 @@ The `.1` is added to the root OID for sigstore for all OIDs set by Fulcio.
...
 
-### 1.3.6.1.4.1.57264.1.4 | GitHub Workflow Name
+### 1.3.6.1.4.1.57264.3.3 | Source Ref
 
-This contains the `workflow` claim from the GitHub OIDC Identity token that
-contains the name of the executed workflow. [(docs)][github-oidc-doc]
+The source ref that the build run was based upon. For example: `refs/head/main`.
 
-### 1.3.6.1.4.1.57264.1.5 | GitHub Workflow Repository
+### 1.3.6.1.4.1.57264.3.4 | Build Config URI
```
fkorotkov, line 43: I have a question about what to do with external modules. Let's take GHA as an example, one can use an external action and reference by a branch and not an presumably immutable SHA or tag:

```yaml
uses: publisher-username/repo-name@branch-name
```

In that case the `branch-name` will be evaluated at the time of build execution. Similar example is possible with CircleCI Orbs and with other CI systems.

I wonder if there is a need to have URIs to all the config files that affected the build. Or should this transitive nature be reflected in the following "Build Config Digest"?

docs/oid-info.md
```diff
@@ -63,6 +77,24 @@ defined by [RFC5280 4.2.1.6](https://datatracker.ietf.org/doc/html/rfc5280#secti
...
+| exp                | ??     | ??       | ??        | N/A                          | Only used to validate the JWT.                                                                                                                                                         |
+| nbf                | ??     | ??       | ??        | N/A                          | Only used to validate the JWT.                                                                                                                                                         |
+| iat                | ??     | ??       | ??        | N/A                          | Only used to validate the JWT.                                                                                                                                                         |
+| repository         | ??     | ??       | ??        | Source URI                   | Should include a fully qualified repository URL.                                                                                                                                       |
```
fkorotkov, line 89: I think right now GitHub's token has repository name here. I wonder of for backward compatibility it's worth to use `repository_url`?

docs/oid-info.md
```diff
@@ -23,35 +23,49 @@ The `.1` is added to the root OID for sigstore for all OIDs set by Fulcio.
...
+
+### 1.3.6.1.4.1.57264.3.7 | Build Siger Digest
+
+An immutable reference to the specific version of the build instructions that is responsible for signing. Should include the digest type followed by the digest, e.g. `sha1:abc123`.
+
+
+### 1.3.6.1.4.1.57264.3.8 | Build Trigger
+
+The event or action that triggered the build.
+### 1.3.6.1.4.1.57264.3.9 | Runner Environment
```
fkorotkov, line 66: `platform-hosted` environments are always ephemeral. In case of `self-hosted` environment they are most likely not ephemeral but can be. Should there be a some optional clarifications like version of environment for `platform-hosted` or host name for `self-hosted`?

fkorotkov (commented):
Philip, thank for summarizing everything! IMO the specification looks good and CI providers should be able to comply with ease. Left a few questions that was not clear to me.

docs/oid-info.md
```diff
@@ -23,35 +23,49 @@ The `.1` is added to the root OID for sigstore for all OIDs set by Fulcio.
...
+### 1.3.6.1.4.1.57264.3.5 | Build Config Digest
 
-This contains the `ref` claim from the GitHub OIDC Identity token that contains
-the git ref that the workflow run was based upon.
-[(docs)][github-oidc-doc]
+An immutable reference to the specific version of the top-level build
+instructions. Should include the digest type followed by the digest, e.g.
+`sha1:abc123`.
+
+### 1.3.6.1.4.1.57264.3.6 | Build Siger URI
```
phillmv, line 53:
```suggestion
### 1.3.6.1.4.1.57264.3.6 | Build Signer URI
```

docs/oid-info.md
phillmv, line 58:
```suggestion
### 1.3.6.1.4.1.57264.3.7 | Build Signer Digest
```

docs/oid-info.md
```diff
@@ -23,35 +23,49 @@ The `.1` is added to the root OID for sigstore for all OIDs set by Fulcio.
...
+
+
+### 1.3.6.1.4.1.57264.3.7 | Build Siger Digest
+
+An immutable reference to the specific version of the build instructions that is responsible for signing. Should include the digest type followed by the digest, e.g. `sha1:abc123`.
+
+
+### 1.3.6.1.4.1.57264.3.8 | Build Trigger
+
+The event or action that triggered the build.
```
phillmv, line 65:
```suggestion
The event or action that triggered the build.

```

haydentherapper:
Thanks so much for starting this proposal! I'll take a look at it in a little bit, cc'ing a few other folks who have worked on verification or CI that might have some thoughts.

@asraa @mattmoor @laurentsimon @sj26

docs/oid-info.md
```diff
@@ -23,35 +23,49 @@ The `.1` is added to the root OID for sigstore for all OIDs set by Fulcio.
 This contains the `issuer` claim from the OIDC Identity Token that was
 presented at the time the code signing certificate was requested to be created.
 This claim is the URI of the OIDC Identity Provider that digitally signed the
-identity token.
+identity token. For example: https://token.actions.githubusercontent.com.
 
-### 1.3.6.1.4.1.57264.1.2 | GitHub Workflow Trigger
+### 1.3.6.1.4.1.57264.3.1 | Source URI
```
nsmith5, line 28: The OID prefix `1.3.6.1.4.1.57264.1` is specifically registered for Fulcio so I think its important that these new definitions all use that prefix instead of changing to the `1.3.6.1.4.1.57264.3` prefix. Ref: http://oid-info.com/get/1.3.6.1.4.1.57264.1

nsmith5 (commented):
I'm a little hesitant on the value of trying to message all the various CI provider claims into to the same OID extensions. Have we considered simple giving each CI provider a different OID prefix (e.g  1.3.6.1.4.1.57264.1.12.xxx for all GitHub Actions,  1.3.6.1.4.1.57264.1.13.xxx for Gitlab etc) so that each provider can encode the claims that might be of value when writing policy without the need to make them fit the same abstraction?

docs/oid-info.md
nsmith5: Ah TIL! I agree that changing the prefix seems unnecessary and can lead to confusion though.

phillmv:
Hello! I work with @feelepxyz and with @steiza 🙂.

@nsmith5:

> I'm a little hesitant on the value of trying to massage all the various CI provider claims into to the same OID extensions.

On our team, we've been talking a lot about the verification and validation of different attestations. For us, these specific fields being proposed have emerged as requirements for building the provenance guarantees we're interested in providing.

By providing a wider surface area of standardized claims embedded in the certs, different CI providers can immediately integrate with, well, what GitHub is going to be shipping.

>Have we considered giving each CI provider a different OID prefix

I can see the utility of fulcio automatically embedding, well, _every_ field in the token in the certificate. That would make it easier to extend in the future 🤔.

docs/oid-info.md
feelepxyz: Yeah I can see how changing the prefix is confusing unless we change all of them. Can the suffix increment to double digits? e.g. `1.3.6.1.4.1.57264.1.12`, assuming we want to keep all existing extensions as they and add the new ones following `1.3.6.1.4.1.57264.1.7 | OtherName SAN`. So "Source URI" would end up as `1.3.6.1.4.1.57264.1.8` etc.

docs/oid-info.md
feelepxyz: Yes the intention here was to provide a immutable reference to the Build Signer URI. So in GitHub Actions case the URI would be the path with branch/tag ref of a reusable workflow. The digest would be the commit sha of the reusable workflow file. In GitHub's case the workflow ref usually contains a mutable branch/tag unless it's been referenced via a commit in the triggering workflow.

Hadn't thought about this directly referencing a bash script. I wonder if this is something that would be possible as all these values will be pulled from the CI/CD providers OIDC claims, so the CI provider would need to extract the location/digest of the script and place it in the token claims.

My intention with calling it the "build signer" was to make it clear that it's the job/container/thing where the CI/CD's OIDC token was requested. It's technically possible to pass this token back to another job where the signing happens though.

In GitHub actions case "Build Signer" and "Build Config" are identical if you're **not** using a reusable workflow. I imagine this would also be the case for Circle CI if using Orbs or not.

Another name idea for "Build Signer" could be "Build Identity" but potentially more confusing.

docs/oid-info.md
feelepxyz: Interesting point, I don't think we have any good way of capturing what exact commit external actions where pulled from in GitHub Actions but this would be nice to know for sure. 

My thinking here was that `Build Config` would only refer to the initiating or top-level config (or entry point in SLSA lingo) that was used when running the job in response to the "Build Trigger". The digest would be the commit sha of this top-level config. Given this you could see the exact top-level workflow that was used but not what exact transitive actions where used unless you did some kind of date based commit lookup in each action at the time the job ran?

docs/oid-info.md
feelepxyz: I don't think this claim exists atm for Actions. This probably means we would need to construct the source/repo URI in Fulcio based on the token issuer. Something like this for GHA:

```js
if (token.iss === 'https://token.actions.githubusercontent.com') {
  source_uri = `https://github.com/${token.repository}`
}
```

docs/oid-info.md
feelepxyz: Good point, we could construct this as a user-agent string that would support versions and extra metadata. 

I'm anticipating that each CI provider probably ends up having custom values that make most sense for them and would be hard to standardize across providers. In GitHub's case, it might be clearer to say `runner_env: github` instead of `runner_env: platform-hosted`.

feelepxyz:
> Have we considered giving each CI provider a different OID prefix (e.g 1.3.6.1.4.1.57264.1.12.xxx for all GitHub Actions, 1.3.6.1.4.1.57264.1.13.xxx for Gitlab etc) so that each provider can encode the claims that might be of value when writing policy without the need to make them fit the same abstraction?

@nsmith5 echoing what @phillmv said. 

Definitely an option that's worth considering. I don't yet know how feasible it will be to align across CI/CD providers on this approach but seems worth trying to see what shakes out.

From GitHub's perspective, we want these extensions in the signing certificate to support two uses-cases:
- Validate that key bits of the provenance statement match the signing certificate when not using a trusted builder
- Support policies that enforce that the provenance for a package matches the expected source for a given package on the npm registry at publish time
  - So a maintainer could say something like, only allow publishing my package from x repo, y branch and z workflow and enforce that the provenance matches this publish policy

These uses-cases could be achieved by adding a few GitHub specific extensions in the Fulcio cert (Build Config URI/Digest. Build Signer URI/Digest and Runner Environment).

The downside of having custom extensions for each provider would mainly fall on the validator having to know which fields each provider is using for what. Landing on some kind of standard here would simplify the work of downstream validators, but worth calling out that we would still risk leaky abstractions as each provider will encode things differently.

docs/oid-info.md
haydentherapper: yep, the suffix can increase to two digits. I think that would be fine to start from `.8`.

docs/deprecated-oid-info.md
```diff
@@ -0,0 +1,72 @@
+# Sigstore OID information (deprecated)
```
haydentherapper, line 1: I would prefer that we keep all OIDs in one document, but document which values are deprecated. These will still be needed for verification of all currently issued certificates.

feelepxyz: Updated 👍

feelepxyz:
Just came across this merged PR in Cosign from @developer-guy that would be affected by this change: https://github.com/sigstore/cosign/pull/1626/files

These commands would have to change if we standardized these cert extensions. We'll want to write both for some period while we update Cosign and give users time to upgrade.

haydentherapper:
@feelepxyz To avoid breaking changes, I would propose that Fulcio include duplicate extension values (same value in two OIDs) for some time before removing the older extensions. This follows the deprecation policy. This should be fine, these extensions are very small in size. We'll keep these around for 6 months or a year, then remove them in a 2.0 release of Fulcio.

This lets us update Cosign in parallel without breaking current users or other clients.

docs/oid-info.md
```diff
@@ -58,11 +60,86 @@ the git ref that the workflow run was based upon.
...
+
 ## 1.3.6.1.4.1.57264.2 | Policy OID for Sigstore Timestamp Authority
 
 Not used by Fulcio. This specifies the policy OID for the [timestamp authority](https://github.com/sigstore/timestamp-authority)
 that Sigstore operates.
 
-<!-- References -->
+## Mapping OIDC token claims to Fulcio OIDs
+
+| GitHub [(docs)][github-oidc-doc] | GitLab | CircleCI | Buildkite | Fulcio Certificate Extension | Why / Notes / Questions                                                                                                                                                                |
```
feelepxyz, line 127: For ref, current workload OIDC claims:
- Buildkite: https://github.com/sigstore/fulcio/pull/890#discussion_r1027648102
- Circle: https://circleci.com/docs/openid-connect-tokens/#format-of-the-openid-connect-id-token
- Gitlab: https://docs.gitlab.com/ee/ci/cloud_services/index.html#how-it-works
- GitHub (outdated): https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#understanding-the-oidc-token

docs/oid-info.md
```diff
@@ -58,11 +60,89 @@ the git ref that the workflow run was based upon.
...
+| ref                | ref     | ??       | build_branch | Source Ref                   | The source ref that the build run was based upon. For example: refs/head/main.                                                                                                         |
+| repository_id         | project_id     | ??       | ??        | Source Identifier     | Stable identifier for the owner of the source repository.                                                                                                                                      |
+| repository_owner         | namespace_path     | ??       | ??        | Source Owner URI     | Fully qualified URL for the owner of the source repository.                                                                                                                                     |
+| repository_owner_id         | namespace_id     | ??       | ??        | Source Owner Identifier     | Stable identifier for the owner of the source repository.                                                                                                                                   |
```
feelepxyz, line 139: Q: For providers like Buildkite and Circle who don't also host the source, should they forward the underlying source/source owner ids in their workload identity? How do we track the "build identity" that executed the build and distinguish this from the source identity? Is the build config URI enough?

woodruffw: Someone more familiar with those CI services can probably answer better than me, but my expectation is that `repository_id` would be the be the "project" or "pipeline" ID on Buildkite or Circle, while `repository_owner_id` would be the ID of the `User` or other owner model that configured the pipeline.

That's contingent on these services having some kind of underlying relational design that looks roughly like the following:

```
User(id, ...) from table of USERS
Project(id, user_id_roles) from table of PROJECTS
```

...but I think that's probably a reasonable assumption 🙂

nkreiger:
> These uses-cases could be achieved by adding a few GitHub specific extensions in the Fulcio cert (Build Config URI/Digest. Build Signer URI/Digest and Runner Environment).

@feelepxyz Is there value in a hybrid approach, or would this be too confusing. Why not have standard fields that the downstream validator should always expect, and map the other claims with the appropriate CI prefix. e.g. `github_runner_id`.

docs/oid-info.md
feelepxyz: Yeah would be good to get some clarity here. I was imaging `Source *` extensions would all point to the source repository. So if you use Circle to build with code on GitHub, Source URI would point to the github repo. Ref, digest would also point to the repo.

The reason we would want Source Owner Identifier and Source Identifier would be to spot transfers/resurrected repositories, in this example this would be the GitHub repo. Given this, I think we actually want Circle et al to provide these source/source owner ids in their OIDC claims. 

Build Config URI would presumably, in most cases, point to a build config file in the same repo. 

The Builder Signer URI, could in Circle's case be a [Circle Orb](https://circleci.com/orbs/) if one is used, otherwise this would point to the same above build config uri if one was not used.

Given this, there's actually no space in the current extensions for any build system identifiers like "project" or "pipeline" ID on Buildkite or Circle 🤔 

Question: do we need something specific here to track separate build system/build owner identifiers when this differs from the source system?

cc @haydentherapper what are your thoughts on above?

feelepxyz:
> @feelepxyz Is there value in a hybrid approach, or would this be too confusing.

@nkreiger Yeah first thought is that this would be confusing and unwieldy in the long term but I can also see some value here in case you want to write custom validators that check everything in a provenance statement against the Fulcio cert for example.

Do you have a specific use case in mind?

nkreiger:
> > @feelepxyz Is there value in a hybrid approach, or would this be too confusing.
> 
> @nkreiger Yeah first thought is that this would be confusing and unwieldy in the long term but I can also see some value here in case you want to write custom validators that check everything in a provenance statement against the Fulcio cert for example.
> 
> Do you have a specific use case in mind?

@feelepxyz That was pretty much my use case. No matter how many claims are added now, there will always be more that someone wants as things evolve. I see flexibility over complexity here maybe being stronger in the long-term. And simplifying it with required/guaranteed claims (the minimum to support a provider) that are well documented, might be able to keep the complexity lower.

That being said, I'm no expert, and have to do more due diligence to support my claims (pun intended). :)

docs/oid-info.md
haydentherapper: We should have a way to differentiate CI platforms, that should be a part of what’s validated by the verifier. Otherwise I could imagine a threat vector where an attacker creates a project on a different platform matching all of the project/repo details, but changing the source.

kommendorkapten:
@nkreiger @feelepxyz similar ideas has been discussed in this issue: https://github.com/sigstore/fulcio/issues/624

IMHO the set of extensions in the certificate should be fairly small, and have "stable" values that can be used to identify the signer, i.e not something like a runner id. Values that are expected to change between each run or just have a more "dynamic" nature is a much better fit for the provenance document, that is signed by the identity referred to by the certificate.

kommendorkapten (approved):

kommendorkapten:
@haydentherapper do you have any more concerns with this proposal? I think it's in a good shape and we can accept it, and so start the implementation of adding the new extensions to the certificate.

docs/oid-info.md
feelepxyz: > We should have a way to differentiate CI platforms, that should be a part of what’s validated by the verifier.

This makes sense, is this not already covered by the OIDC issuer extension?

docs/oid-info.md
```diff
@@ -58,11 +60,89 @@ the git ref that the workflow run was based upon.
...
+
+The source ref that the build run was based upon. For example: `refs/head/main`.
+
+### 1.3.6.1.4.1.57264.1.11 | Source Identifier
+
+The stable identifier for the source repository the workflow was based upon.
+
+### 1.3.6.1.4.1.57264.1.12 | Source Owner URI
+
+Fully qualified URL for the owner of the source repository that the build was based
```
laurentsimon, line 85: nit: Adding an example for each of these fields would be useful.

haydentherapper:
@kommendorkapten i have not yet had a chance to review it and plan to do it either next week or the following. I’d also like some folks from SLSA to chime in, since they’ve had some experience with verifying with GH claims.

trevrosen:
This is a hard blocker for quickly onboarding a collection of CI providers who are keep to participate in the (rapidly approaching) beta programs surrounding [npm integration](https://github.com/npm/rfcs/blob/e000b367d9e595bc694893c3d845df269f9b875f/accepted/0049-link-packages-to-source-and-build.md), and so I feel we need to bring this to conclusion in the next few days.

Unless I'm missing something, we have the contours of the solution basically set? From my POV the main outstanding item at this point is documenting examples per suggestion from @laurentsimon. Does anyone feel that there are fields _missing_ which need to be accounted for?

( cc @bobcallaway @lukehinds @SantiagoTorres @priyawadhwa )

docs/oid-info.md
```diff
@@ -58,11 +60,89 @@ the git ref that the workflow run was based upon.
...
+### 1.3.6.1.4.1.57264.1.8 | Source URI
+
+Should include a fully qualified source repository URL that the build was based
+on.
+
+### 1.3.6.1.4.1.57264.1.9 | Source Digest
+
+An immutable reference to a specific version of the source code that the build
+was based upon. Should include the digest type followed by the digest, e.g.
+`sha1:abc123`.
```
phillmv, line 73:
```suggestion
was based upon. Should include the digest type followed by the digest, e.g.
`sha1:abc123` for a git commit.
```

docs/oid-info.md
```diff
@@ -58,11 +60,89 @@ the git ref that the workflow run was based upon.
 This specifies the username identity in the OtherName Subject Alternative Name, as
 defined by [RFC5280 4.2.1.6](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6).
 
+
+### 1.3.6.1.4.1.57264.1.8 | Source URI
+
+Should include a fully qualified source repository URL that the build was based
+on.
```
phillmv, line 67:
```suggestion
on, e.g. `https://github.com/sigstore/fulcio`.
```

docs/oid-info.md
```diff
@@ -58,11 +60,89 @@ the git ref that the workflow run was based upon.
 This specifies the username identity in the OtherName Subject Alternative Name, as
 defined by [RFC5280 4.2.1.6](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6).
 
+
+### 1.3.6.1.4.1.57264.1.8 | Source URI
```
phillmv, line 64: In order to disambiguate between "Source URI" and "Source Repository URI", how about:

```suggestion
### 1.3.6.1.4.1.57264.1.8 | Source Repository URI
```

docs/oid-info.md
```diff
@@ -58,11 +60,89 @@ the git ref that the workflow run was based upon.
...
+
+Should include a fully qualified source repository URL that the build was based
+on.
+
+### 1.3.6.1.4.1.57264.1.9 | Source Digest
```
phillmv, line 69:
```suggestion
### 1.3.6.1.4.1.57264.1.9 | Source Repository Digest
```

docs/oid-info.md
```diff
@@ -58,11 +60,89 @@ the git ref that the workflow run was based upon.
...
+
+An immutable reference to a specific version of the source code that the build
+was based upon. Should include the digest type followed by the digest, e.g.
+`sha1:abc123`.
+
+### 1.3.6.1.4.1.57264.1.10 | Source Ref
```
phillmv, line 75:
```suggestion
### 1.3.6.1.4.1.57264.1.10 | Source Repository Ref
```

docs/oid-info.md
```diff
@@ -58,11 +60,89 @@ the git ref that the workflow run was based upon.
...
+
+The source ref that the build run was based upon. For example: `refs/head/main`.
+
+### 1.3.6.1.4.1.57264.1.11 | Source Identifier
```
phillmv, line 79:
```suggestion
### 1.3.6.1.4.1.57264.1.11 | Source Repository Identifier
```

docs/oid-info.md
```diff
@@ -58,11 +60,89 @@ the git ref that the workflow run was based upon.
...
+
+The stable identifier for the source repository the workflow was based upon.
```
phillmv, line 81:
```suggestion
The immutable identifier for the source repository the workflow was based upon, for e.g. `1234` if using a primary key.
```

docs/oid-info.md
```diff
@@ -58,11 +60,89 @@ the git ref that the workflow run was based upon.
...
+
+### 1.3.6.1.4.1.57264.1.12 | Source Owner URI
+
+Fully qualified URL for the owner of the source repository that the build was based
+on.
```
phillmv, line 86:
```suggestion
on, e.g. `https://github.com/sigstore`
```

docs/oid-info.md
```diff
@@ -58,11 +60,89 @@ the git ref that the workflow run was based upon.
...
+
+### 1.3.6.1.4.1.57264.1.13 | Source Owner Identifier
+
+The stable identifier for the owner of the source repository that the workflow was based upon.
```
phillmv, line 90:
```suggestion
The immutable identifier for the owner of the source repository that the workflow was based upon, e.g. `5678` if using a primary key.
```

docs/oid-info.md
```diff
@@ -58,11 +60,89 @@ the git ref that the workflow run was based upon.
...
+
+An immutable reference to the specific version of the top-level build
+instructions. Should include the digest type followed by the digest, e.g.
+`sha1:abc123`.
+
+### 1.3.6.1.4.1.57264.1.16 | Build Signer URI
+
+Reference to specific build instructions that are responsible for signing. Can be the same as Build Config URI. For example a reusable workflow in GitHub Actions or a Circle CI Orbs.
+
+Note this value would also be set in the Subject Alternative Name (SAN).
```
phillmv, line 106: For simplicity sake, let's just cut this line out. Even if it is the SAN, this value should still be filled out.

```suggestion
```

phillmv (commented):
Added some editing suggestions & proposed some examples.

docs/oid-info.md
```diff
@@ -58,11 +60,89 @@ the git ref that the workflow run was based upon.
...
+An immutable reference to the specific version of the build instructions that is responsible for signing. Should include the digest type followed by the digest, e.g. `sha1:abc123`.
+
+### 1.3.6.1.4.1.57264.1.18 | Build Trigger
+
+The event or action that triggered the build.
+
+### 1.3.6.1.4.1.57264.1.19 | Runner Environment
+
+For platforms to specify whether the build took place in platform-hosted cloud infrastructure or customer-hosted infrastructure. For example: `platform-hosted` and `self-hosted`.
+
```
phillmv, line 119:
```suggestion

### 1.3.6.1.4.1.57264.1.20 | Runner Execution Identifier

An immutable value that can uniquely identify the build' sexecution. For example: `4043059413`

```

docs/oid-info.md
```diff
@@ -58,11 +60,89 @@ the git ref that the workflow run was based upon.
...
+| job_workflow_ref   | ??     | ??       | ??        | Build Signer URI             | Reference to specific build instructions that are responsible for signing. Can be the same as Build Config URI. For example a reusable workflow in GitHub Actions or a Circle CI Orbs. |
+| job_workflow_sha   | ??     | ??       | ??        | Build Signer Digest          | An immutable reference to the specific version of the build instructions that is responsible for signing. Should include the digest type followed by the digest, e.g. `sha1:abc123`.   |
+| event_name         | pipeline_source | ??       | ??        | Build Trigger                | The event or action that triggered the build.                                                                                                                                          |
+| runner_environment | ??     | ??       | ??        | Runner Environment           | For platforms to specify whether the build took place in platform-hosted cloud infrastructure or customer-hosted infrastructure. For example: `platform-hosted` and `self-hosted`.     |
```
phillmv, line 145:
```suggestion
| runner_environment | ??     | ??       | ??        | Runner Environment           | For platforms to specify whether the build took place in platform-hosted cloud infrastructure or customer-hosted infrastructure. For example: `platform-hosted` and `self-hosted`.     |
| run_id | ?? | ?? | ?? | Runner Execution Identifier | An immutable value that can uniquely identify the build's execution. For example: `4043059413`
```

phillmv (commented):
This is a late-stage addition, so my apologies, but after chatting with @kommendorkapten, I decided to propose adding the `run_id` to the cert as an added "transparency" measure.

We want to push everyone towards a future of re-usable workflows, but in the meantime being able to provide a non-falsifiable audit trail back to the specific build execution that requested the cert feels like an interesting/important property – especially in forensic contexts.

Cheers,

docs/oid-info.md
```diff
@@ -58,11 +60,89 @@ the git ref that the workflow run was based upon.
...
+
+For platforms to specify whether the build took place in platform-hosted cloud infrastructure or customer-hosted infrastructure. For example: `platform-hosted` and `self-hosted`.
+
 ## 1.3.6.1.4.1.57264.2 | Policy OID for Sigstore Timestamp Authority
 
 Not used by Fulcio. This specifies the policy OID for the [timestamp authority](https://github.com/sigstore/timestamp-authority)
 that Sigstore operates.
 
-<!-- References -->
+## Mapping OIDC token claims to Fulcio OIDs
```
asraa, line 125: A proposed new claim: a version claim. I like the idea of sticking direct OIDC claims into the extension fields in the table. What I'm worried about is that the OID extensions are not versioned, and there is no way of allowing an OIDC provider to "update" the mapping (because we don't want to keep adding fields from this standardization). Having a Fulcio-integration "version" will help allow providers to update the table claims.

For e.g. if GH decides to add a sha prefix to the source digest, assuming that Fulcio is not sanitizing or modifying claims into this table, then that would be breaking to policy verifiers.

woodruffw: Agreed, and IMO this is another good motivation for a single "composite" extension that wraps all of these claims as suggested in #981 (although that will need to happen with a breaking version, not here).

docs/oid-info.md
haydentherapper: Yes this is! OIDC issuer is sufficient

docs/oid-info.md
asraa: > but in the meantime being able to provide a non-falsifiable audit trail back to the specific build execution that requested the cert feels like an interesting/important property – especially in forensic contexts.

Would this one that is deprecated eventually? 

I understand the benefit of one, but this seems like a temporary solution to the broader problem of actually creating a **non-falsifiable builder** to get that audit trail

docs/oid-info.md
haydentherapper: Thought a bit more on this - Does a version that Fulcio maintains solve the problem of breaking policy verifiers? And furthermore, is that Fulcio's problem?

GH making a change to the format of the source digest would affect a verifier if they were directly consuming the OIDC token. This seems more like a problem for the OIDC providers, that they should version their tokens and we can then also include that version value in an extension. Fulcio isn't opinionated on each claim structure (at least that's what we decided after chatting - If it was opinionated, then we would need a version that fulcio maintains)

Adding a claim in a new extension should not be a breaking change, and clients should search for it and gracefully handle it not being present (verification policies can require the extension if it should be critical).

asraa (commented):
One more comment, can we delinieate the MUSTs and SHOULDs?

I have been thinking about this, and many of the fields that are GH-only are things that are nice to have, but can be trusted once minimal MUSTs are met and trusted.

To me the MUST will include at least include job_workflow_ref, immutable digest to it, runner, issuer. I think it also includes the owner info, since that may change trust on the builder. These are what I think constitutes a "builder identity"

The tier 1 SHOULDs are the source info (repo / id / ref / digest), as I believe it's pretty important to be able to get the target of the build. It just so happens this is the minimal policy we require for provenance verification when we think of SLSA.

The rest feel like tier 2 SHOULDs (workflow_ref/sha, event_name, run_id)

It's convenient to have the rest of fields, but it's really just provenance inside x509. But please let me know.

docs/oid-info.md
haydentherapper: Chatting with Asra, throwing around some ideas (sorry if this is not clear, just trying to get this down to remember this):

* Each extension becomes a list of structs of {version, value}. If we have to update a extension value, we add a new struct to the list, with an updated version, while maintaining the previous version-value. 
* We must always operate in append-only, else we break verification for clients
* If we were to do something like #981, while being append-only, we'd have to duplicate the entire composite every time we'd do an update for any extension value for any CI provider
* The first bullet point here minimizes the duplications. 
* Another idea is a version-per-issuer, but this needs more thought

docs/oid-info.md
kommendorkapten: I think it actually makes sense. I would think of this kind of similar to the serial number for a cert. It adds a trust relation to it's origin (i.e its not at all intended for any policy decision).

Today, if we would like to understand if certificate from Fulcio is fraudulent or not, we have very limited information to go on. We have the the time it was created, but that's about it, the rest of the correlation would be a bit fuzzy. If we have the e.g. the run id, we know it's origin and can easier compare logs.

Having transparency is of course what enables this, similar to how it's done with WebPKI. What I think can be a problem here, is that the order of certificates created is on a different scale for signing certificates compared to domain certificates, so a "weak mapping" is harder to implement, and so including the run id would create a stronger link that's easier to automate on.

asraa: > I would think of this kind of similar to the serial number for a cert

Yeah, I was thinking about this, and wondering if it's something like a "session id" for the dex/email case.

My one question is: does it uniquely identify the identity/trust of the run? From what I understand, a single run_id can reference maybe many reusable workflows, which have different "identities". (Or is it the combo of the run_id and the job_workflow_ref/sha that can give you a unique reference and auditable issuance to the signer?)

What do you think then is the criticality of this extension? Honestly now maybe I can't decide if it's a MUST or very optional. MUST because it's an audit number to the OIDC token (does any other field suffice?), or optional because it's not related really to a useful part of the "identity" of the signer. Maybe I don't understand this: are you trying to audit the OIDC token issuance, or the invocation it was used in? The latter is definitely not "identity" policy related, so would be optional.

kommendorkapten: Yes, you are correct that we need to figure out how to make it unique, like a combination of`run_id/number/attempt`. This information together with e.g. the workflow_ref or sha as you propose. Not sure if the extension need to be cater for this, or if the "backing system" should be responsible for putting them altogether, I think that may be more useful as it's a bit more flexible/future-proof?

I would say it's RECOMMENDED? As from a general signer/identity perspective, it doesn't really add much?

kommendorkapten: A potential use-case I can see is for a CI provider to monitor the CT log and correlate each certificate with a workflow run to scan for forged certificates.

asraa: > I would say it's RECOMMENDED? As from a general signer/identity perspective, it doesn't really add much?

I'd be happy with that! It actually works for almost all extensions as well "here's the description of the extension and here's the "usage hint"" (which can help distinguish between critical for identity verification and informational and for what). WDYT @haydentherapper? 

Ideally the critical for identity verification line up with the required extensions Fulcio provides.

Other things, we should document and be clear what it's purpose is -- so that if some client abuses usage, it's up to them really but at least the intent for the extension is clear.

feelepxyz: I like this addition! As mentioned, this needs to include the run id and attempt to make it unique in GHA. Thoughts on following the lead of SLSA 1.0 draft and constructing a URI for this? https://github.com/slsa-framework/slsa/blob/main/docs/github-actions-workflow/v0.1/index.md#metadata

Something like `Run Invocation URI: github.server_url + "/actions/runs/" + github.run_id + "/attempts/" + github.run_attempt`. This way the value is immediately parseable and useful without having to know some custom format for GHA, e.g. `github.run_id + '-' + github.run_attempt`.

feelepxyz: `Build Invocation URI`?

kommendorkapten: While I don't think strict alignment with SLSA is relevant for Fulcio, I do think it's valuable to rely on prior art, so I like the idea or reusing this format.

docs/oid-info.md
feelepxyz: Good callout! @phillmv helpfully provided a bunch of examples. I'll add a terms section.

docs/oid-info.md
feelepxyz: Added this as `Run Invocation URI` in the current proposal.

docs/oid-info.md
moensch: > but my expectation is that `repository_id` would be the be the "project" or "pipeline" ID on Buildkite or Circle, while `repository_owner_id` would be the ID of the `User` or other owner model that configured the pipeline.

I am familiar with Buildkite. Reading this spec, I would assume any OID mentioning "Source" or "Repository" would be some pointer back to, or meta data about the GitHub (in my case) repository and commit/ref that triggered the build.

And I'd expect `Run Invocation URI` to be something like `https://buildkite.com/mycompany/mypipeline/builds/118`

And my issuer (Buildkite) would be trusted to set the appropriate fields related to the source repository.

docs/oid-info.md
```diff
@@ -8,46 +8,83 @@ code signing certificates issued by Fulcio instances. This document aims to
...
+- SHOULD include claim to support: `Source Repository Identifier`
+- SHOULD include claim to support: `Source Repository Owner URI`
+- SHOULD include claim to support: `Source Repository Owner Identifier`
+
+Nice-to-haves:
+
+- SHOULD include claim to support: `Build Config URI`
+- SHOULD include claim to support: `Build Config Digest`
+- SHOULD include claim to support: `Build Trigger`
+- SHOULD include claim to support: `Run Invocation URI`
```
feelepxyz, line 37:
```suggestion
- MAY include claim to support: `Build Config URI`
- MAY include claim to support: `Build Config Digest`
- MAY include claim to support: `Build Trigger`
- MAY include claim to support: `Run Invocation URI`
```

docs/oid-info.md
```diff
@@ -58,11 +95,93 @@ the git ref that the workflow run was based upon.
...
+### 1.3.6.1.4.1.57264.1.8 | Source Repository URI
+
+Should include a fully qualified source repository URL that the build was based
+on. For example: `https://github.com/sigstore/fulcio`.
+
+### 1.3.6.1.4.1.57264.1.9 | Source Repository Digest
+
+An immutable reference to a specific version of the source code that the build
+was based upon. Should include the digest type followed by the digest. For example:
+`sha1:abc123` for a git commit.
```
woodruffw, line 108: Not to make things more difficult here (sorry!), but is there any possibility we could make this more normative?

For example, we could enforce the following:

1. Only certain digest types are valid (e.g. `sha1`, `sha256`)
2. The digest itself must be a hexadecimal string, not another representation (like base64)
3. The digest itself must be a full digest, not a truncated one

(This is a good argument IMO for using ASN.1 structures, since we'd get a lot of these restrictions for "free" with ASN.1 definitions. But that's further out, so just listing these as plain language constraints would be good IMO.)

haydentherapper: Our thought was that Fulcio should not be opinionated on claim values and it's up to the verifier to understand what claims will look like (i.e. they must consult the documentation of the provider, not Fulcio).

woodruffw: That makes sense! I'm a big fan of ratcheting down formats/flexibility wherever it can be ratcheted, but I can understand if this is a place where provider variability/diversity would make doing so too complicated or onerous.

Given that, I think many of these claims would benefit from `MAY` language that emphasize that providers may choose to fill these values as they please, e.g.:

```
Providers MAY choose to emit this extension's value in other formats, and consumers MUST NOT assume the format above. 
```

haydentherapper: +1, I think we should be more clear in describing examples as optional suggestions.

We wanted to make it as easy as possible to onboard new providers, which was another justification for fulcio just forwarding provided claims. Otherwise we're requiring CI platforms to conform to fulcio's requirements, which would likely not happen, or fulcio would mutate the claim before certifying it, which increases trust in fulcio.

woodruffw: > Otherwise we're requiring CI platforms to conform to fulcio's requirements, which would likely not happen, or fulcio would mutate the claim before certifying it, which increases trust in fulcio.

Oh, this is a great point -- I hadn't considered how this places additional trust in Fulcio. Thanks!

docs/oid-info.md
asraa: Thinking more on this: I think we should not complicate client logic with versioning logic, and push back on integrators who will be making version breaking changes to their  OIDC tokens. In a future world, I'd like OIDC tokens maybe to be versioned and for that to be represented in a claim, but I think Fulcio self-managing a versioning scheme would be pretty tough as an intermediate versioning layer between the OIDC provider and the client.

As in: let's push back on integrators who won't satisfy stable claims for the ones they are committing to tin the mapping.

docs/oid-info.md
```diff
@@ -8,46 +8,83 @@ code signing certificates issued by Fulcio instances. This document aims to
...
+
+- MAY include claim to support: `Build Config URI`
+- MAY include claim to support: `Build Config Digest`
+- MAY include claim to support: `Build Trigger`
+- MAY include claim to support: `Run Invocation URI`
+
+## Terminology
+
+- `Identifier`: Immutable resource identifier, e.g. uuid/primary key ID
+- `URI`: SHOULD be a fully qualified URL when available. CAN be a mutable resource identifier, e.g. `https://github.com/sigstore/fulcio`
```
haydentherapper, line 42: nit, can we use MAY rather than CAN (from https://www.rfc-editor.org/rfc/rfc2119)

feelepxyz:
@haydentherapper @asraa @trevrosen made the final updates. Should be good to go now. Want to give it a final review?

docs/oid-info.md
```diff
@@ -58,11 +95,90 @@ the git ref that the workflow run was based upon.
 This specifies the username identity in the OtherName Subject Alternative Name, as
 defined by [RFC5280 4.2.1.6](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6).
 
+
+### 1.3.6.1.4.1.57264.1.8 | Source Repository URI
+
+Should include a fully qualified source repository URL that the build was based
+on. For example: `example/repository`.
```
kommendorkapten, line 102: The comment mentions a fully qualified source repository URL, which according to the terminology is: "Complete URL with protocol", this example does not match that.

docs/oid-info.md
```diff
@@ -58,11 +95,90 @@ the git ref that the workflow run was based upon.
...
+The source ref that the build run was based upon. For example: `refs/head/main`.
+
+### 1.3.6.1.4.1.57264.1.11 | Source Repository Identifier
+
+The immutable identifier for the source repository the workflow was based upon. For example: `1234` if using a primary key.
+
+### 1.3.6.1.4.1.57264.1.12 | Source Repository Owner URI
+
+Fully qualified URL for the owner of the source repository that the build was based
+on, e.g. `example`
```
kommendorkapten, line 120: Same here, the example is not fully qualified.

docs/oid-info.md
```diff
@@ -58,11 +95,90 @@ the git ref that the workflow run was based upon.
...
+A reference to the initiating build instructions. For example: `example/repository/.github/workflows/publish.yml@refs/head/main`.
+
+### 1.3.6.1.4.1.57264.1.15 | Build Config Digest
+
+An immutable reference to the specific version of the top-level build
+instructions. For example: `abc123` git commit SHA.
+
+### 1.3.6.1.4.1.57264.1.16 | Build Signer URI
+
+Reference to specific build instructions that are responsible for signing. Can be the same as Build Config URI. Could be a reusable workflow ref in GitHub Actions or a Circle CI Orb name. For example: `slsa-framework/slsa-github-generator/.github/workflows/generator_container_slsa3.yml@v1.4.0`.
```
kommendorkapten, line 137: Same here on fully qualified.

docs/oid-info.md
```diff
@@ -58,11 +95,90 @@ the git ref that the workflow run was based upon.
...
+An immutable reference to the specific version of the top-level build
+instructions. For example: `abc123` git commit SHA.
+
+### 1.3.6.1.4.1.57264.1.16 | Build Signer URI
+
+Reference to specific build instructions that are responsible for signing. Can be the same as Build Config URI. Could be a reusable workflow ref in GitHub Actions or a Circle CI Orb name. For example: `https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_container_slsa3.yml@v1.4.0`.
+
+### 1.3.6.1.4.1.57264.1.17 | Build Signer Digest
+
+An immutable reference to the specific version of the build instructions that is responsible for signing. For example: `abc123` git commit SHA.
```
feelepxyz, line 141: @haydentherapper @asraa @kommendorkapten It just occurred to me that we won't just be dealing with Git SHA1's fairly soon. Circle has the concept of [Orb's](https://circleci.com/orbs/) that could be used to create a trusted builder and AFAICT these are OCI bundles so probs use SHA256.

Do you see any issue with not including the algo here if it's mostly SHA1 for all Git SCM's but SHA256 for Circle Orb's? It seems ok to push this on clients to figure out or by prior knowledge based on issuer 🤷 

From our chats I'm not sold on doing anything clever in Fulcio as this seems like it might just cause more problems down the road but keen to get your thoughts.

kommendorkapten: @feelepxyz I think keeping the value in Fulcio as-is from the integrator is preferred. I wouldn't think it's a big problem to omit the algorithm, as you said the clients should have prior knowledge on what to expect. This I think is also consistent with the other extensions in the certificate which values are specific per integration.

haydentherapper: I agree with keeping the value as-is, though I'll play devils advocate here because I do see your concern. Are there any other extensions where verifiers are going to need to be aware of which CI provider the certificate came from? If not, then this is additional work for verifiers for just a single claim.
On the flip side, how will fulcio determine the algorithm? If we're just going to check digest length for example, that's something a client could do.

docs/oid-info.md
```diff
@@ -2,52 +2,89 @@
...
+Requirements:
+
+- MUST include `iss` claim for `Issuer` extension.
+- MUST include claim to support: `Build Signer URI` that identifies the specific build instructions that are responsible for signing.
+- MUST include claim to support: `Build Signer Digest` which is an immutable reference to a specific version of the build instructions that are responsible for signing.
+- MUST include claim to support: `Runner Environment` that differentiates between builds that took place in platform-hosted cloud infrastructure or customer-hosted infrastructure.
+
+Recommended:
+
+- SHOULD include `iss` that uniquely identifies ID tokens originating from the CI/CD system, e.g. not shared with OIDC OAuth 2.0 tokens for email/username logins.
```
haydentherapper, line 24: What is the reason for this requirement?

docs/oid-info.md
```diff
@@ -2,52 +2,89 @@
...
+- MAY include claim to support: `Build Config Digest`
+- MAY include claim to support: `Build Trigger`
+- MAY include claim to support: `Run Invocation URI`
+
+## Terminology
+
+- `Identifier`: Immutable resource identifier, e.g. uuid/primary key ID
+- `URI`: SHOULD be a fully qualified URL when available. MAY be a mutable resource identifier, e.g. `https://scm.com/example/repository`
+  - Fully qualified URL: Complete URL with protocol.
+- `Digest`: Immutable source repository reference, e.g. git commit SHA
```
haydentherapper, line 44: nit: I assume a digest isn't always a source repository ref, like for "build config digest", so should this instead say "Output of a cryptographic hash function"

docs/oid-info.md
```diff
@@ -2,52 +2,89 @@
...
+
 ## Directory
 
 Note that all values begin from the root OID 1.3.6.1.4.1.57264 [registered by Sigstore][oid-link].
 
 When adding additional OIDs under the root, please update the above link with the child OID.
 
+GitHub Workflow specific OID extensions have been deprecated in favour of provider generic extensions starting from 1.3.6.1.4.1.57264.1.8.
```
haydentherapper, line 52: nit: "favour" -> "favor" (I like UK English spellings, but just being consistent)

docs/oid-info.md
```diff
@@ -58,11 +95,90 @@ the git ref that the workflow run was based upon.
 This specifies the username identity in the OtherName Subject Alternative Name, as
 defined by [RFC5280 4.2.1.6](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6).
 
+
+### 1.3.6.1.4.1.57264.1.8 | Source Repository URI
+
+Should include a fully qualified source repository URL that the build was based
```
haydentherapper, line 101: This one says "should" but we don't include should in other extensions. Is this referring to that the URL SHOULD be fully qualified, but may not be? Suggestion to rephrase: "Source repository URL that the build was based on. SHOULD be fully qualified. For example: blah", and then do the same with other URL extensions.

If being fully qualified is a MUST, we should note this since we'll need to enforce that.

docs/oid-info.md
```diff
@@ -58,11 +95,90 @@ the git ref that the workflow run was based upon.
...
+| repository_id         | project_id     | ??       | ??        | Source Repository Identifier     | Stable identifier for the owner of the source repository.                                                                                                                                      |
+| repository_owner         | namespace_path     | ??       | ??        | Source Repository Owner URI     | Fully qualified URL for the owner of the source repository.                                                                                                                                     |
+| repository_owner_id         | namespace_id     | ??       | ??        | Source Repository Owner Identifier     | Stable identifier for the owner of the source repository.                                                                                                                                   |
+| workflow_ref       | ??     | ??       | ??        | Build Config URI             | A reference to the initiating build instructions.                                                                                                                                      |
```
haydentherapper, line 175: For this and job_workflow_ref, do we need to append `server_url` to make them into URIs?

docs/oid-info.md
```diff
@@ -58,11 +95,90 @@ the git ref that the workflow run was based upon.
...
+### 1.3.6.1.4.1.57264.1.12 | Source Repository Owner URI
+
+Fully qualified URL for the owner of the source repository that the build was based
+on, e.g. `https://example.com/owner`
+
+### 1.3.6.1.4.1.57264.1.13 | Source Repository Owner Identifier
+
+The immutable identifier for the owner of the source repository that the workflow was based upon. For example: `5678` if using a primary key.
+
+### 1.3.6.1.4.1.57264.1.14 | Build Config URI
```
haydentherapper, line 126: To confirm, should this include the server URL? For GH, I see that workflow_ref and job_worflow_ref don't. If so, we should add SHOULD or MUST.

docs/oid-info.md
```diff
@@ -58,11 +95,90 @@ the git ref that the workflow run was based upon.
...
+
+A reference to the initiating build instructions. For example: `https://example.com/owner/repository/build-config.yml`.
+
+### 1.3.6.1.4.1.57264.1.15 | Build Config Digest
+
+An immutable reference to the specific version of the top-level build
+instructions. For example: `abc123` git commit SHA.
+
+### 1.3.6.1.4.1.57264.1.16 | Build Signer URI
```
haydentherapper, line 135: Was there a previous iteration of this doc that called out that this would be the Subject Alternative Name? Do we want to include that?

docs/oid-info.md
haydentherapper: I don't see any other claims where it's expected to know the CI provider.

Slightly tangential, is hash_alg:digest standardized anywhere?

docs/oid-info.md
```diff
@@ -58,11 +95,90 @@ the git ref that the workflow run was based upon.
...
+
+An immutable reference to the specific version of the build instructions that is responsible for signing. For example: `abc123` git commit SHA.
+
+### 1.3.6.1.4.1.57264.1.18 | Build Trigger
+
+The event or action that triggered the build. For example: `push`.
+
+### 1.3.6.1.4.1.57264.1.19 | Run Invocation URI
+
+An immutable identifier that can uniquely identify the build execution. For GitHub Actions this needs to include both the run id and attempt. For example: `https://github.com/example/repository/actions/runs/1536140711/attempts/1`.
```
haydentherapper, line 149: Should we omit "For GHA this needs to include ...", since that is documented below already?

docs/oid-info.md
haydentherapper: Leaving this open in case anyone has any other thoughts, but I think we're all in agreement that we won't attempt to do any versioning.

docs/oid-info.md
kommendorkapten: > If we're just going to check digest length for example, that's something a client could do.

Totally agree. A better option IMHO is to push for the CI vendors to switch to alg:digest. 

Also thinking how a verifier should consume the digest during policy execution. By definition the digest does not have any semantic structure, and so what a client can do is to keep an allowlist of known "good" digests for the builders. That would mean that the digest is an opaque string used to compare a against a list of approved values. In that case, the inclusion of algorithm would not make a difference. 

If we compare this with the subject digest in an in-toto attestation, the hash is important, as we are mapping artifacts (where the client needs to first compute the message digest) to attestations, which will likely be of high cardinality. For builders I would expect low cardinality and a highly specialized scenrio with much tighter policies, where the client typically don't compute a digest.

> Slightly tangential, is hash_alg:digest standardized anywhere?

I'm only aware of the OCI image spec: https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md#digests

docs/oid-info.md
feelepxyz: Yep 👍

docs/oid-info.md
feelepxyz: > Is this referring to that the URL SHOULD be fully qualified, but may not be?

👍 good spot, updated all the examples to make this clearer.

docs/oid-info.md
feelepxyz: Updated these to clarify the url SHOULD be fully qualified.

Just to nail down the GHA details, we're now prefixing these:
- `Run Invocation URI`: `server_url + repository + "/actions/runs/" + run_id + "/attempts/" + run_attempt`
- `Build Signer URI`:  `server_url + job_workflow_ref`

But not these:
- `Source Repository URI`: `repository` (owner/repo)
- `Source Repository Owner URI`: `repository_owner` (owner)
- `Build Config URI`: `workflow_ref`

.. should we be consistent across the board and prefix those with `server_url`? 😅

docs/oid-info.md
feelepxyz: > On the flip side, how will fulcio determine the algorithm? If we're just going to check digest length for example, that's something a client could do.

Yeah was thinking either check length or just hardcode expected algo for a given issuer and claim but this might break in future. But doesn't feel like a major value add?

> I don't see any other claims where it's expected to know the CI provider.

You kinda need to know what to expect for each CI provider for all of them to be able to verify things. For `Runner Environment` this might be set to `github-hosted` and `circle-hosted` etc. 

I think what @kommendorkapten is is right, if the verifier is mainly checking extension values against approved values, none of this would be a problem as you're just comparing strings.

> Totally agree. A better option IMHO is to push for the CI vendors to switch to alg:digest.

Yup something needs to happen here across the industry to support sha256 in git 🙈

docs/oid-info.md
feelepxyz: > Given that, I think many of these claims would benefit from `MAY` language that emphasize that providers may choose to fill these values as they please, e.g.:

@woodruffw good idea! I added this as top-level call out to cover all the values.

docs/oid-info.md
feelepxyz: TBH seems ok as is. Should we get this in as is and noodle on specific GHA details when implementing if needed?

kommendorkapten (dismissed):
Nice job! 🚀

docs/oid-info.md
haydentherapper: Makes sense, although you could still run into issues if the OIDC issuers are run within the same trust domain such that if one is compromised, both could be. For example a CI issuer issuing a token for an email address, or an email-based issuer issuing a token for a CI identity (we do add some checks in fulcio to make sure a subject looks like an email when it's expected, I would hesitate to always enforce that a CI identity is not an email though)

docs/oid-info.md
haydentherapper: I'm fine updating in a follow up, but my plan would be to use exactly what is in this doc when implementing for any CI provider (also makes it easier to audit). So if you want to prefix a URI with server_url, we should specify it.

docs/oid-info.md
haydentherapper: Closing on this, I believe we're all in agreement now to not include a hash algorithm and copy the value as-is, correct? We can always update during implementation if we see a need for it.

haydentherapper (dismissed):
Fantastic job on this and thanks for the many iterations!
I'll leave it open til Monday morning for any remaining comments and merge then.

A small follow-up (or feel free to do it in this PR too, I'll reapprove) is to make sure the GitHub mapping of OID value to claims is accurate (with any prefixes or concatenations noted), as I'll use this as the guide for implementation.

bobcallaway (approved):
lgtm, thx all!

haydentherapper (approved):

kommendorkapten (approved):

docs/oid-info.md
```diff
@@ -2,52 +2,91 @@
...
+
+In order to support Sigstore code signing with CI/CD based workfload identities the following claims must be included in the OIDC ID Token. See example claim values for each extension in the detailed [Directory](#directory). 
+
+Providers MAY choose to emit extension value in other formats to generic examples, and consumers MUST NOT assume the generic example format.
+
+Requirements:
+
+- MUST include `iss` claim for `Issuer` extension.
+- MUST include claim to support: `Build Signer URI` that identifies the specific build instructions that are responsible for signing.
+- MUST include claim to support: `Build Signer Digest` which is an immutable reference to a specific version of the build instructions that are responsible for signing.
```
sj26, line 21: How should these attributes be supported by services like Buildkite which offer dynamic pipelines, i.e. build instructions are not immutable and change over the course of an execution? I suppose we could include a reference to the _initial_ build instructions?

sj26: Ah, I see, the directory examples suggest this could be a link to the source repository at a specific commit sha. That doesn't represent the whole signing environment for us, the instructions also involve the running pipeline, but it might be close enough.

steiza: Yes, so ideally this would be an immutable reference to the Buildkite pipeline steps at the time the build ran.

How are Buildkite pipeline steps stored / versioned? If I update the pipeline steps **after** a build has run, is there a URL I can go to that references the previous version of the pipeline steps (e.g. the pipeline steps when the build ran)?

PR #966, opened 2023-01-17
Add custom error logs when communicating with the CA backend

hectorj2f:

#### Summary

We realized most of the errors returned the same error message (e.g. genericCAError) when communicating with the CA backend. I added few more error logs and new error types depending on the logic of the error statements.

### DIFF
--- a/pkg/server/grpc_server.go
+++ b/pkg/server/grpc_server.go
@@ -140,6 +140,7 @@ func (g *grpcCAServer) CreateSigningCertificate(ctx context.Context, request *fu
 		// currently configured CA doesn't support pre-certificate flow required to embed SCT in final certificate
 		csc, err = g.ca.CreateCertificate(ctx, principal, publicKey)
 		if err != nil {
+			logger.Error("Error creating certificate with CA backend: ", err)
 			// if the error was due to invalid input in the request, return HTTP 400
 			if _, ok := err.(certauth.ValidationError); ok {
 				return nil, handleFulcioGRPCError(ctx, codes.InvalidArgument, err, err.Error())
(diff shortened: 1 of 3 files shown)


hectorj2f:
cc @priyawadhwa

pkg/server/grpc_server.go
```diff
@@ -140,6 +140,7 @@ func (g *grpcCAServer) CreateSigningCertificate(ctx context.Context, request *fu
 		// currently configured CA doesn't support pre-certificate flow required to embed SCT in final certificate
 		csc, err = g.ca.CreateCertificate(ctx, principal, publicKey)
 		if err != nil {
+			logger.Error("Error creating certificate with CA backend: ", err)
```
bobcallaway, line 143: Perhaps just wrap `err` with this string - a log statement will fire in the subsequent call to `handleFulcioGRPCError`

same comment in other cases below.

hectorj2f: True! I did that for those whose error messages are similar to the fired error. There are few that I kept so they return  descriptive error message to easily distinguish where they are occurring.

bobcallaway: Sorry, after re-reading my comment I think I could have worded it better.

My thought was that within each of those `if err != nil` conditions, a log message will be generated to reflect the error (within the call to `handleFulcioGRPCError`). I agree that the errors could be scoped to be more useful, but with your change we'll have two log entries with slightly different error messages which might confuse someone. 

I was thinking of something like:

```
if err != nil {
    err = fmt.Errorf("Error creating certificate with CA backend: %w", err)
    // all code as is below
    // ...
}
```

That way you get the specificity you're looking for without the duplicate log message. Curious to hear your thoughts on that approach.

hectorj2f: I can change that to avoid duplicating log errors. However I'd like to keep some custom err messages to be consistent with the current implementation of other error types, e.g. invalidIdentityToken, invalidCSR error messages.

haydentherapper (approved):
thanks!

Issue #970, opened 2023-01-20
AKS issuer URL is different from configured format


yorinasub17:
**Description**

It appears that AKS has updated the OIDC Issuer URL format and it no longer matches what Fulcio expects. The AKS cluster that I have running in the East US region has the following OIDC Issuer URL for the workload identity preview:

```
%~> az aks show -n myCluster -g myGroup --query "oidcIssuerProfile.issuerUrl" -otsv
The behavior of this command has been altered by the following extension: aks-preview
https://eastus.oic.prod-aks.azure.com/****
```

This may also be a difference in the specific region that the AKS cluster is provisioned in.

**Version**

cosign: v2.0.0-rc.0
fulcio: Public instance (`https://fulcio.sigstore.dev`)

haydentherapper:
@yorinasub17 Do you want to try now, https://fulcio.sigstore.dev/api/v2/configuration shows it should be rolled out

yorinasub17:
Just tried and it worked! I was able to successfully get a signing certificate from `fulcio` using `cosign`! Many thanks for the review and help getting this rolled out!

PR #384, opened 2022-02-03
Add AKS as a meta issuer

tcnghia:
#### Summary
Add AKS as a meta issuer, now that AKS OIDC Issuer feature is available. 

This was tested using an AKS cluster with the feature `EnableOIDCIssuerPreview` following instruction [here](https://docs.microsoft.com/en-us/azure/aks/cluster-configuration#oidc-issuer-preview). 

Signed-off-by: Nghia Tran <tcnghia@gmail.com>

#### Ticket Link

Fixes https://github.com/sigstore/fulcio/issues/231

#### Release Note

```release-note
Config change to support AKS per-cluster issuer URLs.
```

PR #971, opened 2023-01-20
Add new format for AKS OIDC issuer

yorinasub17:
Closes https://github.com/sigstore/fulcio/issues/970

#### Summary

This adds the newly detected OIDC Issuer URL format for AKS clusters as an additional meta issuer. I added a new one instead of replacing the existing one in case that issuer URL is being used under different circumstances.

#### Release Note

* Support the new format of AKS OIDC Issuers

#### Documentation

N/A

### DIFF
--- a/config/fulcio-config.yaml
+++ b/config/fulcio-config.yaml
@@ -41,6 +41,10 @@ data:
             }
           },
           "MetaIssuers": {
+            "https://*.oic.prod-aks.azure.com/*": {
+              "ClientID": "sigstore",
+              "Type": "kubernetes"
+            },
             "https://container.googleapis.com/v1/projects/*/locations/*/clusters/*": {
               "ClientID": "sigstore",
               "Type": "kubernetes"
@@ -50,10 +54,6 @@ data:
               "Type": "kubernetes"
             },
             "https://oidc.prod-aks.azure.com/*": {
-              "ClientID": "sigstore",
-              "Type": "kubernetes"
-            },
-            "https://*.oic.prod-aks.azure.com/*": {
               "ClientID": "sigstore",
               "Type": "kubernetes"
             }
--- a/config/fulcio-config.yaml
+++ b/config/fulcio-config.yaml
@@ -52,6 +52,10 @@ data:
             "https://oidc.prod-aks.azure.com/*": {
               "ClientID": "sigstore",
               "Type": "kubernetes"
+            },
+            "https://*.oic.prod-aks.azure.com/*": {
+              "ClientID": "sigstore",
+              "Type": "kubernetes"
             }
           }
         }
--- a/federation/main.go
+++ b/federation/main.go
@@ -76,6 +76,10 @@ func main() {
 				ClientID: "sigstore",
 				Type:     "kubernetes",
 			},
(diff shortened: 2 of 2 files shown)


config/fulcio-config.yaml
```diff
@@ -52,6 +52,10 @@ data:
             "https://oidc.prod-aks.azure.com/*": {
```
haydentherapper, line 52: Do you have any documentation from Azure about this? Is there a migration ongoing? I would assume this has to do with providing regionalized OIDC providers.

haydentherapper: There are a few references in the Azure org to this new URL - https://cs.github.com/?scopeName=All+repos&scope=&q=oic.prod-aks.azure.com+org%3Aazure

Contrast to https://cs.github.com/?scopeName=All+repos&scope=&q=https%3A%2F%2Foidc.prod-aks.azure.com+org%3Aazure, with more references in code. 

I can't find any public documentation about this, but seems fine to me.

yorinasub17: Unfortunately I couldn't find any documentation on this. I actually don't have much more information than what you could find (I'm not associated with Azure/Microsoft).

The only thing I do know is that:

- I created the AKS cluster on December 20th in the East US region, following the same steps referenced in https://github.com/sigstore/fulcio/pull/384 ([these docs](https://learn.microsoft.com/en-us/azure/aks/cluster-configuration#oidc-issuer-preview))
- The OIDC issuer URL that I got is different from what's configured here: `https://eastus.oic.prod-aks.azure.com/****`

I posted a question on their discussion forum (https://learn.microsoft.com/en-us/answers/questions/1162598/what-is-the-canonical-format-of-the-oidc-issuer-ur). Hopefully someone from Microsoft can answer and give us some concrete info we can rely on!

yorinasub17: Also posted on their main AKS issue for the feature (since that pings the right team): https://github.com/Azure/AKS/issues/1480#issuecomment-1397808462

yorinasub17: UPDATE: I got a response from Microsoft support on [the learn Q&A ticket](https://learn.microsoft.com/en-us/answers/questions/1162598/what-is-the-canonical-format-of-the-oidc-issuer-ur) which isn't definitive, but is another data point that the format appears to be `https://REGION.oic.prod-aks.azure.com/***`.

Is that sufficient data for adding this in? If not, is there something else I can do to unblock myself (e.g., is there a way for me to run a custom fulcio instance and link with the public rekor. I think in theory that's possible but not sure...)?

FWIW, I've also opened [a ticket on the Microsoft docs page](https://github.com/MicrosoftDocs/azure-docs/issues/104085) requesting for the format to be documented.

yorinasub17:
@haydentherapper Looks like the build failed again, and I see what I did wrong, so this time I generated the file locally and pushed the diff ([3289328](https://github.com/sigstore/fulcio/pull/971/commits/32893285e6e4ad5dd7e5c61c85d2b8cb749ca6a3)). I believe this should cause the test to pass now!

haydentherapper:
Thanks! I'll get this checked in and rolled out to prod soon after.

PR #996, opened 2023-02-08
Add an option to skip clientID checks  for OIDC issuers

hectorj2f:

#### Summary

We should expose an option to skip the ClientID verification for the OIDC Issuers on demand.

#### Release Note

Add option to skip the ClientID verifications for the configured OIDC Issuers.

### DIFF
--- a/federation/main.go
+++ b/federation/main.go
@@ -46,6 +46,7 @@ type federationConfig struct {
 	Type              string
 	IssuerClaim       string
 	SpiffeTrustDomain string
+	SkipClientIDCheck bool
 }
 
 func main() {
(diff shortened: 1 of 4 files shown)


haydentherapper (commented):
What is the reason for this change? I would prefer we not make this optional, this is a part of OIDC token verification.

hectorj2f:
@haydentherapper Yes, I understand it is not ideal. But we want to tell Fulcio to skip checking the token’s audience on-demand.

haydentherapper:
I'm still uncertain why that would be needed unless you're testing with malformed tokens. A token should always contain an audience, and that audience should be static per provider. Fulcio operators should not be able to ignore audience, that opens them up to some risk.

hectorj2f:
We aren't using malformed tokens. This is only required for OIDC issuer types working with Github-Actions. We need to instruct Fulcio to skip clientID checks, or enable Fulcio to accept custom issuer types which isn't possible today.  
cc @priyawadhwa

haydentherapper:
GitHub Actions supports setting a custom audience, that’s what’s currently done when GHAs call fulcio.

https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#:~:text=you%20can%20set%20a%20custom%20audience%20with%20a%20toolkit%20command%3A%20core.getidtoken(audience)

I also don’t think OIDC token verifiers should accept multiple client IDs, you should be given tokens for an expected audience.

haydentherapper:
The spec allows for multiple audiences, so it would be fine if we supported multiple audiences (albeit I’m not sure the use case) - https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation

We aren’t going to add something that allows for deviation from the spec and a potential foot-gun for service operators.

hectorj2f:
@haydentherapper Sorry for my late reply! Thanks for the comments. We'll reconsider our alternatives here and attempt a different approach.

haydentherapper:
@hectorj2f Good to close this PR?

haydentherapper:
Closing since I believe the other PRs will supersede this, lemme know if not.

PR #1007, opened 2023-02-16
Allow specifying ChallengeClaim for an Issuer in the Fulcio config

priyawadhwa:
This will allow us to start setting up custom issuers! Right now this [switch](https://github.com/sigstore/fulcio/blob/main/pkg/config/config.go#L462) statement prevents users from doing that, but allowing the challenge claim to be specified in the Fulcio config should resolve that.


cpanato (commented):
lgtm

haydentherapper (approved):
Thanks!

PR #1008, opened 2023-02-16
Use Issuer interface to allow for custom issuers

priyawadhwa:
This plumbs through IssuerPool and starts using it to set up Fulcio.


haydentherapper:
LGTM with the concept, as we chatted about, server implementers who want to use private implementations of an issuer will instantiate the gRPC server with their own issuer pool.

haydentherapper:
Will take a look Monday!

priyawadhwa:
@haydentherapper just so you know it's still a WIP, have a couple more things to add!

pkg/server/issuer_pool.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	for meta, i := range cfg.MetaIssuers {
+		ip = append(ip, getIssuer(meta, i))
+	}
+	return ip
+}
+
+func getIssuer(meta string, i config.OIDCIssuer) identity.Issuer {
+	issuerURL := i.IssuerURL
+	if meta == "" {
```
haydentherapper, line 42: != ?

pkg/server/grpc_server_test.go
```diff
@@ -522,6 +524,9 @@ func TestAPIWithUriSubject(t *testing.T) {
 		t.Fatalf("config.Read() = %v", err)
 	}
 
+	// Give a few seconds for the oidc servers to start up
+	time.Sleep(5 * time.Second)
```
haydentherapper, line 528: It'd be preferable to poll an endpoint until it returns something, maybe do this in newOIDCIssuer?

pkg/identity/base/issuer.go
```diff
@@ -20,9 +20,15 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/sigstore/fulcio/pkg/identity"
 )
 
+var (
+	// For testing
+	CmpOptions = cmpopts.IgnoreUnexported(baseIssuer{})
```
haydentherapper, line 29: Just curious, what was the reason for this?

priyawadhwa: when you use `cmp.Diff` you need to specify that unexported fields can be ignored when comparing two objects, otherwise the library panics. it's discussed a bit in the project [readme](https://github.com/google/go-cmp) but I'm using it for testing [here](https://github.com/sigstore/fulcio/pull/1008/files#diff-82a7a7c995f812a9240e539c523cf8f8c1b8484598447464beb9a0681e17a7ecR48)

pkg/server/grpc_server_test.go
priyawadhwa: i don't think we need this actually, the bug you caught above was the real issue!

priyawadhwa:
@haydentherapper this should be RFAL! Thanks :)

pkg/server/grpc_server_test.go
```diff
@@ -522,6 +524,8 @@ func TestAPIWithUriSubject(t *testing.T) {
 		t.Fatalf("config.Read() = %v", err)
 	}
 
+	t.Logf("%v", cfg)
```
haydentherapper, line 527: nit: delete

haydentherapper (commented):
Do you want to delete https://github.com/sigstore/fulcio/blob/main/pkg/challenges/challenges.go#L53-L83 also?

priyawadhwa:
@haydentherapper i was planning on deleting everything in a follow-up!

haydentherapper (approved):
Sounds good! Thanks!

PR #1039, opened 2023-03-06
Create BaseIssuer type to implement Match for all Issuers

priyawadhwa:
Also includes support for regex's in Match for MetaIssuer types

### DIFF
--- a/pkg/identity/base/issuer.go
+++ b/pkg/identity/base/issuer.go
@@ -1,0 +1,66 @@
+// Copyright 2023 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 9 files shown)


cpanato (approved):
looks nice

haydentherapper (commented):
Looks good!

pkg/identity/base/issuer.go
```diff
@@ -0,0 +1,66 @@
...
+	"strings"
+
+	"github.com/sigstore/fulcio/pkg/identity"
+)
+
+type baseIssuer struct {
+	issuerURL string
+}
+
+func Issuer(issuerURL string) identity.Issuer {
```
haydentherapper, line 30: Do we need a public Issuer function? If this is only used for tests, can we make it private or just initialize with the `baseIssuer` type?

priyawadhwa: all of the other issuer types call it!

PR #1066, opened 2023-03-16
Update deps in /hack/tools

haydentherapper:


cpanato (approved):
thank you!

i was planning to do that today but got side tracked :(

Issue #243, opened 2021-11-18
Add support for GitLab tokens


dlorenc:
It looks like GitLab offers JWT tokens as well to CI jobs:

https://docs.gitlab.com/ee/ci/examples/authenticating-with-hashicorp-vault/#how-it-works

They're available as the `CI_JOB_JWT` environment variable, and the endpoint is `https://gitlab.example.com/-/jwks`.

It sounds like they don't support the audience field, which isn't great but should still be workable.

dlorenc:
cc @colek42

colek42:
It looks like this is a premium feature for GitLab as well.

dlorenc:
Hmm, over here it says it might be in all job tiers: https://docs.gitlab.com/ee/ci/secrets/

mattmoor:
```
curl https://gitlab.com/.well-known/openid-configuration
```

```json
{
  "issuer": "https://gitlab.com",
  "authorization_endpoint": "https://gitlab.com/oauth/authorize",
  "token_endpoint": "https://gitlab.com/oauth/token",
  "revocation_endpoint": "https://gitlab.com/oauth/revoke",
  "introspection_endpoint": "https://gitlab.com/oauth/introspect",
  "userinfo_endpoint": "https://gitlab.com/oauth/userinfo",
  "jwks_uri": "https://gitlab.com/oauth/discovery/keys",
  "scopes_supported": [
    "api",
    "read_user",
    "read_api",
    "read_repository",
    "write_repository",
    "read_registry",
    "write_registry",
    "sudo",
    "openid",
    "profile",
    "email"
  ],
  "response_types_supported": [
    "code",
    "token"
  ],
  "response_modes_supported": [
    "query",
    "fragment"
  ],
  "grant_types_supported": [
    "authorization_code",
    "implicit",
    "password",
    "client_credentials",
    "refresh_token"
  ],
  "token_endpoint_auth_methods_supported": [
    "client_secret_basic",
    "client_secret_post"
  ],
  "subject_types_supported": [
    "public"
  ],
  "id_token_signing_alg_values_supported": [
    "RS256"
  ],
  "claim_types_supported": [
    "normal"
  ],
  "claims_supported": [
    "iss",
    "sub",
    "aud",
    "exp",
    "iat",
    "sub_legacy",
    "name",
    "nickname",
    "email",
    "email_verified",
    "website",
    "profile",
    "picture",
    "groups",
    "groups_direct"
  ]
}

```

[...]

PR #1097, opened 2023-04-05
Map GitLab OIDC token claims to Fulcio OIDs

aladh:
#### Summary

GitLab issue: https://gitlab.com/gitlab-org/gitlab/-/issues/388517

Related to https://github.com/sigstore/fulcio/issues/243 and https://github.com/sigstore/fulcio/pull/983

Note that some of the GitLab claims added in this PR do not exist in the GitLab token yet.


docs/oid-info.md
```diff
@@ -182,7 +182,7 @@ that Sigstore operates.
 
 | GitHub [(docs)][github-oidc-doc] | GitLab | CircleCI | Buildkite | Fulcio Certificate Extension | Why / Notes / Questions                                                                                                                                                                |
```
aladh, line 183:
```suggestion
| GitHub [(docs)][github-oidc-doc] | GitLab [(docs)][https://docs.gitlab.com/ee/ci/cloud_services/#how-it-works] | CircleCI | Buildkite | Fulcio Certificate Extension | Why / Notes / Questions                                                                                                                                                                |
```

Seems like we have the claims documented in two places, not sure which one is better:
- https://docs.gitlab.com/ee/ci/cloud_services/#how-it-works
- https://docs.gitlab.com/ee/ci/secrets/id_token_authentication.html#token-payload

docs/oid-info.md
```diff
@@ -182,12 +182,12 @@ that Sigstore operates.
...
 | nbf                | nbf     | nbf       | nbf        | N/A                          | Only used to validate the JWT.                                                                                                                                                         |
 | iat                | iat     | iat       | iat        | N/A                          | Only used to validate the JWT.                                                                                                                                                         |
-| server_url + job_workflow_ref   | ??     | ??       | ??        | Build Signer URI             | Reference to specific build instructions that are responsible for signing. Can be the same as Build Config URI. For example a reusable workflow in GitHub Actions or a Circle CI Orbs. |
+| server_url + job_workflow_ref   | server_url + project_path + /-/blob/ + ci_config_sha + / + ci_config_path | ??       | ??        | Build Signer URI             | Reference to specific build instructions that are responsible for signing. Can be the same as Build Config URI. For example a reusable workflow in GitHub Actions or a Circle CI Orbs. |
```
aladh, line 190: `ci_config_sha` and `ci_config_path`  are claims we would need to add to the token.

docs/oid-info.md
```diff
@@ -182,12 +182,12 @@ that Sigstore operates.
...
 | job_workflow_sha   | ??     | ??       | ??        | Build Signer Digest          | An immutable reference to the specific version of the build instructions that is responsible for signing. Should include the digest type followed by the digest, e.g. `sha1:abc123`.   |
```
aladh, line 191: Is the `sha` prefix required here? The description of Build Signer Digest above does not mention this.

aladh:
```suggestion
| job_workflow_sha   | ci_config_sha     | ??       | ??        | Build Signer Digest          | An immutable reference to the specific version of the build instructions that is responsible for signing. Should include the digest type followed by the digest, e.g. `sha1:abc123`.   |
```

haydentherapper:
Thanks for working on this! If you can fill in the MUST extensions in particular, that would be very helpful!

docs/oid-info.md
```diff
@@ -182,12 +182,12 @@ that Sigstore operates.
...
 | runner_environment | ??     | ??       | ??        | Runner Environment           | For platforms to specify whether the build took place in platform-hosted cloud infrastructure or customer-hosted infrastructure. For example: `platform-hosted` and `self-hosted`.     |
```
aladh, line 192:
```suggestion
| runner_environment |  runner_type  | ??       | ??        | Runner Environment           | For platforms to specify whether the build took place in platform-hosted cloud infrastructure or customer-hosted infrastructure. For example: `platform-hosted` and `self-hosted`.     |
```

Could be `instance`, `group`, or `project`.

docs/oid-info.md
aladh: Should this be the same as Build Config URI?

docs/oid-info.md
```diff
@@ -182,12 +182,12 @@ that Sigstore operates.
...
 | server_url + repository         | project_path     | ??       | ??        | Source Repository URI                   | Should include a fully qualified repository URL.                                                                                                                                       |
```
aladh, line 193:
```suggestion
| server_url + repository         | server_url + project_path | ??       | ??        | Source Repository URI                   | Should include a fully qualified repository URL.                                                                                                                                       |
```

docs/oid-info.md
```diff
@@ -182,15 +182,15 @@ that Sigstore operates.
...
 | runner_environment | ??     | ??       | ??        | Runner Environment           | For platforms to specify whether the build took place in platform-hosted cloud infrastructure or customer-hosted infrastructure. For example: `platform-hosted` and `self-hosted`.     |
-| server_url + repository         | project_path     | ??       | ??        | Source Repository URI                   | Should include a fully qualified repository URL.                                                                                                                                       |
+| server_url + repository         | server_url + project_path | ??       | ??        | Source Repository URI                   | Should include a fully qualified repository URL.                                                                                                                                       |
 | sha                | ??     | ??       | build_commit | Source Repository Digest                | An immutable reference to a specific version of the source code. Should include the digest type followed by the digest, e.g. `sha1:abc123`.                                            |
```
aladh, line 194:
```suggestion
| sha                | ci_commit_sha | ??       | build_commit | Source Repository Digest                | An immutable reference to a specific version of the source code. Should include the digest type followed by the digest, e.g. `sha1:abc123`.                                            |
```

docs/oid-info.md
aladh:
```suggestion
| server_url + job_workflow_ref   | server_url + project_path + /-/blob/ + ci_commit_sha + / + ci_config_path | ??       | ??        | Build Signer URI             | Reference to specific build instructions that are responsible for signing. Can be the same as Build Config URI. For example a reusable workflow in GitHub Actions or a Circle CI Orbs. |
```

docs/oid-info.md
aladh: It would be useful to have runner ID and runner description in the claims as well. Can we put them here and concatenate it all together, or should they go in a different claim? e.g. `instance-11574084-3-green.shared-gitlab-org.runners-manager.gitlab.com`

docs/oid-info.md
```diff
@@ -182,16 +182,16 @@ that Sigstore operates.
...
 | nbf                | nbf     | nbf       | nbf        | N/A                          | Only used to validate the JWT.                                                                                                                                                         |
 | iat                | iat     | iat       | iat        | N/A                          | Only used to validate the JWT.                                                                                                                                                         |
-| server_url + job_workflow_ref   | ??     | ??       | ??        | Build Signer URI             | Reference to specific build instructions that are responsible for signing. Can be the same as Build Config URI. For example a reusable workflow in GitHub Actions or a Circle CI Orbs. |
+| server_url + job_workflow_ref   | server_url + project_path + /-/blob/ + ci_commit_sha + / + ci_config_path | ??       | ??        | Build Signer URI             | Reference to specific build instructions that are responsible for signing. Can be the same as Build Config URI. For example a reusable workflow in GitHub Actions or a Circle CI Orbs. |
```
aladh, line 190: Example URL: https://gitlab.com/gitlab-org/gitlab/-/blob/d33e88e2d80b97b0c15e5c11aaf90e17d0916fda/.gitlab-ci.yml

Should this be the same as Build Config URI?

docs/oid-info.md
```diff
@@ -182,16 +182,16 @@ that Sigstore operates.
...
+| sha                | ci_commit_sha | ??       | build_commit | Source Repository Digest                | An immutable reference to a specific version of the source code. Should include the digest type followed by the digest, e.g. `sha1:abc123`.                                            |
 | ref                | ref     | ??       | build_branch | Source Repository Ref                   | The source ref that the build run was based upon. For example: refs/head/main.                                                                                                         |
 | repository_id         | project_id     | ??       | ??        | Source Repository Identifier     | Stable identifier for the owner of the source repository.                                                                                                                                      |
 | server_url + repository_owner         | namespace_path     | ??       | ??        | Source Repository Owner URI     | Fully qualified URL for the owner of the source repository.                                                                                                                                     |
```
aladh, line 197:
```suggestion
| server_url + repository_owner         | server_url + namespace_path | ??       | ??        | Source Repository Owner URI     | Fully qualified URL for the owner of the source repository.                                                                                                                                     |
```

docs/oid-info.md
```diff
@@ -182,19 +182,19 @@ that Sigstore operates.
...
+| server_url + repository_owner         | server_url + namespace_path | ??       | ??        | Source Repository Owner URI     | Fully qualified URL for the owner of the source repository.                                                                                                                                     |
 | repository_owner_id         | namespace_id     | ??       | ??        | Source Repository Owner Identifier     | Stable identifier for the owner of the source repository.                                                                                                                                   |
 | server_url + workflow_ref       | ??     | ??       | ??        | Build Config URI             | A reference to the initiating build instructions.                                                                                                                                      |
```
aladh, line 199:
```suggestion
| server_url + workflow_ref       |  server_url + project_path + /-/blob/ + ci_commit_sha + / + ci_config_path  | ??       | ??        | Build Config URI             | A reference to the initiating build instructions.                                                                                                                                      |
```

docs/oid-info.md
```diff
@@ -182,19 +182,19 @@ that Sigstore operates.
...
 | workflow_sha       | ??     | ??       | ??        | Build Config Digest          | An immutable reference to the specific version of the top-level build instructions. Should include the digest type followed by the digest, e.g. `sha1:abc123`.                         |
```
aladh, line 200:
```suggestion
| workflow_sha       |  ci_commit_sha  | ??       | ??        | Build Config Digest          | An immutable reference to the specific version of the top-level build instructions. Should include the digest type followed by the digest, e.g. `sha1:abc123`.                         |
```

docs/oid-info.md
```diff
@@ -182,22 +182,22 @@ that Sigstore operates.
...
+| server_url + workflow_ref       |  server_url + project_path + /-/blob/ + ci_commit_sha + / + ci_config_path  | ??       | ??        | Build Config URI             | A reference to the initiating build instructions.                                                                                                                                      |
+| workflow_sha       |  ci_commit_sha  | ??       | ??        | Build Config Digest          | An immutable reference to the specific version of the top-level build instructions. Should include the digest type followed by the digest, e.g. `sha1:abc123`.                         |
 | event_name         | pipeline_source | ??       | ??        | Build Trigger                | The event or action that triggered the build.                                                                                                                                          |
 | server_url + repository + "/actions/runs/" + run_id + "/attempts/" + run_attempt | ??     | ??       | ??        | Run Invocation URI           | An immutable identifier that can uniquely identify the build execution     |
```
aladh, line 202:
```suggestion
| server_url + repository + "/actions/runs/" + run_id + "/attempts/" + run_attempt | server_url + project_path + /-/jobs/ + job_id    | ??       | ??        | Run Invocation URI           | An immutable identifier that can uniquely identify the build execution     |
```

aladh: Example URL: https://gitlab.com/gitlab-org/gitlab/-/jobs/4068841111

docs/oid-info.md
```diff
@@ -182,24 +182,24 @@ that Sigstore operates.
...
+| workflow_sha       |  ci_commit_sha  | ??       | ??        | Build Config Digest          | An immutable reference to the specific version of the top-level build instructions. Should include the digest type followed by the digest, e.g. `sha1:abc123`.                         |
 | event_name         | pipeline_source | ??       | ??        | Build Trigger                | The event or action that triggered the build.                                                                                                                                          |
-| server_url + repository + "/actions/runs/" + run_id + "/attempts/" + run_attempt | ??     | ??       | ??        | Run Invocation URI           | An immutable identifier that can uniquely identify the build execution     |
+| server_url + repository + "/actions/runs/" + run_id + "/attempts/" + run_attempt | server_url + project_path + /-/jobs/ + job_id    | ??       | ??        | Run Invocation URI           | An immutable identifier that can uniquely identify the build execution     |
```
aladh, line 202: Example URL: https://gitlab.com/gitlab-org/gitlab/-/jobs/4068841111

docs/oid-info.md
haydentherapper: Fulcio will not prepend the digest to the claim, we'll just copy the value from the claim.

aladh:
@Brcrwilliams @cpanato @marshall007 could you please take a look and add any thoughts?

docs/oid-info.md
```diff
@@ -182,24 +182,24 @@ that Sigstore operates.
...
-| server_url + job_workflow_ref   | ??     | ??       | ??        | Build Signer URI             | Reference to specific build instructions that are responsible for signing. Can be the same as Build Config URI. For example a reusable workflow in GitHub Actions or a Circle CI Orbs. |
-| job_workflow_sha   | ??     | ??       | ??        | Build Signer Digest          | An immutable reference to the specific version of the build instructions that is responsible for signing. Should include the digest type followed by the digest, e.g. `sha1:abc123`.   |
+| server_url + job_workflow_ref   | server_url + project_path + /-/blob/ + ci_commit_sha + / + ci_config_path | ??       | ??        | Build Signer URI             | Reference to specific build instructions that are responsible for signing. Can be the same as Build Config URI. For example a reusable workflow in GitHub Actions or a Circle CI Orbs. |
+| job_workflow_sha   | ci_commit_sha | ??       | ??        | Build Signer Digest          | An immutable reference to the specific version of the build instructions that is responsible for signing. Should include the digest type followed by the digest, e.g. `sha1:abc123`.   |
```
wlynch, line 191: Same thing here for we may want this to reference the revision in `include` (since this can differ from the commit that kicked off the build).

docs/oid-info.md
```diff
@@ -182,24 +182,24 @@ that Sigstore operates.
...
 | repository_id         | project_id     | ??       | ??        | Source Repository Identifier     | Stable identifier for the owner of the source repository.                                                                                                                                      |
-| server_url + repository_owner         | namespace_path     | ??       | ??        | Source Repository Owner URI     | Fully qualified URL for the owner of the source repository.                                                                                                                                     |
+| server_url + repository_owner         | server_url + namespace_path | ??       | ??        | Source Repository Owner URI     | Fully qualified URL for the owner of the source repository.                                                                                                                                     |
```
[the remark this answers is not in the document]
cpanato: yes correct

docs/oid-info.md
marshall007: > Additionally, include is not versioned and it could change between multiple different CI pipeline runs.

@Brcrwilliams the immutability problem is not unique to us and it can't be solved in this context anyway. A workflow ref in GitHub Actions has no immutability guarantee either. Folks should be version pinning their `include` references by SHA if they want such guarantees. For local `include` references (like in your example) it doesn't even matter because the ref is the same as the root `.gitlab-ci.yml` pipeline definition.

We do need to account for [custom CI/CD config file locations](https://docs.gitlab.com/ee/ci/pipelines/settings.html#specify-a-custom-cicd-configuration-file) which can be set to anything including arbitrary remote HTTP endpoints. As such we may need to normalize this claim value to a fully qualified URI in all cases for consistency. We may need to manually checksum references to files that are not stored in git.

cc @wlynch

docs/oid-info.md
marshall007: > My initial thought (assuming GitLab runner descriptions are predefined labels and not user generated content)

@wlynch runner description is an arbitrary string. We just happen to be using a string that looks like a FQDN for shared runners on gitlab.com.

---

@aladh @Brcrwilliams from a supply chain security perspective, you cannot trust the results of builds executed on self-hosted runners. A self-hosted runner may receive its build instructions, executes something entirely different, and sends us fake logs indicating otherwise.

This claim is used specifically in verifying that the build was executed on managed infrastructure where stronger guarantees can be made.

@wlynch any thoughts around self-hosted instances? We plan to support managed runners for GitLab Dedicated instances eventually and potentially offer managed runner fleets for self-hosted instances. We have no details on that yet, but I think it's safe to assume that "managed runner fleets" could be connected at any level (instance/group/project) and I would still expect them to be considered `platform-hosted`.

I think it's really the fleet itself that may be managed or self-hosted and that is independent of the instance it is connected to and at what level. Perhaps users should be establishing trust based on the hostname of the runner fleet/manager specifically instead of the abstract claim?

docs/oid-info.md
marshall007: I think we should name the new claims `pipeline_ref` and `pipeline_sha`. For the reasons noted above, `ci_config_path` is not always going to be `.gitlab-ci.yml` and the checksum is not necessarily based on `ci_commit_sha`.

Also note, we will be making some pretty significant changes to pipeline definitions in the future (https://gitlab.com/gitlab-org/gitlab/-/issues/398129#note_1323114974) and may wish to change the URI format of `pipeline_ref` to align with the syntax of [CI pipeline components](https://docs.gitlab.com/ee/architecture/blueprints/ci_pipeline_components/). Keeping that logic for constructing that on our side is another good reason to include fully qualified URIs in the claim.

docs/oid-info.md
haydentherapper: Runner invocation may be a place to concatenate values together if there's other useful provenance claims

Note that the reason why we have runner_environment is so that clients can choose whether or not to trust the results - like you said, anything self-hosted should not be trusted for provenance metadata

wlynch: > @wlynch runner description is an arbitrary string. We just happen to be using a string that looks like a FQDN for shared runners on gitlab.com.

Can this value be arbitrarily modified by end users, or is it only set by the CI platform?

> @wlynch any thoughts around self-hosted instances?

+1 to what @haydentherapper said, though I think I have a slightly softer opinion in that trustworthy-ness of a runner pool is going to vary depending on the user. e.g.  A company may choose to trust it's own company-managed pool, but this may not be suitable for open source usage since it's opaque externally.
The important thing is to make sure end users have the data to distinguish these environments so they can write meaningful policy.

We generally have flexibility here, so long as the format is well-documented.

Brcrwilliams: > > @wlynch runner description is an arbitrary string. We just happen to be using a string that looks like a FQDN for shared runners on gitlab.com.
>
> Can this value be arbitrarily modified by end users, or is it only set by the CI platform?

This can be arbitrarily set by users via [the runner name](https://docs.gitlab.com/runner/configuration/advanced-configuration.html#the-runners-section). So, if we did it this way then it would be possible for self-hosted runners to impersonate platform-hosted runners.

> Perhaps users should be establishing trust based on the hostname of the runner fleet/manager specifically instead of an opinionated abstraction?

I don't think that a solution based on hostname is a good idea, because we don't have any domain validation mechanism and runners can be inside any network.

> +1 to what @haydentherapper said, though I think I have a slightly softer opinion in that trustworthy-ness of a runner pool is going to vary depending on the user. e.g. A company may choose to trust it's own company-managed pool, but this may not be suitable for open source usage since it's opaque externally.
The important thing is to make sure end users have the data to distinguish these environments so they can write meaningful policy.

In that case, I think that the values proposed by Ali have the most flexibility for this. We can provide: `runner_type`, `instance_url`, `group_id`, `project_id`. Policies can be defined for each runner type.

- `runner_type=instance`: validate based on `instance_url` (e.g. `https://gitlab.com`)
- `runner_type=group`: validate based on `instance_url` and `group_id`
- `runner_type=project`: validate based on `instance_url` and `project_id`

docs/oid-info.md
marshall007: @wlynch just for clarification, other providers do not appear to provide digests for the fully expanded workflow/pipeline/job definition either. GitHub, for example, is only providing the commit SHA in their `workflow_sha` claim.

I agree this will be great to have eventually, but unless I'm missing something, including the commit SHA in "Build Config Digest" would be consistent with GitHub.

docs/oid-info.md
marshall007: @Brcrwilliams I was looking at this a bit and I am not sure there is value in including the `runner_type` claim at the moment. For `runner_type=group` does not tell you which group the runner is associated with. Likewise, `runner_type=project` runners can be shared across multiple projects: https://docs.gitlab.com/ee/ci/runners/runners_scope.html#enable-a-project-runner-for-a-different-project

I propose that we keep things simple for now and provide `runner_environment=gitlab-hosted|self-hosted` computed on our side.

Brcrwilliams: > I propose that we keep things simple for now and provide runner_environment=gitlab-hosted|self-hosted computed on our side.

Works for me. Let's do that.

docs/oid-info.md
wlynch: GitHub is a bit different in that it makes the distinction between `workflow_sha` for the entire workflow vs `job_workflow_sha` that's just for the particular job.

We don't have a similar concept of referencing a digest for a job in GitLab. While we can use the pipeline_sha to try to and get the job, I don't think it actually gives us much extra value than if we omitted it.

I was thinking that it would look something like:

Field | Value
-- | --
Build Signer URI | https://gitlab.com/org/repo/-/job/1234
Build Signer Digest | 
Build Config URI | https://gitlab.com/org/repo/.gitlab-ci.yml (pipeline_ref)
Build Config SHA | abcdef1234 (pipeline_sha)
Run Invocation URI | https://gitlab.com/org/repo/-/pipeline/1234

marshall007: @wlynch ah, agreed! I missed that "Build Signer Digest" was a separate field. We will revisit once CI components are implemented: https://docs.gitlab.com/ee/architecture/blueprints/ci_pipeline_components/#definition-of-pipeline-component

docs/oid-info.md
```diff
@@ -180,26 +180,26 @@ that Sigstore operates.
 
 ## Mapping OIDC token claims to Fulcio OIDs
 
-| GitHub [(docs)][github-oidc-doc] | GitLab | CircleCI | Buildkite | Fulcio Certificate Extension | Why / Notes / Questions                                                                                                                                                                |
+| GitHub [(docs)][github-oidc-doc] | GitLab [(docs)][https://docs.gitlab.com/ee/ci/secrets/id_token_authentication.html#token-payload] | CircleCI | Buildkite | Fulcio Certificate Extension | Why / Notes / Questions                                                                                                                                                                |
```
aladh, line 183:
```suggestion
| GitHub [(docs)][github-oidc-doc] | GitLab [(docs)](https://docs.gitlab.com/ee/ci/secrets/id_token_authentication.html#token-payload]) | CircleCI | Buildkite | Fulcio Certificate Extension | Why / Notes / Questions                                                                                                                                                                |
```

docs/oid-info.md
```diff
@@ -180,26 +180,26 @@ that Sigstore operates.
 
 ## Mapping OIDC token claims to Fulcio OIDs
 
-| GitHub [(docs)][github-oidc-doc] | GitLab | CircleCI | Buildkite | Fulcio Certificate Extension | Why / Notes / Questions                                                                                                                                                                |
+| GitHub [(docs)][github-oidc-doc] | GitLab [(docs)](https://docs.gitlab.com/ee/ci/secrets/id_token_authentication.html#token-payload]) | CircleCI | Buildkite | Fulcio Certificate Extension | Why / Notes / Questions                                                                                                                                                                |
```
aladh, line 183:
```suggestion
| GitHub [(docs)][github-oidc-doc] | GitLab [(docs)](https://docs.gitlab.com/ee/ci/secrets/id_token_authentication.html#token-payload) | CircleCI | Buildkite | Fulcio Certificate Extension | Why / Notes / Questions                                                                                                                                                                |
```

docs/oid-info.md
```diff
@@ -182,24 +182,24 @@ that Sigstore operates.
...
 | iat                | iat     | iat       | iat        | N/A                          | Only used to validate the JWT.                                                                                                                                                         |
-| server_url + job_workflow_ref   | ??     | ??       | ??        | Build Signer URI             | Reference to specific build instructions that are responsible for signing. Can be the same as Build Config URI. For example a reusable workflow in GitHub Actions or a Circle CI Orbs. |
-| job_workflow_sha   | ??     | ??       | ??        | Build Signer Digest          | An immutable reference to the specific version of the build instructions that is responsible for signing. Should include the digest type followed by the digest, e.g. `sha1:abc123`.   |
+| server_url + job_workflow_ref   | server_url + project_path + /-/blob/ + ci_commit_sha + / + ci_config_path | ??       | ??        | Build Signer URI             | Reference to specific build instructions that are responsible for signing. Can be the same as Build Config URI. For example a reusable workflow in GitHub Actions or a Circle CI Orbs. |
```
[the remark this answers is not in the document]
aladh: We can leave this PR open until https://gitlab.com/gitlab-org/gitlab/-/merge_requests/116867 is merged.

docs/oid-info.md
wlynch: @haydentherapper - [it was pointed out](https://gitlab.com/gitlab-org/gitlab/-/issues/388517#note_1350509539) that https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md says:

> - MUST include claim to support: `Build Signer Digest` which is an immutable reference to a specific version of the build instructions that are responsible for signing.

This is a case where we don't have a good value to put here right now, though I think there's general agreement to work towards this.

Off the top of my head, I think other CI providers like Tekton and Cloud Build that let you inline build instructions would also be impacted because there isn't a specific digest to source from (though you can still fetch the build instructions via the URI).

While I agree this is important for verifying integrity, this feels like we may have over-indexed on GitHub Actions and made the barrier a little too high to start.
I'm thinking we should drop this down to `SHOULD` - wdyt?

haydentherapper: My concern would be that the Build Signer URI is mutable, effectively making the identity mutable, so there's no way to link back to the actual build instructions. With that said, if none of the other CI platforms provide this, we might have to drop MUST.

cc @laurentsimon @asraa @feelepxyz if you have more thoughts

wlynch: The Build Signer URI (in GitLab's case our plan is to use Job ID) would still be useful for asserting identity since it is considered a unique immutable identifier for the service - though it places trust on GitLab to do the right thing and limits a client's ability to detect if the CI service misbehaves.

I agree that having a digest that can be verified independently by clients is something that CI providers *should* do - but I think having this requirement as `MUST` is a high bar that not all CI providers may be able to do right away.

haydentherapper: Im not opposed to changing it to SHOULD, but I’d like to get some more input from GH, given we have discussed this as a requirement.

kommendorkapten: I think we are fine to lower this to SHOULD. Even for GitHub supporting the `Build Signer Digest` is a recent addition. By lowering this to SHOULD we still allow policy makers to REQUIRE this value, which I think is a good tradeoff.

haydentherapper: @wlynch Can you create a separate PR to update this?

wlynch: Of course! Will do.

feelepxyz (approved):
This looks great and very excited to see this! 😍 I was OOO for the last couple weeks so just coming around to this now.

wlynch (approved):
LGTM!

docs/oid-info.md
```diff
@@ -180,26 +180,26 @@ that Sigstore operates.
...
-| runner_environment | ??     | ??       | ??        | Runner Environment           | For platforms to specify whether the build took place in platform-hosted cloud infrastructure or customer-hosted infrastructure. For example: `platform-hosted` and `self-hosted`.     |
-| server_url + repository         | project_path     | ??       | ??        | Source Repository URI                   | Should include a fully qualified repository URL.                                                                                                                                       |
-| sha                | ??     | ??       | build_commit | Source Repository Digest                | An immutable reference to a specific version of the source code. Should include the digest type followed by the digest, e.g. `sha1:abc123`.                                            |
+| server_url + job_workflow_ref   | server_url + project_path + /-/jobs/ + job_id | ??       | ??        | Build Signer URI             | Reference to specific build instructions that are responsible for signing. Can be the same as Build Config URI. For example a reusable workflow in GitHub Actions or a Circle CI Orbs. |
```
haydentherapper, line 190:
```suggestion
| server_url + job_workflow_ref   | server_url + project_path + "/-/jobs/" + job_id | ??       | ??        | Build Signer URI             | Reference to specific build instructions that are responsible for signing. Can be the same as Build Config URI. For example a reusable workflow in GitHub Actions or a Circle CI Orbs. |
```

haydentherapper: To confirm, you would like to include `-`? This differs from the GHA example, not sure if that's more in line for a GitLab URI.

docs/oid-info.md
```diff
@@ -180,26 +180,26 @@ that Sigstore operates.
...
+| workflow_sha       |  pipeline_sha  | ??       | ??        | Build Config Digest          | An immutable reference to the specific version of the top-level build instructions. Should include the digest type followed by the digest, e.g. `sha1:abc123`.                         |
 | event_name         | pipeline_source | ??       | ??        | Build Trigger                | The event or action that triggered the build.                                                                                                                                          |
-| server_url + repository + "/actions/runs/" + run_id + "/attempts/" + run_attempt | ??     | ??       | ??        | Run Invocation URI           | An immutable identifier that can uniquely identify the build execution     |
+| server_url + repository + "/actions/runs/" + run_id + "/attempts/" + run_attempt | server_url + project_path + /-/pipelines/ + pipeline_id | ??       | ??        | Run Invocation URI           | An immutable identifier that can uniquely identify the build execution     |
```
haydentherapper, line 202:
```suggestion
| server_url + repository + "/actions/runs/" + run_id + "/attempts/" + run_attempt | server_url + project_path + "/-/pipelines/" + pipeline_id | ??       | ??        | Run Invocation URI           | An immutable identifier that can uniquely identify the build execution     |
```

docs/oid-info.md
```diff
@@ -180,26 +180,26 @@ that Sigstore operates.
...
 | repository_owner_id         | namespace_id     | ??       | ??        | Source Repository Owner Identifier     | Stable identifier for the owner of the source repository.                                                                                                                                   |
-| server_url + workflow_ref       | ??     | ??       | ??        | Build Config URI             | A reference to the initiating build instructions.                                                                                                                                      |
-| workflow_sha       | ??     | ??       | ??        | Build Config Digest          | An immutable reference to the specific version of the top-level build instructions. Should include the digest type followed by the digest, e.g. `sha1:abc123`.                         |
+| server_url + workflow_ref       |  pipeline_ref | ??       | ??        | Build Config URI             | A reference to the initiating build instructions.                                                                                                                                      |
```
haydentherapper, line 199: To confirm, this is fully qualified with a URL?

haydentherapper (commented):
Fantastic work on this! Just a few nits

docs/oid-info.md
feelepxyz: @aladh thinking about this a bit more. Would it make more sense to set this to the same value as `Build Config URI` or what you'll set the SAN uri to (essentially project path I believe)? For github actions, the signer is either the custom workflow in a repo or, if using reusable workflows, the ref to the reusable workflow.

Attempting to capture some kind of signing identity that you can could write policies against, trusting certain ones for example.

How does this work in GitLab? When running a pipeline step that eventually runs `npm publish --provenance`, what resource or entity ends up wrapping that step in an isolated job/run? Can it be executed by entities different to the project or is it always 1:1?

wlynch: @haydentherapper - yeah the `/-/` is part of GitLab's URI spec.

@feelepxyz some related discussion up above - https://github.com/sigstore/fulcio/pull/1097#discussion_r1158910514

I don't think we want to do that because the Build Config can point to a different repo or even a remote URL. The Job ID we're including here is still useful information to encode in the cert to differentiate jobs in a pipeline, since that's what's doing the signing. However, the Job ID is far to granular to be a useful SAN since it relies on a UID.

I think ideally we would use a named CI config + named job, but that's not a concept that exists in GitLab today (though @marshall007 mentioned that there are discussions happening that may lead to this in the future).

Today there's only 1 CI config per repo, which is why for the SAN we're planning to use is the repo+ref (which also covers merge requests from forks since the identity in that case would be the forked repo).
When GitLab supports multiple CI configs per repo, we can iterate on this and take that into account.

haydentherapper: If there's plans to support multiple CI configs per repo in the near future, should we handle that now? Otherwise unless we proactively account for that, there will be a period where there are multiple configs and the SAN remains the same as now

wlynch: @marshall007 can confirm, but I don't think there's a solid timeline / commitment at the moment, so I wouldn't design around it for now.

docs/oid-info.md
wlynch: Yes. See https://gitlab.com/gitlab-org/gitlab/-/issues/404722 for more details.

haydentherapper (approved):
Going to approve  - There's just two open nits to change to ", but that's very small. Any more comments @wlynch @priyawadhwa @aladh?

Thanks so much for working on this!!

priyawadhwa:
This LGTM, I'll go ahead and merge. @aladh thank you for your work on this! would you mind addressing the nits in a follow-up PR? Thank you!

PR #1126, opened 2023-04-14
oid-info: Drop Build Signer Digest requirement from MUST -> SHOULD

wlynch:
#### Summary

It was pointed out that CI providers may not be able to meet the Build Signer Digest requirement if they've been using server-assigned UUIDs to identify CI jobs rather than a user-verifiable digest.

Everyone seems to be in agreement that this is something CI providers SHOULD strive for so that users have a means of detecting misbehavior by the CI system, but this is not necessarily something we may want to gate acceptance on, since it would require work to implement and include.

As a result, dropping this down from MUST to SHOULD.

/cc @haydentherapper @kommendorkapten @feelepxyz @asraa @laurentsimon @aladh @marshall007 for any additional feedback

#### Release Note

Dropped `Build Signer Digest` OID requirement from `MUST` to `SHOULD`.

#### Documentation

✅

### DIFF
--- a/docs/oid-info.md
+++ b/docs/oid-info.md
@@ -118,7 +118,7 @@ is formatted to the RFC 5280 specification as a DER-encoded string.
 
 ### 1.3.6.1.4.1.57264.1.9 | Build Signer URI
 
-Reference to specific build instructions that are responsible for signing. SHOULD be fully qualified. MAY be the same as Build Config URI. Build Signer URI is also included in the Subject Alternative Name.
+Reference to specific build instructions that are responsible for signing. SHOULD be fully qualified. MAY be the same as Build Config URI.
 
 For example a reusable workflow ref in GitHub Actions or a Circle CI Orb name/version. For example: `https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_container_slsa3.yml@v1.4.0`.
 
--- a/docs/oid-info.md
+++ b/docs/oid-info.md
@@ -18,12 +18,12 @@ Requirements:
 
 - MUST include `iss` claim for `Issuer` extension.
 - MUST include claim to support: `Build Signer URI` that identifies the specific build instructions that are responsible for signing.
-- MUST include claim to support: `Build Signer Digest` which is an immutable reference to a specific version of the build instructions that are responsible for signing.
 - MUST include claim to support: `Runner Environment` that differentiates between builds that took place in platform-hosted cloud infrastructure or customer-hosted infrastructure.
 
 Recommended:
 
 - SHOULD include `iss` that uniquely identifies ID tokens originating from the CI/CD system, e.g. not shared with OIDC OAuth 2.0 tokens for email/username logins.
+- SHOULD include claim to support: `Build Signer Digest` which is an immutable reference to a specific version of the build instructions that are responsible for signing.
 - SHOULD include claim to support: `Source Repository URI`
 - SHOULD include claim to support: `Source Repository Digest`
 - SHOULD include claim to support: `Source Repository Ref`


priyawadhwa (dismissed):

haydentherapper (approved):
Thanks! 
Will leave open for a day to see if there's any other feedback

laurentsimon:
> Build Signer Digest requirement if they've been using server-assigned UUIDs to identify CI runs rather than a user-verifiable digest

a run is different from a signer. The "Build Signer Digest", iiuc, identifier the builder, not the run. GHA also use some sort of UUID to identify runs (RUN_ID, RUN_ATTEMPT, etc). If the code requesting an OIDC token is in source code repo, then there *is* a digest associated with it. 

I'm certainly mis-understanding something :-)

wlynch:
> Build Signer Digest requirement if they've been using server-assigned UUIDs to identify CI runs rather than a user-verifiable digest
> 
> a run is different from a signer. The "Build Signer Digest", iiuc, identifies the builder, not the run. GHA also uses some sort of UUID to identify runs (RUN_ID, RUN_ATTEMPT, etc). If the code requesting an OIDC token is in source code repo, then there _is_ a digest associated with it.
> 
> I'm certainly mis-understanding something :-)

You're right- this should be s/runs/jobs (I'll update this).

We anticipate this looking like:

Field | GitHub | GitLab
-- | -- | --
Build Signer URI | Job Workflow Ref | Job ID
Build Signer Digest | Job Workflow SHA | - (there is no current equivalent)
Build Config URI | Workflow Ref | Pipeline Ref
Build Config SHA | Workflow SHA | Pipeline SHA
Run Invocation URI | Run ID + Run Attempt | Pipeline ID

kommendorkapten (approved):

kommendorkapten:
Just one more question, the `Job ID` is stable across multiple invocations, right? It's just that when a new job is created, it gets an UUID?

feelepxyz (dismissed):
SGTM 👍

feelepxyz:
@wlynch for GitLab, what are you adding in the Subject Alternative Name URI? For GitHub Actions, `Builder Signer URI` and Subject Alternative Name URI are set to the same value. It's also expected this is what you want to write policies against when verifying trusted cert identities. 

Can you write stable-ish policies against the Job ID or does it change between each job?

wlynch:
> @wlynch for GitLab, what are you adding in the Subject Alternative Name URI? For GitHub Actions, `Builder Signer URI` and Subject Alternative Name URI are set to the same value. It's also expected this is what you want to write policies against when verifying trusted cert identities.
> 
> Can you write stable-ish policies against the Job ID or does it change between each job?

It's an opaque id that changes per job. There's no way today to reliably ID a job with a known name because of the flexibility in how jobs can be referenced/layered.

The plan for the SAN is to use `https://gitlab.com/org/repo@ref`, since as of now you can only have a single config per repo. AFAIK Builder Signer URI == SAN is not a strict requirement, though I think it makes sense to always use the most-specific stable identifier.

wlynch:
> Can you also update the text of 1.9 then, since it includes `Build Signer URI is also included in the Subject Alternative Name`?

Done.

wlynch:
> Just one more question, the Job ID is stable across multiple invocations, right? It's just that when a new job is created, it gets an UUID?

Yes. A job retry results in a new job with a new UID.

docs/oid-info.md
```diff
@@ -118,7 +118,7 @@ is formatted to the RFC 5280 specification as a DER-encoded string.
 
 ### 1.3.6.1.4.1.57264.1.9 | Build Signer URI
 
-Reference to specific build instructions that are responsible for signing. SHOULD be fully qualified. MAY be the same as Build Config URI. Build Signer URI is also included in the Subject Alternative Name.
```
haydentherapper, line 121: What do you think about including a line somewhere (maybe under the MUST section) on suggestions for what the SAN should be, e.g. "The Subject Alternative Name should be a stable identifier", something like that?

wlynch: Yeah I was thinking about this. This probably deserves its own section with good/bad examples. I'll tackle this in another PR, since I suspect this might be a bit lengthy.

haydentherapper: SG, thanks!

haydentherapper (approved):

priyawadhwa:
@haydentherapper, seems like this should be good to merge?

PR #1152, opened 2023-05-04
Add support to specify hash function for KMS CA

codysoyland:
If using KMS providers other than GCP, the hash function is hardcoded to SHA256. (GCP provider looks up the keyversion and applies the correct hash function, other providers do not). This patch allows the user to configure the hash function with a CLI flag.

### DIFF
--- a/pkg/ca/kmsca/kmsca_test.go
+++ b/pkg/ca/kmsca/kmsca_test.go
@@ -17,6 +17,7 @@ package kmsca
 
 import (
 	"context"
+	"crypto"
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rand"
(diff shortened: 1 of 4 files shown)


pkg/ca/kmsca/kmsca_test.go
```diff
@@ -36,7 +37,7 @@ func TestNewKMSCA(t *testing.T) {
 
 	chain := []*x509.Certificate{subCert, rootCert}
 
-	ca, err := NewKMSCA(context.WithValue(context.TODO(), fake.KmsCtxKey{}, subKey), "fakekms://key", chain)
+	ca, err := NewKMSCA(context.WithValue(context.TODO(), fake.KmsCtxKey{}, subKey), "fakekms://key", chain, crypto.SHA256)
```
codysoyland, line 40: I would like to add a test for this, but that would require further code changes.

haydentherapper:
sigstore/sigstore#1149 makes this mostly unnecessary. The only use case is hashivault, which uses the hash function passed to it (AWS, Azure, and GCP fetch it automatically, based on the key parameters). Supporting atypical pairings like RSA4096 + SHA256 is also something to consider, but for Azure and AWS, this isn't supported, and for GCP, there's a hardcoded list of pairings.

I'd propose updating the flag name to include "hcvault" or something like that, and passing `nil` when the flag isn't set to make it clear that the hash function is handled via the library in most cases.

haydentherapper:
Closing as there isn't any demand for hc currently, and we can maybe rewrite the HC client to automatically detect the hash algorithm

PR #1183, opened 2023-05-24
gitlab: remove build config URI.

wlynch:
#### Summary

This was erroneously set to match the build signer URI. This will eventually be pipeline_ref when it is populated (dependent on https://gitlab.com/gitlab-org/gitlab/-/merge_requests/121597)

#### Release Note

BREAKING CHANGE: GitLab Build Config URI claim is temporarily removed. This was erroneously set to an incorrect value that did not match the documentation. This will repopulated once the correct value is available in upstream claims.


wlynch:
cc @cpanato @haydentherapper

priyawadhwa (approved):

PR #1256, opened 2023-07-02
add fsnotify-backed cache for reading TLS PKI material

bobcallaway:
#### Summary
If the PKI material used to expose a TLS endpoint changes while an instance is running, the instance must be restarted to read the new material. This change adds an fsnotify-based watcher for the certificate, and caches the value of the cert in process memory to save file I/O during the establishment of each TLS connection.

This scenario is likely in Kubernetes where PKI material may be refreshed by a renewal process and exposed to Fulcio through a volume mount.

### DIFF
--- a/cmd/app/grpc.go
+++ b/cmd/app/grpc.go
@@ -22,7 +22,9 @@ import (
 	"net"
 	"os"
 	"runtime"
+	"sync"
 
+	"github.com/fsnotify/fsnotify"
 	"github.com/goadesign/goa/grpc/middleware"
 	ctclient "github.com/google/certificate-transparency-go/client"
 	grpcmw "github.com/grpc-ecosystem/go-grpc-middleware"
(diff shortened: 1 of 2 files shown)


cmd/app/grpc.go
```diff
@@ -67,16 +70,85 @@ func PassFulcioConfigThruContext(cfg *config.FulcioConfig) grpc.UnaryServerInter
...
+	// Add a path.
+	if err = cachedTLSCert.Watcher.Add(certPath); err != nil {
+		return nil, err
+	}
+	return cachedTLSCert, nil
+}
+
+func (c *cachedTLSCert) GetCertificate() *tls.Certificate {
+	// get reader lock
+	c.RLock()
```
haydentherapper, line 126: Do you expect this would have negligible performance impact? I would assume it's fine given it's a reader lock, but it is introducing a lock for every call.

bobcallaway:
Yes, negligible. Getting a reader lock is simply an atomic check and
increment an int, and the writer lock should only block readers temporarily
for the file I/O to read the new cert from disk and parse it. Even if it’s
rotated every 24h that’s extremely minimal.

cmd/app/grpc.go
bobcallaway: Yes, negligible. Getting a reader lock is simply an atomic check and
increment an int, and the writer lock should only block readers temporarily
for the file I/O to read the new cert from disk and parse it. Even if it’s
rotated every 24h that’s extremely minimal.

haydentherapper (approved):

PR #1271, opened 2023-07-11
Expose SkipExpiryCheck OIDC Config Option in Verifier

priyawadhwa:
This PR exposes some oidc config options allowed by the go-oidc library that we use, specifically [SkipExpiryCheck](https://github.com/coreos/go-oidc/blob/b203e58c24394ddf5e816706a7645f01280245c7/oidc/verify.go#L96). 

It updates the LRU cache for verifiers to include both the verifier and the config in the cache, so that Verifier A with a normal config can be distinguished from Verifier A with a config with SkipExpiryCheck set

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -24,6 +24,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"reflect"
 	"regexp"
 	"strings"
 	"time"
@@ -41,6 +42,11 @@ const defaultOIDCDiscoveryTimeout = 10 * time.Second
 // top-level and second-level domain
 const minimumHostnameLength = 2
 
+type verifierWithConfig struct {
+	*oidc.IDTokenVerifier
+	*oidc.Config
+}
+
 type FulcioConfig struct {
 	OIDCIssuers map[string]OIDCIssuer `json:"OIDCIssuers,omitempty"`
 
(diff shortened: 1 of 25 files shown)


pkg/config/config.go
```diff
@@ -129,25 +135,38 @@ func (fc *FulcioConfig) GetIssuer(issuerURL string) (OIDCIssuer, bool) {
...
+	}
+	cfg := &oidc.Config{ClientID: iss.ClientID}
+	for _, o := range opts {
+		o(cfg)
+	}
 	// Look up our fixed issuer verifiers
 	v, ok := fc.verifiers[issuerURL]
 	if ok {
-		return v, true
+		for _, c := range v {
```
haydentherapper, line 150: Is this a use case that needs to be handled? What's the case when you'd have a configuration that specifies both skipping the expiration and not skipping the expiration?

pkg/identity/issuer.go
```diff
@@ -14,13 +14,17 @@
...
+)
 
 type Issuer interface {
 	// Match checks if this issuer can authenticate tokens from a given issuer URL
 	Match(ctx context.Context, url string) bool
 
 	// Authenticate ID token and return Principal on success. The ID token's signature
 	// is verified in the call -- invalid signature must result in an error.
-	Authenticate(ctx context.Context, token string) (Principal, error)
+	Authenticate(ctx context.Context, token string, opts ...config.InsecureOIDCConfigOption) (Principal, error)
```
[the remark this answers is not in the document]
priyawadhwa: Yes!

pkg/config/config.go
priyawadhwa: Yah, so our custom issuer does have this use case. Not sure if I can give more details though 😅

haydentherapper (approved):

Issue #1182, opened 2023-05-24
Generalize Fulcio OIDs, clarify guidance for providers


marshall007:
**Description**

All build systems have some concept of a "pipeline", which is typically composed of multiple "jobs". Some providers support reusable/identifiable "job" definitions or even composing reusable "steps" together to form "jobs", but this is a less common/consistent abstraction.

Most (probably all) build systems can easily generate URI references to "pipeline"-level invocations and their source config, but "job"-level metadata is highly dependent on how "pipeline" configuration is defined. This is challenging to do if pipeline configurations can be composed dynamically at runtime, for example.

Fulcio OIDs are loosely based on the assumption that providers can reliably produce immutable references to "job"-level configuration. When that is not possible, the guidance is not very clear. I think we should consider standardizing OIDs that map more cleanly to both "pipeline" and "job" abstractions to:

1. improve the consistency of how these claims are interpreted across providers
1. make incremental adoption easier for providers that cannot easily generate immutable "job"-level refs

I propose something along the lines of:

- `Run Invocation URI` (required): uniquely identifies pipeline/workflow execution (typically a human-readable summary page)
- `Run Config URI/Digest` (required): actual workflow/pipeline configuration ref
- `Run Trigger`: workflow/pipeline trigger/dispatch event
- `Build Invocation URI`: uniquely identifies job execution (typically a human-readable summary page)
- `Build Config URI/Digest`: actual build/job configuration ref
- ~~`Build Signer URI/Digest`~~: no longer necessary. the same as `Run/Build Config`. `Build` is preferred, but `Run` is acceptable if that's all the provider can support

Note the current intent behind `Run Invocation URI` is to uniquely identify the run, in practice it is typically represented by a "summary page" for human users to view directly. I think we should embrace this `Invocation URI` terminology and definition as this ends up being quite useful for UI implementations (like "view build summary" links in the npm registry UI).

/cc @wlynch

haydentherapper:
cc @trevrosen @feelepxyz @laurentsimon who might also be interested

feelepxyz:
@marshall007 thanks for writing this up!

> Fulcio OIDs are loosely based on the assumption that providers can reliably produce immutable references to "job"-level configuration

This was not my intention when writing the initial proposal on these extensions and we should look to clarify this in the docs! I agree that we should primarily reference the "pipeline"/workflow-level/ invocations and their source config.

[...]

PR #1307, opened 2023-08-06
Update Buildkite issuer to include some of the new certificate extensions

### LINKED ISSUE #1074: Update Buildkite issuer to include new extensions
**Description**

Following #754, we should update the Buildkite integration to include additional extensions. We should leverage https://github.com/sigstore/fulcio/pull/1073 and refactor to avoid duplication.

yob:
#### Summary

The Buildkite Issuer was added in #890, prior to the efforts to standardise certificate extensions for CI providers, and #1074 calls for the Buildkite issuer to be updated to use the new extensions (where applicable).

This is an early attempt to make those changes.

I've added the extensions that make the most sense in a Buildkite context, like RunInvocationURI, RunnerEnvironment and SourceRepositoryDiget. Many of the other extensions don't apply because we're not a code host as well, or need further discussion.

I have not added tests yet. This is my first contribution to fulcio and I'm keen to confirm I'm heading in the right direction before adding tests. However, I have tested this locally with a Buildkite agent and OIDC token, and the certificate was issued as expected.

Using `git cat-file commit HEAD` and piping it through `openssl pkcs7 -print -print_certs -text`, the extensions section looks like this:

    X509v3 extensions:
        X509v3 Key Usage: critical
            Digital Signature
        X509v3 Extended Key Usage:
            Code Signing
            19:9E:E7:53:4D:F6:65:D4:23:9D:60:21:B8:F3:12:80:FD:11:30:7F
            8A:3E:9E:34:19:F7:32:18:3D:2A:1B:F9:5F:60:29:24:0F:70:0B:B4
            URI:https://buildkite.com/yob-opensource/oidc-signing-experiment
        1.3.6.1.4.1.57264.1.1:
            https://agent.buildkite.com
        1.3.6.1.4.1.57264.1.8:
            ..https://agent.buildkite.com
        1.3.6.1.4.1.57264.1.11:
            ..self-hosted
        1.3.6.1.4.1.57264.1.13:
            .(5242de9e5c2ca164cb3dfc500fb605f0b8b86763
        1.3.6.1.4.1.57264.1.21:
            .mhttps://buildkite.com/yob-opensource/oidc-signing-experiment/builds/35%230189cb29-62fa-41af-8641-62e1d6c5edfd

Fixes #1074

/cc @sj26 

#### Release Note

NONE

#### Documentation

Uncertain


pkg/identity/buildkite/principal.go
```diff
@@ -69,16 +111,23 @@ func (p jobPrincipal) Name(_ context.Context) string {
...
+	jobUrl := baseURL.JoinPath(p.organizationSlug, p.pipelineSlug, "builds", strconv.FormatInt(p.buildNumber, 10)+"#"+p.jobId)
+
+	// Set SubjectAlternativeName to the pipeline URL on the certificate
+	cert.URIs = []*url.URL{pipelineUrl}
 
 	// Embed additional information into custom extensions
 	cert.ExtraExtensions, err = certificate.Extensions{
-		Issuer: p.issuer,
+		Issuer:                 p.issuer,
+		RunInvocationURI:       jobUrl.String(),
```
yob, line 128: This is the one I'm most interested in - having the specific job URL in the certificate to provide provenance would be excellent

haydentherapper:
cc @feelepxyz @marshall007 @wlynch for more thoughts

Thank you for starting this!

pkg/identity/buildkite/principal.go
marshall007: @yob I think it makes sense, we went job-specific for the GitLab provider: https://github.com/sigstore/fulcio/blob/04e4ac9b2d867d0f94e58a858f373a831d87653d/pkg/identity/gitlabcom/principal.go#L233C63-L233C63

Some related discussion in https://github.com/sigstore/fulcio/issues/1182. I found it confusing that this extension was named `RunInvocationURI` (instead of `BuildInvocationURI`), sort of implying it ought to be pipeline-level.

GitHub [does not have `job_id`](https://docs.github.com/en/enterprise-cloud@latest/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#understanding-the-oidc-token) available in their ID tokens, so they don't have the option to align with us on that presently.

pkg/identity/buildkite/principal.go
```diff
@@ -69,16 +111,23 @@ func (p jobPrincipal) Name(_ context.Context) string {
...
+		RunnerEnvironment:      p.runnerEnvironment,
+		SourceRepositoryDigest: p.buildCommit,
```
feelepxyz, line 130: I would love to see support for a few additional extensions so we could get richer provenance info:
- Would it not be possible to also forward the repo url and ref from the originating source host to populate `SourceRepositoryURI` and `SourceRepositoryRef`? What about the repo and owner IDs? 
- Could you point the `BuildConfigURI` to [`pipeline.yml`](https://buildkite.com/docs/pipelines/defining-steps) file that configures the build? Is there something else that configures the build? Is the sha of this always the same as `p.buildCommit`?
- Is there not some concept of the event trigger that kicks of the job in buildkite that could be used to populate `BuildTrigger`?

pkg/identity/buildkite/principal.go
```diff
@@ -69,16 +111,23 @@ func (p jobPrincipal) Name(_ context.Context) string {
...
 	if err != nil {
 		return err
 	}
-	cert.URIs = []*url.URL{parsed}
+
+	pipelineUrl := baseURL.JoinPath(p.organizationSlug, p.pipelineSlug)
+	jobUrl := baseURL.JoinPath(p.organizationSlug, p.pipelineSlug, "builds", strconv.FormatInt(p.buildNumber, 10)+"#"+p.jobId)
+
+	// Set SubjectAlternativeName to the pipeline URL on the certificate
+	cert.URIs = []*url.URL{pipelineUrl}
```
feelepxyz, line 123: Would this support a use case to write a verification policy to only allow builds from a particular buildkite project/pipeline/thing. Looks like this identifier would be stable across jobs but unique to the parent resource responsible for running the job. Could you have different build's from different repos running with the same org and pipeline slug?

I would also use this value to populate `BuildSignerURI` as these are currently mirrored for github and gitlab. If a pipeline slug is always configured from a pipeline.yml and we know the sha of the pipleline file we could also use this to populate `BuildConfigDigest` and `BuildSignerDigest`.

haydentherapper:
Hey all, just wanted to check in, has there been any progress on this?

pkg/identity/buildkite/principal.go
sj26: Buildkite has no access to the source repository.

Because of this, `SourceRepositoryURI` and `SourceRepositoryRef` are resolved within the runner environment, and are not always known on the Buildkite side which generates the OIDC tokens. We have a `repository` url which is provided as an input to the runner, but it is not always a fully-qualified URI. We have a build `branch` and `commit`, but these often start as values like `HEAD` and are resolved to a concrete `ref` by the runner which sends back the resolve commit sha.

Similarly for `BuildConfigURI`. The initial steps are defined as part of the pipeline on Buildkite. But those initial steps are usually just a single "upload" step. That step runs within the runner environment, to access the source repository. It conventionally looks at `buildkite.yml` or `.buildkite/pipeline.yml`, but can choose any file in the source repository. It may not even be a file, the runner may generate content, like with a script, which is piped into an upload command to use as build config. And the build is not limited to one upload - many files may be uploaded - and in fact the build config may be piped in and not sourced from an actual file in the repository.

`BuildTrigger` is probably something we could add. But I'd lean towards another PR for that.

sj26: (I'd love to see the contents of this PR land as an increment, and consider additional extensions later.)

richardfan1126:
I believe this PR is obsolete due to the change introduced in https://github.com/sigstore/fulcio/pull/1729

Instead, we need to update https://github.com/sigstore/fulcio/blob/523797928237760a828cc863768ace5ae9d22026/config/identity/config.yaml#L230-L233

haydentherapper:
I agree no more work is needed on the PR - If someone could convert the changes over to configuration in the above linked file, that would be appreciated!

yob:
> I agree no more work is needed on the PR - If someone could convert the changes over to configuration in the above linked file, that would be appreciated!

I'm working on this at the moment, so I'll close this PR and open a new ones hortly

Issue #1341, opened 2023-08-31
Add Shutdown behavior


haydentherapper:
**Description**

Currently, the server does not listen for an interrupt, so a server restart may terminate ongoing connections. Golang HTTP servers provide a shutdown method to gracefully terminate, which we are currently not calling.

See https://pkg.go.dev/net/http#Server.Shutdown for an example

PR #1342, opened 2023-08-31
Gracefully shutdown HTTP, gRPC, and Prom servers

haydentherapper:
This catches sigint and sigterm and calls each server's shutdown method, blocking until shutdown completes.

This is only added for non-Duplex, we'll need to add a shutdown method for Duplex servers.

Fixes https://github.com/sigstore/fulcio/issues/1341

### DIFF
--- a/cmd/app/http.go
+++ b/cmd/app/http.go
@@ -21,8 +21,12 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"os"
+	"os/signal"
 	"strconv"
 	"strings"
+	"sync"
+	"syscall"
 	"time"
 
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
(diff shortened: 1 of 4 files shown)


haydentherapper:
Tested with `go run main.go serve --port 5555 --grpc-port 5554 --ca ephemeralca --ct-log-url=""` and confirmed we saw all expected logging statements.

I tried to get this working for Duplex by calling `d.Server.GracefulStop()`, but the process hung. I assume I need to also shutdown the metrics server, which isn't exposed to the caller.

cpanato (dismissed):
Cool

cmd/app/http.go
```diff
@@ -102,12 +106,32 @@ func createHTTPServer(ctx context.Context, serverEndpoint string, grpcServer, le
...
+			// error from closing listeners, or context timeout
+			log.Logger.Errorf("HTTP server Shutdown: %v", err)
+		}
+		close(idleConnsClosed)
+		log.Logger.Info("stopped http server")
+	}()
+
+	wg.Add(1)
 	go func() {
 		if err := h.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
```
bobcallaway, line 129: If the server fails to start immediately for some reason (e.g. the port is already in use), I think you'll be blocked waiting on `idleConnsClosed` to be written to but it won't be because the signal handler is still waiting (requiring intervention to stop). 

we probably should consider changing this to say if `err != nil && !errors.Is(err, http.ErrServerClosed)`, then `Fatal` log the error, otherwise just `Error` log the error and continue.

cmd/app/serve.go
```diff
@@ -304,15 +311,37 @@ func runServeCmd(cmd *cobra.Command, args []string) { //nolint: revive
...
+
+		// received an interrupt signal, shut down
+		if err := prom.Shutdown(context.Background()); err != nil {
+			// error from closing listeners, or context timeout
+			log.Logger.Errorf("HTTP server Shutdown: %v", err)
+		}
+		close(idleConnsClosed)
+		log.Logger.Info("stopped prom server")
+	}()
+	if err := prom.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
```
bobcallaway, line 337: same comment as above, i think if the metrics port is already in use this will hang

bobcallaway (commented):
if ports are already in use (e.g. `tcp :2112: bind: address already in use`), the server will hang instead of quickly failing and exiting

cmd/app/http.go
haydentherapper: Good call, I confirmed this hangs with just Error. We only need to log for fatal, err would only be for logging server closed.

cmd/app/serve.go
haydentherapper: done

haydentherapper:
@bobcallaway Fixed comments, also made the unix domain socket configurable so that you can start up two fulcios on the same machine

bobcallaway (approved):

PR #1368, opened 2023-09-14
Disable CGO on release builds

loosebazooka:
Update to base-debian12

Open questions?
- [ ] Does fulcio still actually need cgo or libssl, could we just use `base-nossl...` instead?
- [ ] Could we actually just run in `non-root` mode?

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -1,12 +1,11 @@
-defaultBaseImage: gcr.io/distroless/base-debian10
+defaultBaseImage: gcr.io/distroless/static-debian12
 builds:
 - main: .
   env:
-  - CGO_ENABLED=1
-# If you are deploying from M1, you can use this (uncomment below, and
-# comment out above), though it does remove the support for the "createca" command.
-# But at least you can deploy it from M1 using this.
-#  - CGO_ENABLED=0
+  - CGO_ENABLED=0
+# If you need support for the the "createca" command, you must enable
+# CGO and use a base image with gblic (base instead of static)
+#  - CGO_ENABLED=1
   flags:
   - -trimpath
   - -tags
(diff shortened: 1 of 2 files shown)


loosebazooka:
cc: @jku

.ko.yaml
```diff
@@ -1,4 +1,4 @@
-defaultBaseImage: gcr.io/distroless/base-debian10
+defaultBaseImage: gcr.io/distroless/base-debian12
 builds:
 - main: .
   env:
```
haydentherapper, line 4: @loosebazooka If we change the default to be CGO_ENABLED=0, can we switch to a different image? CGO is only needed for the pkcs11 library, which we don't use in the production CA.

haydentherapper:
I don't know any reason we need root. @bobcallaway @cpanato Are you aware of any reason?

.ko.yaml
cpanato: maybe we can change that to 0 and run on ci, last time i think that was not building properly

bobcallaway (dismissed):
no, i can't think of a reason we need CGO or root

.ko.yaml
loosebazooka: kk testing it out. also switched from base -> static

haydentherapper: Looks good!

haydentherapper:
Can you also change it in https://github.com/sigstore/fulcio/blob/dd9a1050eb7a7b885ffb0a661eabdf32931b4a1a/.goreleaser.yml#L5?

loosebazooka:
@cpanato what's the testing strategy for this change? should I swap to the `nonroot` tag

haydentherapper:
I've switched to `nonroot`. I tested locally with `ko build .`, pulled the generated container, and verified I could start and query the CA.

bobcallaway (approved):

cpanato:
sorry, i am a bit busy with company work 

this looks good to me as well

cpanato (approved):

Issue #1453, opened 2023-11-07
Fulcio doesn't pass http customization to go-oidc


ccordoui:
**Description**

`go-oidc` use http `DefaultClient` and it does not take into account Fulcio custom config of `http.DefaultTransport`.

In my case, I have an `oidc` provider endpoint signed by the CA in `/var/run/fulcio/ca.crt` in kubernetes and it fails with an error: `x509: certificate signed by unknown authority`.

I suppose that if we needed to go through a proxy we would have the same issue.

**Version**

Fulcio latest (main branch)

PR #1454, opened 2023-11-07
Move kubernetes CA handling in config.prepare

ccordoui:
#### Summary

Resolves sigstore/fulcio#1453

#### Release Note

- Fix a bug where fulcio wasn't passing http custom configuration to `go-oidc`

#### Documentation

NONE


pkg/config/config.go
```diff
@@ -168,8 +168,14 @@ func (fc *FulcioConfig) GetVerifier(issuerURL string, opts ...InsecureOIDCConfig
...
 	// and add it to the LRU cache.
 
-	ctx, cancel := context.WithTimeout(context.Background(), defaultOIDCDiscoveryTimeout)
+	parentContext, cancel := context.WithTimeout(context.Background(), defaultOIDCDiscoveryTimeout)
 	defer cancel()
+
+	// go-oidc use DefaultClient and it does not take into account our custom config of http.DefaultTransport
+	// ClientContext create a new Context that carries our http client.
+
+	oidcClient := &http.Client{}
```
haydentherapper, line 177: I might be misunderstanding, but isn't this equivalent to `http.DefaultClient`?

bobcallaway: https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/net/http/client.go;drc=97df9c0b051b6c06bff7ec7e61522e7bbef40c91;l=30
```
// A Client is an HTTP client. Its zero value (DefaultClient) is a
// usable client that uses DefaultTransport.
```

ccordoui: You are both correct, my problem was elsewhere, I backported the previous patch to my 1.4.0 code and I thought it solved issue but I also imported this change:

```go
diff --git a/pkg/config/config.go b/pkg/config/config.go
index 45f0c17..9068bd0 100644
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -229,10 +229,11 @@ func (fc *FulcioConfig) prepare() error {
                defer cancel()
                provider, err := oidc.NewProvider(ctx, iss.IssuerURL)
                if err != nil {
-                       return fmt.Errorf("provider %s: %w", iss.IssuerURL, err)
+                       log.Logger.Errorf("error creating provider for issuer URL %q: %v", iss.IssuerURL, err)
+               } else {
+                       cfg := &oidc.Config{ClientID: iss.ClientID}
+                       fc.verifiers[iss.IssuerURL] = []*verifierWithConfig{{provider.Verifier(cfg), cfg}}
                }
-               cfg := &oidc.Config{ClientID: iss.ClientID}
-               fc.verifiers[iss.IssuerURL] = []*verifierWithConfig{{provider.Verifier(cfg), cfg}}
        }

        cache, err := lru.New2Q(100 /* size */)
```

Which made the pod running even though no valid OIDC was present.

ccordoui:
`config.prepare` is called two times:

- [config.go#L459](https://github.com/sigstore/fulcio/blob/49eee1e9afecfe11b1f42091169873819869a10a/pkg/config/config.go#L459)
- [config.go#L409](https://github.com/sigstore/fulcio/blob/49eee1e9afecfe11b1f42091169873819869a10a/pkg/config/config.go#L409)

The first time was after the import of the CA of kubernetes [config.go#L433](https://github.com/sigstore/fulcio/blob/49eee1e9afecfe11b1f42091169873819869a10a/pkg/config/config.go#L433) but the second was not.

So to prevent that kind of behavior, I have moved the code in the prepare, like this it is always called before.

haydentherapper (approved):
@ccordoui Have you been able to compile Fulcio locally and confirm this is working for you?

ccordoui:
> @ccordoui Have you been able to compile Fulcio locally and confirm this is working for you?

@haydentherapper Yes, the built image can be found here `quay.io/ccordoui/sigstore/fulcio:ccordoui-pr1454`

PR #1494, opened 2023-12-13
docs: Add back descriptive content on cert issuing

cmurphy:
As part of doc enhancements in ae4c766f, the description in step 7 was removed. Although the final step is simple and the diagram and section header are clear, it is jarring for the final step to be inconsistent with the others by having no text body. It is moreover unfriendly to screen readers since there is no context for the diagram. This change adds the brief description back for the sake of readability.

#### Summary

Enhance the documentation readability by having a consistent flow from step to step.

#### Release Note

NONE

#### Documentation

n/a


haydentherapper (approved):
Sorry I missed this, thanks!

PR #1517, opened 2024-01-08
Allow configurable client signing algorithms

tetsuo-cpp:
#### Summary

This PR adds a `--client-signing-algorithms` flag to Fulcio to restrict what key/hash combinations are allowed.

### DIFF
--- a/cmd/app/serve.go
+++ b/cmd/app/serve.go
@@ -33,6 +33,9 @@ import (
 	"syscall"
 	"time"
 
+	v1 "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
+	"github.com/sigstore/sigstore/pkg/signature"
+
 	"chainguard.dev/go-grpc-kit/pkg/duplex"
 	"github.com/goadesign/goa/grpc/middleware"
 	ctclient "github.com/google/certificate-transparency-go/client"
@@ -113,6 +116,12 @@ func newServeCmd() *cobra.Command {
 	cmd.Flags().String("grpc-tls-key", "", "the private key file to use for secure connections (without passphrase) - only applies to grpc-port")
 	cmd.Flags().Duration("idle-connection-timeout", 30*time.Second, "The time allowed for connections (HTTP or gRPC) to go idle before being closed by the server")
 	cmd.Flags().String("ct-log.tls-ca-cert", "", "Path to TLS CA certificate used to connect to ct-log")
+	cmd.Flags().StringSlice("client-signing-algorithms", buildDefaultClientSigningAlgorithms([]v1.PublicKeyDetails{
+		v1.PublicKeyDetails_PKIX_ECDSA_P256_SHA_256,
+		v1.PublicKeyDetails_PKIX_ECDSA_P384_SHA_384,
+		v1.PublicKeyDetails_PKIX_ECDSA_P521_SHA_512,
+		v1.PublicKeyDetails_PKIX_ED25519,
+	}), "the list of allowed client signing algorithms")
 
 	// convert "http-host" flag to "host" and "http-port" flag to be "port"
 	cmd.Flags().SetNormalizeFunc(func(_ *pflag.FlagSet, name string) pflag.NormalizedName {
@@ -210,6 +219,20 @@ func runServeCmd(cmd *cobra.Command, args []string) { //nolint: revive
 
 	// Setup the logger to dev/prod
 	log.ConfigureLogger(viper.GetString("log_type"))
+
+	algorithmStrings := viper.GetStringSlice("client-signing-algorithms")
+	var algorithmConfig []v1.PublicKeyDetails
+	for _, s := range algorithmStrings {
+		algorithmValue, err := signature.ParseSignatureAlgorithmFlag(s)
+		if err != nil {
+			log.Logger.Fatal(err)
+		}
+		algorithmConfig = append(algorithmConfig, algorithmValue)
+	}
+	algorithmRegistry, err := signature.NewAlgorithmRegistryConfig(algorithmConfig)
+	if err != nil {
+		log.Logger.Fatalf("error loading --client-signing-algorithms=%s: %v", algorithmConfig, err)
+	}
 
 	// from https://github.com/golang/glog/commit/fca8c8854093a154ff1eb580aae10276ad6b1b5f
 	_ = flag.CommandLine.Parse([]string{})
@@ -324,7 +347,7 @@ func runServeCmd(cmd *cobra.Command, args []string) { //nolint: revive
 		port := viper.GetInt("port")
 		metricsPort := viper.GetInt("metrics-port")
 		// StartDuplexServer will always return an error, log fatally if it's non-nil
-		if err := StartDuplexServer(ctx, cfg, ctClient, baseca, viper.GetString("host"), port, metricsPort, ip); err != http.ErrServerClosed {
+		if err := StartDuplexServer(ctx, cfg, ctClient, baseca, algorithmRegistry, viper.GetString("host"), port, metricsPort, ip); err != http.ErrServerClosed {
 			log.Logger.Fatal(err)
 		}
 		return
@@ -337,7 +360,7 @@ func runServeCmd(cmd *cobra.Command, args []string) { //nolint: revive
 
 	reg := prometheus.NewRegistry()
 
-	grpcServer, err := createGRPCServer(cfg, ctClient, baseca, ip)
+	grpcServer, err := createGRPCServer(cfg, ctClient, baseca, algorithmRegistry, ip)
 	if err != nil {
 		log.Logger.Fatal(err)
 	}
@@ -415,7 +438,7 @@ func checkServeCmdConfigFile() error {
 	return nil
 }
 
-func StartDuplexServer(ctx context.Context, cfg *config.FulcioConfig, ctClient *ctclient.LogClient, baseca certauth.CertificateAuthority, host string, port, metricsPort int, ip identity.IssuerPool) error {
+func StartDuplexServer(ctx context.Context, cfg *config.FulcioConfig, ctClient *ctclient.LogClient, baseca certauth.CertificateAuthority, algorithmRegistry *signature.AlgorithmRegistryConfig, host string, port, metricsPort int, ip identity.IssuerPool) error {
 	logger, opts := log.SetupGRPCLogging()
 
 	d := duplex.New(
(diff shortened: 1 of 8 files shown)


tetsuo-cpp:
@woodruffw @ret2libc
As discussed, this isn't ready but I've opened it up for discussion since it probably makes sense to share some of this logic with Rekor. I haven't written Go before so please let me know if I'm not doing things idiomatically, I don't really know what I'm doing.

pkg/server/algorithm_registry.go
tetsuo-cpp, line 1: It's this logic here that I think needs to be shared. I think we'll pretty much need to do the same thing in Rekor (in a different spot of course).

pkg/server/grpc_server.go
```diff
@@ -105,6 +109,11 @@ func (g *grpcaCAServer) CreateSigningCertificate(ctx context.Context, request *f
 		if err := csr.CheckSignature(); err != nil {
 			return nil, handleFulcioGRPCError(ctx, codes.InvalidArgument, err, invalidSignature)
 		}
+
+		hashFunc, err = getHashFuncForSignatureAlgorithm(csr.SignatureAlgorithm)
```
tetsuo-cpp, line 113: Unfortunately, this type in the x509 library doesn't expose a way to get information about the hash function (although it's obvious from looking at it). So I've had to write this function that maps from the `crypto.x509.SignatureAlgorithm` to `crypto.Hash` (documented before at https://github.com/golang/go/issues/33317).

pkg/server/algorithm_registry.go
```diff
@@ -0,0 +1,63 @@
...
+
+type AlgorithmRegistry struct {
+	permittedAlgorithms []Algorithm
+}
+
+func NewAlgorithmRegistry(algorithmConfig []string) (*AlgorithmRegistry, error) {
+	var permittedAlgorithms []Algorithm
+	for _, algorithm := range algorithmConfig {
+		switch algorithm {
+		case "ecdsa-sha2-256-nistp256":
```
tetsuo-cpp, line 45: TODO: Put these somewhere as constants since we're using them as default value for the CLI flag.

Also, regardless of whether this code gets moved to `sigstore/sigstore`, should this code rely on an enumeration in `sigstore/protobuf-specs` so we have a single source of truth for what combinations are allowed in the ecosystem?

tetsuo-cpp: As far as I understand, `sigstore/sigstore` doesn't have a dependency to `protobuf-specs` so we'd need to add that.

tetsuo-cpp:
I've shown in [d95ff74](https://github.com/sigstore/fulcio/pull/1517/commits/d95ff74a85eae11b9c6eca4e5e98b59e59b3de5a) how we can use an enum in `sigstore/protobuf-specs` to avoid redundancy in Fulcio and Rekor in CLI parsing, etc.

go.mod
```diff
@@ -167,3 +168,5 @@ require (
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 )
+
+replace github.com/sigstore/protobuf-specs => /Users/tetsuo/Code/protobuf-specs
```
tetsuo-cpp, line 172: If you want to test this locally, clone this branch: https://github.com/trail-of-forks/protobuf-specs/tree/alex/configurable-crypto
And point it to your local copy.

pkg/server/grpc_server.go
```diff
@@ -279,3 +295,13 @@ func (g *grpcaCAServer) Check(_ context.Context, _ *health.HealthCheckRequest) (
 func (g *grpcaCAServer) Watch(_ *health.HealthCheckRequest, _ health.Health_WatchServer) error {
 	return status.Error(codes.Unimplemented, "unimplemented")
 }
+
+func getHashFuncForSignatureAlgorithm(signatureAlgorithm x509.SignatureAlgorithm) (crypto.Hash, error) {
+	switch signatureAlgorithm {
+	case x509.ECDSAWithSHA256:
```
ret2libc, line 301: BTW, the public Fulcio now issues certificates with ecdsa-with-SHA384.

pkg/server/grpc_server.go
```diff
@@ -132,6 +141,13 @@ func (g *grpcaCAServer) CreateSigningCertificate(ctx context.Context, request *f
 		if err := challenges.CheckSignature(publicKey, proofOfPossession, principal.Name(ctx)); err != nil {
 			return nil, handleFulcioGRPCError(ctx, codes.InvalidArgument, err, invalidSignature)
 		}
+
+		hashFunc = crypto.SHA256
```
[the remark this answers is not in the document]
ret2libc: I don't know whether this path is still supported or deprecated, but we should assume SHA256 when RSA/ECDSA are used and Hash(0) when ED25519 is used, for backward compatibility.

If we need to keep supporting this, we probably have to update the `PublicKeyRequest` in https://github.com/sigstore/fulcio/blob/main/fulcio.proto#L122 to include the sign/hash algorithm used for the proof of possession.

pkg/server/grpc_server.go
ret2libc: Shall we get the sign/hash algos pair as part of the `CreateSigningCertificateRequest` parameter? Otherwise, I'm not sure how we can reliably identify which hash algorithm is used by the client. BTW, if I understand it correctly, the hash algorithm used on the client side is probably never really useful on the Fulcio side, except for doing these kind of checks. However we can't infer the hash algo, because you could have (maybe in the future) ECDSA with a different SHA.

cmd/app/serve.go
```diff
@@ -204,6 +207,20 @@ func runServeCmd(cmd *cobra.Command, args []string) { //nolint: revive
...
+	algorithmStrings := viper.GetStringSlice("client-signing-algorithms")
+	var algorithmConfig []v1.SupportedAlgorithm
+	for _, s := range algorithmStrings {
+		algorithmValue, ok := v1.SupportedAlgorithm_value[s]
+		if !ok {
+			log.Logger.Fatalf("%s is not a valid value for \"client-signing-algorithms\". Try: %s", s, maps.Keys(v1.SupportedAlgorithm_value))
+		}
+		algorithmConfig = append(algorithmConfig, v1.SupportedAlgorithm(algorithmValue))
+	}
+	algorithmRegistry, err := server.NewAlgorithmRegistry(algorithmConfig)
```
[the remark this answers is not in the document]
tetsuo-cpp: Just my 2c, but from an API point of view, it seems error-prone to take a string config. For example, if you were going to instantiate this without a CLI, the compiler will help get this right whereas with raw strings, a typo will throw an error out of the constructor.

pkg/server/grpc_server.go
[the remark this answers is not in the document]
tetsuo-cpp: Interesting, I thought that SHA256 was hard-coded in this path. I'll do a bit of digging to see where I went wrong.

pkg/server/grpc_server.go
tetsuo-cpp: Yeah, good point. I think we need to include this in `CreateSigningCertificateRequest`. I misunderstood the `SignatureAlgorithm` type as the signing algorithm that the client used not Fulcio itself.

ret2libc: We were discussing with @woodruffw that maybe Fulcio should not care about the Hash algorithm used by the client but just about the signing algorithm. Rekor will have to know about the Hash algorithm.

tetsuo-cpp: Hmm, I think `SignatureAlgorithm` is what we want. In the server, we check the signature by calling [this code](https://github.com/golang/go/blob/master/src/crypto/x509/x509.go#L884) which determines the key and hash algorithm by a similar method.

pkg/server/grpc_server.go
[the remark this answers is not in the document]
tetsuo-cpp: Ah, I see my mistake. It's always SHA-256 unless we're using ED25519, in which case it's Hash(0). I'll fix that.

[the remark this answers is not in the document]
tetsuo-cpp: [6208232](https://github.com/sigstore/fulcio/pull/1517/commits/6208232c97e5e7c8093e88ec7e2df91da848954c)

pkg/server/grpc_server.go
ret2libc: Ok, I think you are right! As the proof of possession in the CSR case is the signature of the CSR itself, using SignatureAlgorithm seems right. 👍

tetsuo-cpp:
I'll follow up with some unit tests tomorrow.

cmd/app/serve.go
```diff
@@ -106,6 +109,7 @@ func newServeCmd() *cobra.Command {
 	cmd.Flags().Duration("read-header-timeout", 10*time.Second, "The time allowed to read the headers of the requests in seconds")
 	cmd.Flags().String("grpc-tls-certificate", "", "the certificate file to use for secure connections - only applies to grpc-port")
 	cmd.Flags().String("grpc-tls-key", "", "the private key file to use for secure connections (without passphrase) - only applies to grpc-port")
+	cmd.Flags().StringSlice("client-signing-algorithms", buildDefaultClientSigningAlgorithms([]v1.KnownSignatureAlgorithm{v1.KnownSignatureAlgorithm_ECDSA_SHA2_256_NISTP256, v1.KnownSignatureAlgorithm_ECDSA_SHA2_384_NISTP384, v1.KnownSignatureAlgorithm_ECDSA_SHA2_512_NISTP521, v1.KnownSignatureAlgorithm_ED25519}), "the list of allowed client signing algorithms")
```
ret2libc, line 112: Shall we default to the 256 algos only right now? I think that's the most conservative option as it does not add new algos. Other SHAs were not working before.

pkg/server/grpc_server_test.go
```diff
@@ -1182,6 +1188,94 @@ func TestAPIWithIssuerClaimConfig(t *testing.T) {
 	}
 }
 
+// Tests API with an RSA key
```
tetsuo-cpp, line 1191: I didn't want to unit test the algorithm registry code too heavily here since that's `sigstore/sigstore`'s responsibility.

I'm focusing on:
- Testing that other key types work (RSA in this case).
- Testing that non-permitted algorithms are rejected (ECDSA with P521 curve in this case).
- Testing both the CSR and non-CSR paths.

PR #1518, opened 2024-01-13
Allow exposed metrics port to be overridden

cmurphy:
Rekor uses the same metrics port, so if both services are run with docker-compose they conflict with each other. Add an environment variable to expose the metrics service on a different docker port.

Needed for an update to the cosign e2e tests.

Relates to https://github.com/sigstore/sigstore-probers/issues/105

#### Release Note

Enables docker-compose to use an alternate port for metrics with `FULCIO_METRICS_PORT`.

#### Documentation

Added in docs/setup.md.


docker-compose.yml
```diff
@@ -34,7 +34,7 @@ services:
     ports:
       - "5555:5555"
       - "5554:5554"
-      - "2112:2112"
+      - "${FULCIO_METRICS_PORT:-2112}:2112"
```
cpanato, line 37: i would do 
```suggestion
      - "${FULCIO_METRICS_PORT:-2113}:2112"
```

haydentherapper: LGTM. You can also use set the `metrics-port` flag to set a different metrics port for the service.

cmurphy: @cpanato I didn't want to change the default since it could be considered a breaking change. Do you think having a better default is worth that breaking change?

@haydentherapper I wanted to run this in docker-compose without editing `docker-compose.yml`, so it doesn't really matter to me what port metrics is running on in the container, only what external port is exposed.

PR #1537, opened 2024-01-25
Fix docker-compose service order

cmurphy:
Without this change, when Fulcio is run in docker-compose, signing an artifact may fail with the error "Error entering certificate in CTL". This happens if the docker-compose service have been run previously on the host and the ctfeConfig volume is populated from the last run, so it would generally only be seen in a developer environment. The error happens because the ctfe_init container starts too soon, and ct_server starts with Fulcio's ephemeral root CA from the last run, which is now the wrong CA. This change fixes the issue by ensuring ct_server only starts after ctfe_init has exited successfully, instead of just after it is started. This also means that Fulcio needs to be one of the first services to start so that it can make the ephemeral CA available to download.

#### Release Note

NONE

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -46,7 +46,6 @@ services:
       start_period: 5s
     depends_on:
       - dex-idp
-      - ct_server
     read_only: true
   dex-idp:
     image: dexidp/dex:v2.30.0
@@ -73,6 +72,7 @@ services:
       dockerfile: Dockerfile.ctfe_init
     depends_on:
       - trillian-log-server
+      - fulcio-server
     volumes:
       - ctfeConfig:/etc/config/:rw
   ct_server:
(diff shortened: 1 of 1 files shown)


haydentherapper (approved):
I wonder if we should also be enforcing that the CT log is healthy when starting up the service, here: https://github.com/sigstore/fulcio/blob/b7b2ebacc90700743766167ce66b60bcf86345f4/cmd/app/serve.go#L262-L277

This would prevent Fulcio starting up while the CT log failed to, and all requests failing.

We would have a chicken-and-egg problem though - Fulcio requiring the CT log, the CT log init script requiring Fulcio. We could generate the CA certificate out of band and give that to the CT log rather than use the init script, though we'll need another script to generate fresh CA certs.

Anywho, not a blocker, just something we could change later on.

cmurphy:
> We would have a chicken-and-egg problem though

Right, this wouldn't be possible with the way we're currently running fulcio with `--ca=ephemeralca` in docker-compose. If we were using a different signing backend for development then we could pass the CA to the CT log out of band and we wouldn't have this ordering issue. Since we're not, fulcio has to be up before the CT log can start, at least for this use case.

PR #1593, opened 2024-03-06
Add Codefresh OIDC provider

### LINKED ISSUE #1592: Codefresh OIDC provider support
Codefresh has recently added support for OIDC in pipelines: 
https://codefresh.io/docs/docs/integrations/oidc-pipelines
For integration with Sigstore Codefresh OIDC provider needs to be added to supported providers in Fulcio.

ilia-medvedev:
Add support for Codefresh OIDC provider to allow keyless signing of artifacts in Codefresh pipelines

Closes #1592


haydentherapper:
Hey @ilia-medvedev-codefresh, thanks so much for contributing the provider! Overall this looks good, I'll add a few folks who worked on the previous OIDC/CI integrations to review the claim mappings.

docs/oid-info.md
```diff
@@ -189,27 +189,27 @@ that Sigstore operates.
...
+| runner_environment | runner_environment | ??       | ??        | runner_environment        | Runner Environment           | For platforms to specify whether the build took place in platform-hosted cloud infrastructure or customer-hosted infrastructure. For example: `platform-hosted` and `self-hosted`.     |
+| server_url + repository         | server_url + project_path | ??       | ??        | scm_repo_url        | Source Repository URI                   | Should include a fully qualified repository URL.                                                                                                                                       |
+| sha                | sha | ??       | build_commit | N/A | Source Repository Digest                | An immutable reference to a specific version of the source code. Should include the digest type followed by the digest, e.g. `sha1:abc123`.                                            |
```
wlynch, line 203: Out of scope for this PR, but an ask for Codefresh: it would be great if this could start to be captured in OIDC claims so that users can know what version of the pipeline config was used for the build.

docs/oid-info.md
```diff
@@ -189,27 +189,27 @@ that Sigstore operates.
...
+| server_url + repository_owner         | server_url + namespace_path | ??       | ?? | N/A        | Source Repository Owner URI     | Fully qualified URL for the owner of the source repository.                                                                                                                                     |
+| repository_owner_id         | namespace_id     | ??       | ?? | N/A        | Source Repository Owner Identifier     | Stable identifier for the owner of the source repository.                                                                                                                                   |
+| server_url + workflow_ref       |  ci_config_ref_uri ([WIP][gitlab-wip-cliams]) | ??       | ??        | platform_url + /build/ +  workflow_id        | Build Config URI             | A reference to the initiating build instructions.                                                                                                                                      |
```
wlynch, line 208: I wonder if this should be pipeline_id? 🤔 

My current understanding is the pipeline_id stores the configuration, and workflow_id is a run of the pipeline.

If not, would another field be more appropriate?

ilia-medvedev-codefresh: Changed it to the API endpoint in Codefresh that returns the pipeline config, as the UI endpoint is a bit awkward as it requires a lot of query parameters. But the API endpoint should be completely valid and it uses the pipeine_id claim.

docs/oid-info.md
ilia-medvedev-codefresh: Thank you for your detailed review @wlynch! We will for sure discuss it internally and try to come up with a relevant claim.

docs/oid-info.md
```diff
@@ -189,27 +189,27 @@ that Sigstore operates.
...
+| nbf                | nbf     | nbf       | nbf        | nbf        | N/A                          | Only used to validate the JWT. Optional, as per the OIDC spec                                                                                                                         |
+| iat                | iat     | iat       | iat        | iat        | N/A                          | Only used to validate the JWT.                                                                                                                                                         |
+| server_url + job_workflow_ref   | ci_config_ref_uri ([WIP][gitlab-wip-cliams]) | ??       | ??        | platform_url + /build/ +  workflow_id        | Build Signer URI             | Reference to specific build instructions that are responsible for signing. Can be the same as Build Config URI. For example a reusable workflow in GitHub Actions or a Circle CI Orbs. |
```
steiza, line 199: I just want to check that the URLs starting with `platform_url` in this file in this pull request are right - they aren't namespaces by `account_name` or `pipeline_name` or anything like that?

steiza (commented):
I just want to double-check if the URLs are namespaced at all; otherwise it looks to me like all the MUSTs in https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md are covered!

docs/oid-info.md
ilia-medvedev-codefresh: Yep, the urls are correct. Accounts in Codefresh are populated upon login (either through the UI or API tokens which are scoped to an account)

docs/oid-info.md
```diff
@@ -189,27 +189,27 @@ that Sigstore operates.
...
+| server_url + workflow_ref       |  ci_config_ref_uri ([WIP][gitlab-wip-cliams]) | ??       | ??        | platform_url + /api/pipelines/ +  pipeline_id        | Build Config URI             | A reference to the initiating build instructions.                                                                                                                                      |
+| workflow_sha       |  ci_config_sha ([WIP][gitlab-wip-cliams]) | ??       | ?? | N/A        | Build Config Digest          | An immutable reference to the specific version of the top-level build instructions. Should include the digest type followed by the digest, e.g. `sha1:abc123`.                         |
+| event_name         | pipeline_source | ??       | ??        | N/A        | Build Trigger                | The event or action that triggered the build.                                                                                                                                          |
```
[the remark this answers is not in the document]
ilia-medvedev-codefresh: It's not supported at the moment but I will take it internally, maybe we can support it In the future. Will make sure to add this extension when we do.

docs/oid-info.md
ilia-medvedev-codefresh: @steiza We will for sure take it into account. I'll work with the relevant dev team to have a relevant claim added in the near future, and will add it as soon as it is available.

steiza (approved):
I'm not a Fulcio code owner, but the claim mapping make sense to me.

pkg/identity/codefresh/principal.go
```diff
@@ -0,0 +1,166 @@
...
+
+	// Set SAN to the <platform url>/<account name>/<pipeline name>:<account id>/<pipeline id> - for example https://g.codefresh.io/codefresh-account/oidc-test/get-token:628a80b693a15c0f9c13ab75/65e5a53e52853dc51a5b0cc1
+	// In Codefresh account names and pipeline names may be changed where as IDs do not.
+	// This pattern will give users the possibility to verify the signature using various forms of `cosign verify --certificate-identity-regexp` i.e https://g.codefresh.io/codefresh-account/oidc-test/get-token:* or https://g.codefresh.io/*:628a80b693a15c0f9c13ab75/65e5a53e52853dc51a5b0cc1
+	cert.URIs = []*url.URL{baseURL.JoinPath(w.accountName, fmt.Sprintf("%s:%s/%s", w.pipelineName, w.accountID, w.pipelineID))}
+
+	cert.ExtraExtensions, err = certificate.Extensions{
+		Issuer: w.issuer,
+		// URL of the build in Codefresh
+		BuildSignerURI:      baseURL.JoinPath("build", w.workflowID).String(),
```
haydentherapper, line 153: To confirm, a workflow comes from a pipeline, correct?

Typically build signer would be the same as the SAN, as it's meant to represent an stable identity rather than a more frequently changing identifier like a run ID. I think builder signer should be composed of the pipeline ID rather than workflow ID.

pkg/identity/codefresh/principal.go
```diff
@@ -0,0 +1,166 @@
...
+	baseURL, err := url.Parse(w.platformURL)
+
+	if err != nil {
+		return err
+	}
+
+	// Set SAN to the <platform url>/<account name>/<pipeline name>:<account id>/<pipeline id> - for example https://g.codefresh.io/codefresh-account/oidc-test/get-token:628a80b693a15c0f9c13ab75/65e5a53e52853dc51a5b0cc1
+	// In Codefresh account names and pipeline names may be changed where as IDs do not.
+	// This pattern will give users the possibility to verify the signature using various forms of `cosign verify --certificate-identity-regexp` i.e https://g.codefresh.io/codefresh-account/oidc-test/get-token:* or https://g.codefresh.io/*:628a80b693a15c0f9c13ab75/65e5a53e52853dc51a5b0cc1
+	cert.URIs = []*url.URL{baseURL.JoinPath(w.accountName, fmt.Sprintf("%s:%s/%s", w.pipelineName, w.accountID, w.pipelineID))}
```
haydentherapper, line 148: If account name and pipeline are mutable, should this be composed on only the account and pipeline IDs?

Are pipeline IDs unique across all pipelines, or just unique within an account? If the former, then this could be composed of just a pipeline ID, or maybe use account name + pipeline ID so it's more human readable?

Hayden-IO (commented):
Code looks great! Just a few questions for subject value.

pkg/identity/codefresh/principal.go
ilia-medvedev-codefresh: Pipeline IDs are unique across accounts. Problem is that it's not very intuitively accessible to users from Codefresh UI, and even more so it's not accessible to unauthenticated users in case of public builds. But I think the usage of `--certificate-identity-regexp` allows complete flexibility in the proposed way of structuring the SAN. It allows the verifier to decide which parameters they trust.

pkg/identity/codefresh/principal.go
ilia-medvedev-codefresh: A workflow is an instance of a pipeline, yes. But it Codefresh, pipeline definitions are only accessible to authenticated users - if you have a public build only the workflow run URL is accessible to unauthenticated users. So in case of an open source project that uses Codefresh, if one wanted to see the pipeline that was used to build their artifact they can only access the workflow if they are not authenticated to Codefresh. Plus the workflow contains the pipeline definition that was used at the time of building the artifact which represents the signer well.

ilia-medvedev-codefresh: As for the runner_environment - it has been added recently to support Fulcio requirements. It is currently behind a feature flag - we will open it for everyone shortly.

ilia-medvedev-codefresh:
@haydentherapper @wlynch @feelepxyz Anything I can do or help clarify further?

pkg/identity/codefresh/principal.go
haydentherapper: Thanks for the explanation, that makes sense. Would you be able to add a comment summarizing that, since this is the first example of a SAN being different than the build signer URI?

Hayden-IO (commented):
@ilia-medvedev-codefresh sorry for the delay, this looks good overall! just two small comments

pkg/identity/codefresh/principal.go
ilia-medvedev-codefresh: Account name is indeed there - It's the first member in `JoinPath`

pkg/identity/codefresh/principal.go
ilia-medvedev-codefresh: @haydentherapper sure thing. Comment was added.

pkg/identity/codefresh/principal.go
haydentherapper: whoops, sorry, i was looking at the latter half!

haydentherapper:
@ilia-medvedev-codefresh just needs a lint fix, then I'll merge, cut a release, and work on getting this out. I'll ping the thread once we have this in our staging environment so you can confirm it's working.

Hayden-IO (approved):

PR #1625, opened 2024-04-07
Upgrade go to 1.22

cpanato:
#### Summary

- Upgrade go to 1.22

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/sigstore/fulcio
 
-go 1.21
+go 1.21.9
 
 require (
 	chainguard.dev/go-grpc-kit v0.17.2
--- a/go.mod
+++ b/go.mod
@@ -2,8 +2,6 @@ module github.com/sigstore/fulcio
 
 go 1.21
 
-toolchain go1.21.3
-
 require (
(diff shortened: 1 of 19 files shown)


go.mod
```diff
@@ -2,8 +2,6 @@ module github.com/sigstore/fulcio
 
 go 1.21
 
-toolchain go1.21.3
```
haydentherapper, line 5: I ran into this when trying to upgrade rekor-monitor and finally figured out the issue - apparently `1.21` or `1.22` is a "development" version according to https://github.com/golang/go/issues/62278 and toolchain should be specified or the exact version of go should be specified like `1.21.x`. I was a bit surprised by this, feels like a significant change. So I think we should say `go 1.21.6` or whatever the last 1.21 is

cpanato: if we set the toolchain to go1.21.x then it will fetch that version to build/test and not go1.22 if we want that i will drop this pr and update the toolchain to the latest one if that is possible but for go1.21.9

haydentherapper: Can we switch to specifying go 1.21.9, and not specifying the tool chain, and still test using 1.22?

cpanato: let me try

haydentherapper (approved):

PR #1661, opened 2024-05-14
docs: Fix extensions for digest values requiring a type prefix

facutuesca:
This PR has two commits:
- The first one fixes the table mapping OIDC token claims to OIDs, changing digest extensions (i.e: `Build Signer Digest`, `Source Repository Digest` and  `Build Config Digest`) to say that their value MAY include the digest type (as in: `sha1:$SHA_VALUE`), rather than SHOULD include it.
- The second one formats the plain text of the table so that each column has the same width (the difference can only be seen in plaintext, the rendered markdown should be the same):
image

The second commit is optional, I can remove it if it's unwanted

cc @woodruffw @haydentherapper


woodruffw (approved):
LGTM in terms of observed behavior! I suppose one could argue that "should" is already non-normative, though 🙂

haydentherapper (approved):
Thank you!

cpanato (approved):
thanks!

Issue #1111, opened 2023-04-12
Simplify integration for new CI/CD providers


haydentherapper:
**Description**

- [ ] Create common CI provider identity (e.g. generic version of https://github.com/sigstore/fulcio/tree/main/pkg/identity/github) that takes in mapping between OIDC claim name and OID
- [ ] Add server configuration option to map OIDC claims to CI/CD OIDs for each CI provider
- [ ] Update public instance with this mapping
- [ ] Add documentation

PR #1679, opened 2024-05-27
CiProvider as a new OIDCIssuer type

javanlacerda:
Contribute towards #1111 

#### Summary
It adds CiProvider as a new OIDCIssuer type. We will migrate all ci providers to use a generic principal by changing their types to this new type.

It should not change any current behavior.

#### Documentation

cc @haydentherapper


pkg/identity/generic/issuer_test.go
```diff
@@ -0,0 +1,81 @@
+// Copyright 2023 The Sigstore Authors.
```
cpanato, line 1: update the boilerplate here and in other new files to be 

```suggestion
// Copyright 2024 The Sigstore Authors.
```

cpanato (changes requested):
and when you have time can you describe the PR and why we need that?

thanks!

haydentherapper:
@cpanato, the motivation is to simplify CI/CD OIDC provider onboarding. Rather than have each OIDC provider have to modify code to add a new provider, they instead should only need to modify a configuration file which will contain the mapping between OIDC claim and x509 extension value.

cpanato:
> @cpanato, the motivation is to simplify CI/CD OIDC provider onboarding. Rather than have each OIDC provider have to modify code to add a new provider, they instead should only need to modify a configuration file which will contain the mapping between OIDC claim and x509 extension value.

that is nice! thanks for the clarification

pkg/identity/generic/issuer.go
haydentherapper, line 1: I would call this "ci_provider" rather than "generic".

pkg/challenges/challenges.go
```diff
@@ -58,25 +59,29 @@ func PrincipalFromIDToken(ctx context.Context, tok *oidc.IDToken) (identity.Prin
...
-		principal, err = github.WorkflowPrincipalFromIDToken(ctx, tok)
-	case config.IssuerTypeKubernetes:
-		principal, err = kubernetes.PrincipalFromIDToken(ctx, tok)
-	case config.IssuerTypeURI:
-		principal, err = uri.PrincipalFromIDToken(ctx, tok)
-	case config.IssuerTypeUsername:
-		principal, err = username.PrincipalFromIDToken(ctx, tok)
-	default:
-		return nil, fmt.Errorf("unsupported issuer: %s", iss.Type)
+	if iss.IsCiProvider {
```
haydentherapper, line 62: Instead of adding a new flag, can we use `iss.Type`? The `Type` of each CI OIDC provider could be `"CIProvider"`. Then the new issuer would simply be another case statement.

javanlacerda: I agree! I was just wondering how could we know the real type of the provider after migrating all of them, because we should need know what claim mapping use for each provider. But I think it will work by using the provider name as string and not as a type.

pkg/identity/ciprovider/principal.go
```diff
@@ -0,0 +1,165 @@
...
+	"gopkg.in/yaml.v3"
+)
+
+type Config struct {
+	Providers map[config.IssuerType]Provider
+}
+type Provider struct {
+	Subject     string
+	Extensions  certificate.Extensions
+	Uris        []string
```
haydentherapper, line 40: Is URIs the SubjectAlternativeName? If so, should we call it `SubjectAlternativeName`? And not make it repeated, since we only support a single SAN?

haydentherapper (changes requested):

pkg/identity/ciprovider/principal.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+		uris = append(uris, url)
+	}
+	// Set workflow ref URL to SubjectAlternativeName on certificate
+	cert.URIs = uris
+
+	var err error
+	// Embed additional information into custom extensions
+	cert.ExtraExtensions, err = certificate.Extensions{
+		Issuer:                              ApplyTemplate(e.Issuer, claims, defaults),
```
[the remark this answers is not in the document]
javanlacerda: What do you mean with "specific claim"? My point here is passing the issuer specific claims as a map to the `applyTemplate` function, and the function knows how to replace the map key to the value by using the template.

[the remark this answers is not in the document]
javanlacerda: Agree, but the `applyTemplate` knows how to replace the claim when the template is not applicable. I'll change the function name to `applyTemplateOrReplace`

pkg/identity/ciprovider/principal.go
```diff
@@ -0,0 +1,141 @@
...
+	return stringClaims
+}
+
+// It makes string interpolation for a given string by using the
+// templates syntax https://pkg.go.dev/text/template
+func applyTemplateOrReplace(path string, data map[string]string, defaultData map[string]string) string {
+
+	// Here we merge the data from was claimed by the id token with the
+	// default data provided by the yaml file.
+	// The order here matter because we want to override the default data
```
haydentherapper, line 46: I think we want the reverse order here. The values provided as defaults are supplemental, as in we don't expect `url` to be set as a value in the token. I don't think we should err out if `url` was set in both because we don't control the contents of the token, but we should specify in the configuration set up to choose additional values where those keys won't be present in the token.

If they are present, then we should choose what's in `defaultData` over `data`.

WDYT?

pkg/identity/ciprovider/principal.go
```diff
@@ -0,0 +1,141 @@
...
+		mergedData[k] = v
+	}
+	for k, v := range data {
+		mergedData[k] = v
+	}
+
+	if strings.Contains(path, "{{") {
+		var doc bytes.Buffer
+		t := template.New("")
+		p, err := t.Parse(path)
```
haydentherapper, line 59: We should add a validation function to parse all templated strings on config load, and return an error and prevent the CA from starting up.

We shouldn't include panics in the codebase since that would cause server crashes. If the template isn't parseable here, we should return an error, but we can add a comment that this shouldn't happen since we parse the template strings on config load.

pkg/identity/ciprovider/principal.go
```diff
@@ -0,0 +1,141 @@
...
+		if err != nil {
+			panic(err)
+		}
+		err = p.Execute(&doc, mergedData)
+		if err != nil {
+			panic(err)
```
haydentherapper, line 65: Same here, should return an error. We can add a message to the error that the token was missing an expected claim (as in, if the value is set to `{{ .url }}/{{ .job_workflow_ref }}`, and `job_workflow_ref` is missing from the token, that's a user error, not server error)

pkg/identity/ciprovider/principal.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+		return doc.String()
+	}
+	return mergedData[path]
```
haydentherapper, line 69: We need to check if the key is present. If it's missing, then it's an error. (also need to update function to return an error as well)

haydentherapper (changes requested):

pkg/identity/ciprovider/principal.go
javanlacerda: It makes sense!

haydentherapper:
Once this PR is ready for review and all comments addressed, can you post here?

pkg/identity/ciprovider/principal.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+type Config struct {
+	Token    *oidc.IDToken
+	Metadata config.IssuersMetadata
+}
+
+func WorkflowPrincipalFromIDToken(ctx context.Context, token *oidc.IDToken) (identity.Principal, error) {
+	cfg := config.FromContext(ctx)
+	issuer, ok := cfg.GetIssuer(token.Issuer)
```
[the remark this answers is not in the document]
javanlacerda: I've set as issuerCfg, as we already have a `cfg` variable.

pkg/identity/ciprovider/principal.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		panic(err)
+	}
+	uris := []*url.URL{subjectAlternativeNameURL}
+	// Set workflow ref URL to SubjectAlternativeName on certificate
+	cert.URIs = uris
+
+	// Embed additional information into custom extensions
+	cert.ExtraExtensions, err = certificate.Extensions{
+		Issuer:                              applyTemplateOrReplace(e.Issuer, claims, defaults),
```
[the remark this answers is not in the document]
javanlacerda: sounds good. It cannot be set on the configuration file as the yaml mapper on the struct for issuer is empty. I also added a lock assuring that this field will always come from the token.issuer

javanlacerda:
> Once this PR is ready for review and all comments addressed, can you post here?

Hey @haydentherapper, could you take a look?

pkg/certificate/extensions.go
```diff
@@ -69,69 +69,69 @@ type Extensions struct {
 	// Deprecated
 	// Triggering event of the Github Workflow. Matches the `event_name` claim of ID
 	// tokens from Github Actions
-	GithubWorkflowTrigger string // OID 1.3.6.1.4.1.57264.1.2
+	GithubWorkflowTrigger string `yaml:"github-workflow-trigger,omitempty"` // OID 1.3.6.1.4.1.57264.1.2
```
haydentherapper, line 72: You should also add the JSON tags as well since this could be used in JSON configuration.

pkg/identity/ciprovider/principal.go
```diff
@@ -0,0 +1,161 @@
...
+		return err
+	}
+	sanURL, err := url.Parse(subjectAlternativeName)
+	if err != nil {
+		return err
+	}
+	uris := []*url.URL{sanURL}
+	cert.URIs = uris
+	mapExtensionsForTemplate := make(map[string]interface{})
+	err = mapstructure.Decode(claimsTemplates, &mapExtensionsForTemplate)
```
haydentherapper, line 131: Same suggestion to use reflection. Something like:

```
v := reflect.ValueOf(claimsTemplates)
for i := 0; i < v.NumField(); i++ {
  s := v.Field(i).String() // value of each field, e.g the template string
  if s == "" {
    continue
  }
  extValue, err := applyTemplateOrReplace(s, claims, defaults)
  if err != nil {
    return err
  }
  v.Field(i).Set(extValue)
}

claimsTemplates.Issuer = principal.Token.Issuer

// claimsTemplates is now a fully populated certificate.Extensions struct
// call Render
```

Reflection is always a bit scary to use, but in this case, I think for what you're trying to do, which is enumerate over a structure without needing to know each field, this is a clean and idiomatic way to do so.

haydentherapper (changes requested):

pkg/config/config.go
```diff
@@ -432,6 +457,36 @@ func FromContext(ctx context.Context) *FulcioConfig {
 	return untyped.(*FulcioConfig)
 }
 
+// It checks that the templates defined are parseable
+// We should check it during the service bootstrap to avoid errors further
+func CheckParseTemplates(fulcioConfig *FulcioConfig) error {
+
+	checkParse := func(temp interface{}) error {
```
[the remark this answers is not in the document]
javanlacerda: make sense

pkg/identity/ciprovider/principal_test.go
```diff
@@ -0,0 +1,342 @@
...
+	"fmt"
+	"reflect"
+	"testing"
+	"unsafe"
+
+	"github.com/coreos/go-oidc/v3/oidc"
+	"github.com/sigstore/fulcio/pkg/certificate"
+	"github.com/sigstore/fulcio/pkg/config"
+)
+
```
haydentherapper, line 33: Are there any tests we can add for `applyTemplateOrReplace`, to check any edge cases?

pkg/server/grpc_server_test.go
```diff
@@ -1123,6 +1131,178 @@ func TestAPIWithGitHub(t *testing.T) {
 	}
 }
 
+// Tests API for CiProvider subject types
+func TestAPIWithCiProvider(t *testing.T) {
```
haydentherapper, line 1135: Nice test!

haydentherapper (commented):
Just a few comments mostly around testing, this is coming together well!

pkg/config/config.go
```diff
@@ -432,6 +458,34 @@ func FromContext(ctx context.Context) *FulcioConfig {
...
+func validateCIIssuerMetadata(fulcioConfig *FulcioConfig) error {
+
+	checkParse := func(temp string) error {
+		t := template.New("").Option("missingkey=error")
+		_, err := t.Parse(temp)
+		return err
+	}
+
+	for _, ciIssuerMetadata := range fulcioConfig.CIIssuerMetadata {
+		v := reflect.Indirect(reflect.ValueOf(&ciIssuerMetadata.ExtensionTemplates))
```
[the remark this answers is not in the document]
javanlacerda: > Does `reflect.ValueOf(ciIssuerMetadata.ExtensionTemplates)` work, or do we need the indirect reflection to resolve the pointer?

We need it for resolving the pointer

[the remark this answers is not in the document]
haydentherapper: Do you need to take a pointer to `ciIssuerMetadata.ExtensionTemplates` though? Are `reflect.ValueOf(ciIssuerMetadata.ExtensionTemplates)` and `reflect.Indirect(reflect.ValueOf(&ciIssuerMetadata.ExtensionTemplates))` equivalent?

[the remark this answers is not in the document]
javanlacerda: It seems that `reflect.ValueOf(ciIssuerMetadata.ExtensionTemplates)` works. I assumed that I should use indirect for this case, as I needed to do this for using `SetString`. Without the indirect I got `reflect.Value.SetString using unaddressable value` as an error

pkg/config/config_network_test.go
```diff
@@ -68,6 +69,47 @@ func TestLoad(t *testing.T) {
...
+		ExtensionTemplates: certificate.Extensions{
+			BuildTrigger: invalidTemplate,
+		},
+	}
+	fulcioConfig := &FulcioConfig{
+		CIIssuerMetadata: ciissuerMetadata,
+	}
+	err := validateCIIssuerMetadata(fulcioConfig)
+	if err == nil {
+		t.Error("It should raise an error")
```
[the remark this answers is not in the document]
javanlacerda: Yes, on it

pkg/config/config_network_test.go
```diff
@@ -68,6 +69,47 @@ func TestLoad(t *testing.T) {
...
+	ciissuerMetadata["github"] = IssuerMetadata{
+		SubjectAlternativeNameTemplate: invalidTemplate,
+	}
+	fulcioConfig = &FulcioConfig{
+		CIIssuerMetadata: ciissuerMetadata,
+	}
+	err = validateCIIssuerMetadata(fulcioConfig)
+	if err == nil {
+		t.Error("It should raise an error")
+	}
```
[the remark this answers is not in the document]
javanlacerda: yep! just did. thanks!!

pkg/identity/ciprovider/principal.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return err
+	}
+	sanURL, err := url.Parse(subjectAlternativeName)
+	if err != nil {
+		return err
+	}
+	uris := []*url.URL{sanURL}
+	cert.URIs = uris
+	v := reflect.Indirect(reflect.ValueOf(&claimsTemplates))
```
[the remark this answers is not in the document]
javanlacerda: In this case it is. As I mentioned, getting the pointer directly by `reflect.ValueOf`, I have an `unaddressable value`. Setting it as a reference and then using the `indirect` was the workaround I found for this.

pkg/identity/ciprovider/principal.go
```diff
@@ -0,0 +1,151 @@
...
+	vType := v.Type()
```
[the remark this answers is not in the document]
javanlacerda: Yes! thanks

pkg/identity/ciprovider/principal_test.go
javanlacerda: Sure! I added the test structure just with the happy path, I'll add more tests for edge cases soon

haydentherapper (approved):
🥳

haydentherapper:
In terms of release, I'm planning to cut 1.5 from before this PR, then we'll merge this and the other related changes and cut v1.6

PR #1705, opened 2024-06-21
Update /healthz probe to use localhost for gRPC readiness checks

### LINKED ISSUE #1704: `/healthz` probe fails in Proxy-Enabled Environment
**Description**

The `/healthz` probe fails to function correctly in environments where a proxy is enabled. The issue stems from the probe using the `0.0.0.0` address to test gRPC readiness. This address cannot be effectively filtered using the `NO_PROXY` setting, resulting in readiness requests to gRPC being routed through the proxy. Consequently, these requests fail because the external proxy attempts to access `127.0.0.1`.

```
HTTP_PROXY="http://x.x.x.x:3128"
HTTPS_PROXY="http://x.x.x.x:3128"
NO_PROXY=".cluster.local,.svc,.us-east-2.compute.internal,10.0.0.0/16,10.128.0.0/14,127.0.0.1,169.254.169.254,172.30.0.0/16,localhost"
```

**Expected Behavior:**
The `/healthz` probe should successfully check the readiness of gRPC services without routing the requests through the proxy. The probe should use an address that can be filtered via the `NO_PROXY` setting.

**Version**
1.4.5

osmman:
#### Summary

This pull request resolves an issue with the `/healthz` probe failing in environments with a proxy enabled. The probe was using the `0.0.0.0` address for gRPC readiness checks, which could not be filtered by the `NO_PROXY` setting, causing readiness requests to be routed through the proxy and fail. This change updates the probe to use the `localhost` address, ensuring that these requests bypass the proxy and function correctly.

Closes #1704

#### Release Note

Updated `/healthz` probe to use localhost for gRPC readiness checks.

#### Documentation
No documentation updates are required for this change.

### DIFF
--- a/cmd/app/http.go
+++ b/cmd/app/http.go
@@ -63,7 +63,8 @@ func createHTTPServer(ctx context.Context, serverEndpoint string, grpcServer, le
 	} else {
 		opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	}
-	cc, err := grpc.NewClient(grpcServer.grpcServerEndpoint, opts...)
+	grpcHealthEndpoint := fmt.Sprintf("localhost:%s", viper.GetString("grpc-port"))
+	cc, err := grpc.NewClient(grpcHealthEndpoint, opts...)
 	if err != nil {
 		log.Logger.Fatal(err)
 	}


bobcallaway (approved):
thanks!

cpanato (approved):

PR #1725, opened 2024-07-10
chainguard issuer: use the same proof of possession cosign expects

priyawadhwa:
this should allow the chainguard issuer to work when the email verified claim is set

### DIFF
--- a/pkg/identity/chainguard/principal.go
+++ b/pkg/identity/chainguard/principal.go
@@ -22,6 +22,7 @@ import (
 	"github.com/coreos/go-oidc/v3/oidc"
 	"github.com/sigstore/fulcio/pkg/certificate"
 	"github.com/sigstore/fulcio/pkg/identity"
+	"github.com/sigstore/sigstore/pkg/oauthflow"
 )
 
 type workflowPrincipal struct {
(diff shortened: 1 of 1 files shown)


pkg/identity/chainguard/principal.go
```diff
@@ -50,9 +51,14 @@ func PrincipalFromIDToken(_ context.Context, token *oidc.IDToken) (identity.Prin
...
 
+	subject, err := oauthflow.SubjectFromToken(token)
+	if err != nil {
+		return nil, err
+	}
+
 	return &workflowPrincipal{
 		issuer:           token.Issuer,
-		subject:          token.Subject,
+		subject:          subject,
```
mattmoor, line 61: We want our subject to be our subject.  It's the challenge that's wrong.

mattmoor: I don't know what else keys off of the `principal.Name` but I could see us changing that.

I think that Fulcio and `cosign` using different functions to extract the information being signed for PoP is incredibly fragile because in cases like this the different pieces of logic will disagree even though nothing is actually incorrect (they are just extracting things inconsistently).

mattmoor: It looks like this is challenging to call directly the way the principal interface is set up, and the only call site of `Name()` is for the PoP so I can stage a change to do this with some test cases.

haydentherapper: I would create another field, `popSubject`, set equal to `oauthflow.SubjectFromToken`, and update `Name()` to return `token.popSubject`.

mattmoor: @haydentherapper I just saw your comment, that's what my PR does essentially.

haydentherapper: PR LGTM

haydentherapper:
Closing in favor of the other

PR #1726, opened 2024-07-10
Surface the right `Name()` from our principal.

mattmoor:
#### Summary

The cosign logic for interacting with Fulcio treats identity tokens as *largely* opaque, and most of the logic for how issuers and subjects and whatnot is handled happens server-side.  However, for the "proof of possession" `cosign` has some logic (from `sigstore/sigstore`) that fumbles with `email` and `sub` claims in ways that have (until now) been compatible with Fulcio principals.

The Chainguard provider is the first provider that optionally includes an `email` claim, but we always want the subject we use to be our opaque identifier string (from `sub`).  This creates a tear in the fulcio/cosign continuum, and so we must surface what `cosign` is signing as `Name()` even though that isn't necessarily what we embed in the certificate.

The only correct way to implement `Name()` today is to match what this function does, and current implementations happen to align, but unfortunately because of how this abstraction is formulated it is challenging to actually change how we confirm the proof of possession to use this directly in place of the principal itself.

Fixes: https://github.com/sigstore/cosign/issues/3777

#### Release Note

This corrects the Chainguard PoP verification flow for tokens that embed email claims (makes `Name()` consistent with `cosign`).

#### Documentation

N/A

### DIFF
--- a/pkg/identity/chainguard/principal.go
+++ b/pkg/identity/chainguard/principal.go
@@ -22,11 +22,13 @@ import (
 	"github.com/coreos/go-oidc/v3/oidc"
 	"github.com/sigstore/fulcio/pkg/certificate"
 	"github.com/sigstore/fulcio/pkg/identity"
+	"github.com/sigstore/sigstore/pkg/oauthflow"
 )
 
 type workflowPrincipal struct {
 	issuer  string
 	subject string
+	name    string
 
 	actor            map[string]string
 	servicePrincipal string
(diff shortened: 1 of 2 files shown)


mattmoor:
whoops, amended with my test file

haydentherapper (approved):

haydentherapper:
I'll cut 1.5.1 with this fix. I'm going to need to revert a PR that when in recently first, as a cherry-picked release is a bit of a pain.

PR #1736, opened 2024-07-15
Removes identity providers federation

javanlacerda:
Contribute towards https://github.com/sigstore/fulcio/issues/1111

It removes completely the identity providers federation, and update the documentation for adding new identity providers, including ci-providers.

#### Documentation

cc @haydentherapper @cpanato


pkg/config/config.go
```diff
@@ -111,6 +111,11 @@ type OIDCIssuer struct {
 	// Optional, the challenge claim expected for the issuer
 	// Set if using a custom issuer
 	ChallengeClaim string `json:"ChallengeClaim,omitempty" yaml:"challenge-claim,omitempty"`
+	// Optional, the description for the issuer
```
haydentherapper, line 114: We want to omit this fields when querying the API, so we need to define another struct without these fields and wire that up as well.

javanlacerda: hmm, is the issuer retuned from the API this one? https://github.com/sigstore/fulcio/blob/b879601e1aaf48de01f2997ba7be339377b7b93e/pkg/config/config.go#L229. because it seems that this one comes from a proto, that omits this information. I also tested it by calling the api locally and the information was omitted.

javanlacerda: image

Example of api's return

haydentherapper: Ah yep, you're right! All good then, thanks.

docs/oidc.md
```diff
@@ -10,13 +10,18 @@ Sigstore runs a federated OIDC identity provider, Dex. Users authenticate to the
...
-* Add a case statement to map the issuer constant to the issuer type you created [here](https://github.com/sigstore/fulcio/blob/4d9d96a/pkg/server/issuer_pool.go#L40-L62)
-* Update the end-to-end gRPC tests:
-   * Update the [configuration test](https://github.com/sigstore/fulcio/blob/572b7c8496c29a04721f608dd0307ba08773c60c/pkg/server/grpc_server_test.go#L175)
-   * Add a test for the new issuer ([example](https://github.com/sigstore/fulcio/blob/572b7c8496c29a04721f608dd0307ba08773c60c/pkg/server/grpc_server_test.go#L331))
+* Add the new issuer to the [configuration](https://github.com/sigstore/fulcio/blob/main/config/identity/config.yaml).
+  * Atention: If your issuer is a ci, you should set the `type` as `ci-provider` and set the field `ci-provider` with the name of your provider. You should also fill the `ci-issuer-metadata` with the `default-template-values`, `extension-templates` and `subject-alternative-name-template`, following the pattern defined on the example ([example](tbd after migrating the github to ci-provider)).
```
haydentherapper, line 14:
```suggestion
  * Attention: If your issuer is for a CI provider, you should set the `type` as `ci-provider` and set the field `ci-provider` with the name of your provider. You should also fill the `ci-issuer-metadata` with the `default-template-values`, `extension-templates` and `subject-alternative-name-template`, following the pattern defined on the example ([example](tbd after migrating the github to ci-provider)).
```

docs/oidc.md
```diff
@@ -10,13 +10,18 @@ Sigstore runs a federated OIDC identity provider, Dex. Users authenticate to the
...
+  * Important notes: The `extension-templates` and the `subject-alternative-name-template` follows the templates [pattern](https://pkg.go.dev/text/template). The name used to fill the `ci-provider` field have to be the same used as key for `ci-issuer-metadata`, we suggest to use a variable for this.
```
haydentherapper, line 15:
```suggestion
  * Important notes: The `extension-templates` and the `subject-alternative-name-template` follows the templates [pattern](https://pkg.go.dev/text/template). The name used to fill the `ci-provider` field has to be the same used as key for `ci-issuer-metadata`, we suggest to use a variable for this.
```

docs/oidc.md
```diff
@@ -10,13 +10,18 @@ Sigstore runs a federated OIDC identity provider, Dex. Users authenticate to the
...
+* If your issuer is not a ci, you need to follow the next steps:
```
haydentherapper, line 16:
```suggestion
* If your issuer is not for a CI provider, you need to follow the next steps:
```

docs/oidc.md
```diff
@@ -10,13 +10,18 @@ Sigstore runs a federated OIDC identity provider, Dex. Users authenticate to the
...
+  * Add the new issuer to the [`identity` folder](https://github.com/sigstore/fulcio/tree/main/pkg/identity) ([example](https://github.com/sigstore/fulcio/tree/main/pkg/identity/buildkite)). You will define an `Issuer` type and a way to map the token to the certificate extensions.
```
haydentherapper, line 17: Should we change the example to a non-CI provider?

haydentherapper (commented):
Thanks, just a few doc comments!

docs/oidc.md
javanlacerda: I think it makes sense

haydentherapper (approved):
Thanks!

PR #1743, opened 2024-07-26
Move codefresh and buildkite to ci-provider identity

javanlacerda:
#### Summary
Contribute towards https://github.com/sigstore/fulcio/issues/1111

It moves the remaining identity providers from CI to using the new ci-provider identity flow.
It included moving `codefresh` and `buildkite`.

It also updates the priority of default templates over claimed data. The defaults should be used only if the claimed data doesn't exist. Then for following this rules, the default key needs to have the same name of the claim key.


config/identity/config.yaml
```diff
@@ -162,3 +166,18 @@ ci-issuer-metadata:
         run-invocation-uri: "{{ .url }}/{{ .project_path }}/-/jobs/{{ .job_id }}"
         source-repository-visibility-at-signing: "repository_visibility"
     subject-alternative-name-template: "https://{{ .ci_config_ref_uri }}"
+  *codefresh-type:
+    default-template-values:
+        url: "https://g.codefresh.io"
+    extension-templates:
+        build-signer-uri: "{{if .platform_url}}{{.platform_ur}}{{ else }}{{.url}}{{end}}/build/{{ .workflow_id }}"
```
haydentherapper, line 173:
```suggestion
        build-signer-uri: "{{if .platform_url}}{{.platform_url}}{{ else }}{{.url}}{{end}}/build/{{ .workflow_id }}"
```

haydentherapper, line 173: Need to update `platform_ur` to `platform_url`

pkg/identity/ciprovider/principal.go
```diff
@@ -66,7 +66,7 @@ func applyTemplateOrReplace(extValueTemplate string, tokenClaims map[string]stri
 		var doc bytes.Buffer
 		// This option forces to having the claim that is required
 		// for the template
-		t := template.New("").Option("missingkey=error")
+		t := template.New("").Option("missingkey=zero")
```
haydentherapper, line 69: Pinged offline about this, but WDYT about what you had mentioned originally, changing the behavior of `default-key-template` to be a default? Then we can maintain the logic of throwing an error if a token is malformed.

config/identity/config.yaml
haydentherapper, line 173: Can we copy in the comments from the principal files, like https://github.com/sigstore/fulcio/blob/main/pkg/identity/codefresh/principal.go#L153-L154 and the ones in buildkite as well?

pkg/identity/ciprovider/principal.go
javanlacerda: I can do that. on it

javanlacerda: Done

config/identity/config.yaml
```diff
@@ -120,45 +124,78 @@ meta-issuers:
...
-          build-config-uri: "{{ .url }}/{{ .workflow_ref }}"
-          build-config-digest: "workflow_sha"
-          build-trigger: "event_name"
-          run-invocation-uri: "{{ .url }}/{{ .repository }}/actions/runs/{{ .run_id }}/attempts/{{ .run_attempt }}"
-          source-repository-visibility-at-signing: "repository_visibility"
-      subject-alternative-name-template: "{{ .url }}/{{ .job_workflow_ref }}"
+    default-template-values:
+      url: "https://github.com"
+    extension-templates:
+      github-workflow-trigger: "event_name"
```
haydentherapper, line 130: While we are updating this, can you add the code comments for each of these claims from the GitHub and gitlab principals?

config/identity/config.yaml
```diff
@@ -120,45 +124,78 @@ meta-issuers:
...
     subject-alternative-name-template: "https://{{ .ci_config_ref_uri }}"
+  *codefresh-type:
+    default-template-values:
+      # We are setting the default value for "platform_url" as the ci-provider
+      # principal gives priority to the claimed value over the default
+      # when they have the same name. Then it will use the default "platform_url" value
+      # for cases that the claimed data doesn't exist.
+      platform_url: "https://g.codefresh.io"
+    extension-templates:
+      # platform_url: Codefresh platform url
```
haydentherapper, line 177: Move this comment up?

pkg/identity/ciprovider/principal.go
```diff
@@ -53,12 +53,12 @@ func applyTemplateOrReplace(extValueTemplate string, tokenClaims map[string]stri
...
 	// with the default data.
-	// The default data will have priority over the claimed data.
+	// The claimed data will have priority over the default data.
 	mergedData := make(map[string]string)
-	for k, v := range tokenClaims {
+	for k, v := range issuerMetadata {
 		mergedData[k] = v
 	}
-	for k, v := range issuerMetadata {
+	for k, v := range tokenClaims {
```
haydentherapper, line 61: Can we add a test that demonstrates this ordering?

haydentherapper (approved):

PR #1762, opened 2024-08-13
fix: removing surplus slash, making logs richer

javanlacerda:
#### Summary
During the validations, we noticed that the the template for `source-repository-ref` for gitlab has a surplus slash.
We also noticed that the error with log `"value <> not present in either claims or defaults.` has spawned several times in production, but we wasn't able to detect the reason for it, then I decided to put at least the issuer that triggered the flow on the error log.
I also added a validation that the metadata should exist for ci-providers.

#### Documentation

cc @haydentherapper

### DIFF
--- a/pkg/identity/ciprovider/principal.go
+++ b/pkg/identity/ciprovider/principal.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/x509"
+	"encoding/json"
 	"fmt"
 	"html/template"
 	"net/url"
@@ -47,8 +48,10 @@ func getTokenClaims(token *oidc.IDToken) (map[string]string, error) {
 
 // It makes string interpolation for a given string by using the
 // templates syntax https://pkg.go.dev/text/template
-// Issuer added as a parameter for having a richer log
-func applyTemplateOrReplace(extValueTemplate string, tokenClaims map[string]string, issuerMetadata map[string]string, issuer string) (string, error) {
+// logMetadata added as a parameter for having a richer log
+func applyTemplateOrReplace(
+	extValueTemplate string, tokenClaims map[string]string,
+	issuerMetadata map[string]string, logMetadata map[string]string) (string, error) {
 
 	// Here we merge the data from was claimed by the id token with the
 	// default data provided by the yaml file.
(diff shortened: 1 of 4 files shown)


pkg/identity/ciprovider/principal.go
```diff
@@ -135,10 +143,10 @@ func (principal ciPrincipal) Embed(_ context.Context, cert *x509.Certificate) er
 		s := v.Field(i).String() // value of each field, e.g the template string
 		// We check the field name to avoid to apply the template for the Issuer
 		// Issuer field should always come from the token issuer
-		if s == "" || vType.Field(i).Name == "Issuer" {
+		if strings.Trim(s, " ") == "" || vType.Field(i).Name == "Issuer" {
```
haydentherapper, line 146: Can you use `TrimSpace` instead, which will trim any whitespace characters as well?

pkg/identity/ciprovider/principal.go
```diff
@@ -115,7 +120,10 @@ func (principal ciPrincipal) Embed(_ context.Context, cert *x509.Certificate) er
 	if err != nil {
 		return err
 	}
-	subjectAlternativeName, err := applyTemplateOrReplace(principal.ClaimsMetadata.SubjectAlternativeNameTemplate, claims, defaults)
+	if strings.Trim(principal.ClaimsMetadata.SubjectAlternativeNameTemplate, " ") == "" {
```
haydentherapper, line 123: Same here, `TrimSpace`?

pkg/identity/ciprovider/principal.go
```diff
@@ -81,7 +82,7 @@ func applyTemplateOrReplace(extValueTemplate string, tokenClaims map[string]stri
 	}
 	claimValue, ok := mergedData[extValueTemplate]
 	if !ok {
-		return "", fmt.Errorf("value <%s> not present in either claims or defaults", extValueTemplate)
+		return "", fmt.Errorf("value <%s> not present in either claims or defaults. Issuer: %s", extValueTemplate, issuer)
```
haydentherapper, line 85: Any more logging we can do? What about which extension we're trying to write to?

javanlacerda: I was worried about logging any sensitive information, but I can add the extension for sure. wdyt?

haydentherapper: Yea, the name of the ext seems fine. It'll help us reproduce this once we know which issuer and which claim is causing the issue.

haydentherapper (approved):

PR #1776, opened 2024-08-19
If custom server url exists, use that instead of the default one.

kommendorkapten:
#### Summary

When producing certificates from ID Tokens issued by GitHub Actions, the server url is always hardcoded to `github.com`. This is not true for all cases, as custom domains may be possible. This PR modifies the behaviour to look for the `enterprise` claim which indicates that a custom domain is used.

#### Release Note

* When issuing certificates for GitHub Actions jobs, the server url is now properly created if custom domains are used.


trevrosen (approved):
:shipit:

pkg/identity/github/principal.go
```diff
@@ -159,10 +161,16 @@ func WorkflowPrincipalFromIDToken(_ context.Context, token *oidc.IDToken) (ident
 		return nil, errors.New("missing run_attempt claim in ID token")
 	}
 
+	baseURL := `https://github.com/`
+
+	if claims.Enterprise != "" {
```
haydentherapper, line 166: @javanlacerda Is this something we can add for the new configuration?

@kommendorkapten Is this something you want enabled for the public instance or is this just for the GitHub deployment?

javanlacerda: Yes, I think this template `{{if .enterprise }}https://{{ .enterprise }.ghe.com{{else}}{{ .url }}{{end}}` should replace every place that the default {{url}} is used for github on identity/config.yaml. 
Also, It could be defined in a variable for avoiding replicating it into the config.

And btw, I saw that this PR was recently merged, but this modifies will not work as we merged github to use the new generic logic that is in `pkg/identity/ciprovider/principal.go`

kommendorkapten: @haydentherapper this is only needed for our internal deployments as of now.

PR #1777, opened 2024-08-19
set empty string as defaults for codefresh's optional fields

### LINKED ISSUE #1770: Signing from Codefresh fails when scm_repo_url claim is not provided
**Description**

After the migration of OIDC providers to the new configuration in https://github.com/sigstore/fulcio/pull/1743/files
signing from Codefresh is failing when scm_repo_url is not provided.

This is the error received from Cosign - 
` Error: signing [ilmedcodefresh/cosign-build-test@sha256:9a14f704ed630d8a73246acf9085e68c9971636fe88f459f408dbaa4669bcad8]: getting signer: getting key from Fulcio: retrieving cert: POST https://fulcio.sigstore.dev/api/v1/signingCert returned 400 Bad Request: "{\"code\":3,\"message\":\"value <scm_repo_url> not present in either claims or defaults. {\\n\\t\\\"ExtensionName\\\": \\\"SourceRepositoryURI\\\",\\n\\t\\\"Issuer\\\": \\\"https://oidc.codefresh.io\\\"\\n}\",\"details\":[]}"`

Repository urls are not always present in the pipeline context in Codefresh - for example when a build is executed manually and not triggered from git.

When I added the Codefresh OIDC provider initially it was taken to consideration, but with the new configuration it seems like all claims are mandatory.

javanlacerda:
#### Summary
Closes #1770

It set empty string as default value for optional fields on codefresh workflow.

### DIFF
--- a/config/identity/config.yaml
+++ b/config/identity/config.yaml
@@ -203,6 +203,11 @@ ci-issuer-metadata:
       # for cases that the claimed data doesn't exist.
       # platform_url: Codefresh platform url
       platform_url: "https://g.codefresh.io"
+      scm_repo_url: ""
+      scm_ref: ""
+      runner_environment: ""
+      account_name: ""
+      pipeline_name: ""
     extension-templates:
       # workflow_id: The ID of the specific workflow authorized in the claim.
       # For example, 64f447c02199f903000gh20.
--- a/config/identity/config.yaml
+++ b/config/identity/config.yaml
@@ -206,8 +206,6 @@ ci-issuer-metadata:
       scm_repo_url: ""
       scm_ref: ""
       runner_environment: ""
-      account_name: ""
-      pipeline_name: ""
     extension-templates:
       # workflow_id: The ID of the specific workflow authorized in the claim.
       # For example, 64f447c02199f903000gh20.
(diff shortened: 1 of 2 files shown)


config/identity/config.yaml
```diff
@@ -203,6 +203,11 @@ ci-issuer-metadata:
       # for cases that the claimed data doesn't exist.
       # platform_url: Codefresh platform url
       platform_url: "https://g.codefresh.io"
+      scm_repo_url: ""
+      scm_ref: ""
+      runner_environment: ""
+      account_name: ""
```
haydentherapper, line 209: @ilia-medvedev-codefresh Could you confirm if `account_name` and `pipeline_name` are optional? Given those are a part of the [SAN](https://github.com/sigstore/fulcio/blob/77325fa27326bb1e6886811def8856f431f1998d/pkg/identity/codefresh/principal.go#L148), can we require they are present?

haydentherapper (dismissed):

config/identity/config.yaml
ilia-medvedev-codefresh: @haydentherapper Those are present by default in any token provided the Codefresh OIDC provider, so yes.

haydentherapper: Thanks, marked as required.

haydentherapper (approved):

bobcallaway (approved):

PR #1792, opened 2024-09-03
It creates a action for a nightly release

javanlacerda:
#### Summary
It implements a action for having a nightly release for Fulcio.
This is part of a project for having a continuous deployment of Fulcio for Public good instance staging.

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -1,0 +1,22 @@
+name: Nightly Release
+
+on:
+  schedule:
+    - cron: "0 0 * * *"  # Runs at midnight every day
+
+concurrency: nightly-release
+
+jobs:
+  nightly-release:
+    name: Nightly release
+    uses: sigstore/community/.github/workflows/reusable-release.yml@main
+    permissions:
+      id-token: write
+      contents: read
+    with:
+      release_tag: nightly-latest
+      key_ring: ${{ github.event.inputs.key_ring }}
+      key_name: ${{ github.event.inputs.key_name }}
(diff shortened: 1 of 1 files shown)


haydentherapper:
I didn't realize that we have CI to push a container to GCR whenever a commit is pushed to HEAD - https://github.com/sigstore/fulcio/blob/50abb65f0bdcb2c0290b3afb535bc77e339e1a19/.github/workflows/container-build.yml

Would this work instead? Do we need an easier way to discover these, would it help if we tagged these releases as well?

cc @cpanato

cpanato:
I would prefer we use the existing one, we will get rid from the cloudbuild soon as well

the existing images that we publish in github works for you?

haydentherapper:
Are we publishing at-HEAD containers to GHCR? From the GHA, it seems like we're only publishing to GCR (which for what it's worth, I think we should maintain given there's an SLO for GCR and not for GHCR)

cpanato:
> Are we publishing at-HEAD containers to GHCR? From the GHA, it seems like we're only publishing to GCR (which for what it's worth, I think we should maintain, given there's an SLO for GCR and not for GHCR)

ahh yeah that is correct.

we also can start publishing to ghcr is need

javanlacerda:
It seems perfect. As I noticed, the tag used is the last commit hash right?

javanlacerda:
I'm closing this PR as we decided to use the images created by the [container-build](https://github.com/sigstore/fulcio/blob/50abb65f0bdcb2c0290b3afb535bc77e339e1a19/.github/workflows/container-build.yml) workflow

PR #1796, opened 2024-09-04
upgrade to go1.23

cpanato:
- upgrade to go1.23

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.22.6@sha256:2bd56f00ff47baf33e64eae7996b65846c7cb5e0a46e0a882ef179fd89654afa AS builder
+FROM golang:1.23.1@sha256:4a3c2bcd243d3dbb7b15237eecb0792db3614900037998c2cd6a579c46888c1e AS builder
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sigstore/fulcio
 
-go 1.22.6
+go 1.23
```
haydentherapper, line 3: Is this fine without the specific version or toolchain? I would prefer this but not sure if go.mod requires otherwise.

cpanato:
@haydentherapper would you mind to take another look?

Dockerfile
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.22.6@sha256:2bd56f00ff47baf33e64eae7996b65846c7cb5e0a46e0a882ef179fd89654afa AS builder
+FROM golang:1.23.0@sha256:613a108a4a4b1dfb6923305db791a19d088f77632317cfc3446825c54fb862cd AS builder
```
haydentherapper, line 16: bump to 1.23.1?

cpanato: done PTAL

haydentherapper (approved):
thanks!

PR #1809, opened 2024-09-20
Adding breaking change label to container

javanlacerda:
#### Summary
It checks and adds the `breaking-change` label to the container that is created by the `build-container` action.
It also adds a label with the `commit-hash` to the container.

#### Documentation

Example of a build with this configuration:

image

### DIFF
--- a/.github/workflows/container-build.yml
+++ b/.github/workflows/container-build.yml
@@ -31,6 +31,7 @@ jobs:
     permissions:
       id-token: write
       contents: read
+      pull-requests: read
 
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
@@ -76,4 +77,4 @@ jobs:
       - name: container
         run: |
           echo "Formated Label: ${{ steps.labels.outputs.FORMATED_LABELS }}"
-          KO_PREFIX=gcr.io/projectsigstore/fulcio/ci/fulcio FORMATED_LABEL=${{ steps.labels.outputs.FORMATED_LABELS }} make sign-keyless-ci
+          KO_PREFIX=gcr.io/projectsigstore/fulcio/ci/fulcio FORMATED_LABEL=${{ steps.labels.outputs.FORMATED_LABELS }} RUN_NUMBER=${{ github.run_number }} make sign-keyless-ci
--- a/.github/workflows/container-build.yml
+++ b/.github/workflows/container-build.yml
@@ -77,4 +77,4 @@ jobs:
       - name: container
         run: |
           echo "Formated Label: ${{ steps.labels.outputs.FORMATED_LABELS }}"
-          KO_PREFIX=gcr.io/projectsigstore/fulcio/ci/fulcio FORMATED_LABEL=${{ steps.labels.outputs.FORMATED_LABELS }} RUN_NUMBER=${{ github.run_number }} make sign-keyless-ci
+          KO_PREFIX=gcr.io/projectsigstore/fulcio/ci/fulcio FORMATED_LABEL=${{ steps.labels.outputs.FORMATED_LABELS }} make sign-keyless-ci
--- a/.github/workflows/container-build.yml
+++ b/.github/workflows/container-build.yml
@@ -59,5 +59,21 @@ jobs:
       - name: creds
         run: gcloud auth configure-docker --quiet
 
+      - name: Formatted labels
+        id: labels
+        run: |
+          FORMATED_LABELS="--image-label commit-hash=$GITHUB_SHA"
+
+          BRANCH_NUMBER=$(gh pr list --state all --search "sha:$GITHUB_SHA" --label "breaking-change" | awk '{print $1}')
+          echo "Branch Number: $BRANCH_NUMBER" 
+
+          # Check if a pull request number was found
+          if [ -n "$BRANCH_NUMBER" ]; then
+            FORMATED_LABELS+=" --image-label breaking-change=true"
+          fi
+          echo "FORMATED_LABELS='$FORMATED_LABELS'" >> $GITHUB_OUTPUT
+
       - name: container
-        run: KO_PREFIX=gcr.io/projectsigstore/fulcio/ci/fulcio make sign-keyless-ci
+        run: |
+          echo "Formated Label: ${{ steps.labels.outputs.FORMATED_LABELS }}"
+          KO_PREFIX=gcr.io/projectsigstore/fulcio/ci/fulcio FORMATED_LABEL=${{ steps.labels.outputs.FORMATED_LABELS }} make sign-keyless-ci
--- a/.github/workflows/container-build.yml
+++ b/.github/workflows/container-build.yml
@@ -62,6 +62,8 @@ jobs:
 
       - name: Formatted labels
         id: labels
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: |
           FORMATED_LABELS="--image-label commit-hash=$GITHUB_SHA"
 
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,11 @@ GHCR_PREFIX ?= ghcr.io/sigstore
 FULCIO_YAML ?= fulcio-$(GIT_TAG).yaml
 
 # It should be blank for default builds
-FORMATED_LABEL =
+FORMATED_LABEL ?=
+
+RUN_NUMBER ?= "local"
+
+FULL_TAG := "0.$(shell date +%Y%m%d).$(RUN_NUMBER)+ref.$(GIT_HASH)"
 
 # Binaries
 PROTOC-GEN-GO := $(TOOLS_BIN_DIR)/protoc-gen-go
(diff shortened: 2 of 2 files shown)


.github/workflows/container-build.yml
```diff
@@ -59,5 +59,21 @@ jobs:
       - name: creds
         run: gcloud auth configure-docker --quiet
 
+      - name: Formatted labels
+        id: labels
+        run: |
+          FORMATED_LABELS="--image-label commit-hash=$GITHUB_SHA"
+
+          BRANCH_NUMBER=$(gh pr list --state all --search "sha:$GITHUB_SHA" --label "breaking-change" | awk '{print $1}')
```
haydentherapper, line 67: Can we check if we need to create a PAT for this or if the default token is fine? We sometimes hit rate limits when querying GitHub. I’d rather not manage another token but just wanna check. 

Cc @cpanato

haydentherapper: One other idea is to change this to fire on PR closure, like what’s documented in  https://stackoverflow.com/questions/71523153/github-actions-on-pull-request-merged-with-specific-labels  

You could then checkout from HEAD based on the git sha.

.github/workflows/container-build.yml
```diff
@@ -59,5 +59,21 @@ jobs:
...
+          # Check if a pull request number was found
+          if [ -n "$BRANCH_NUMBER" ]; then
+            FORMATED_LABELS+=" --image-label breaking-change=true"
+          fi
+          echo "FORMATED_LABELS='$FORMATED_LABELS'" >> $GITHUB_OUTPUT
+
       - name: container
-        run: KO_PREFIX=gcr.io/projectsigstore/fulcio/ci/fulcio make sign-keyless-ci
+        run: |
+          echo "Formated Label: ${{ steps.labels.outputs.FORMATED_LABELS }}"
```
haydentherapper, line 78: We could tackle this in the next PR, but I think we also need to handle labeling containers with merge times such that we can sort them and deploy only the latest commit. Otherwise if we merge two PRs at the same time, and the earlier GitHub action completes after the later GitHub action, then we would deploy an older commit.

.github/workflows/container-build.yml
haydentherapper, line 67: Thinking about how this will be used in PGI: Have you thought about how we will handle if multiple PRs are merged in a short period with only one labeled as a breaking change? Are we going to search back through all containers built after the last deployment?

haydentherapper: Eh, that might not work - commit shas will differ after being squashed to merge to HEAD.

javanlacerda: I think it should be fine as I tested this same action in my fork and it worked properly. https://github.com/javanlacerda/fulcio/actions/runs/10997017464/job/30531560844. Just hadn't the authorization to push the image, what is expected.

cpanato: i see this and will reply tomorrow (25/sept/2024) sorry !

.github/workflows/container-build.yml
javanlacerda: It would be good. but we can also look to the repo's commit history and check the order, right?

.github/workflows/container-build.yml
javanlacerda: Yes! The project that will deploy it will be responsible for detecting it.

javanlacerda: @cpanato could you take a look ?

cpanato: we need to add the permission  `pull-request: read`

cpanato: but this will also require that we add this label in the PRs, do we have a process documented for that?

Makefile
```diff
@@ -122,7 +122,7 @@ $(PROTOC-API-LINTER): $(TOOLS_DIR)/go.mod
 ko:
 	# fulcio
 	LDFLAGS="$(LDFLAGS)" GIT_HASH=$(GIT_HASH) GIT_VERSION=$(GIT_VERSION) \
-	KO_DOCKER_REPO=$(KO_PREFIX)/fulcio ko resolve --bare \
+	KO_DOCKER_REPO=$(KO_PREFIX)/fulcio ko resolve $(FORMATED_LABEL) --bare \
```
cpanato, line 125: should we define this variable in the makefile as well and maybe set as blank?

.github/workflows/container-build.yml
javanlacerda: We don't have a document that cover all the process, but this label usage started by the necessity to detect builds that requires adicional configuration on the helm charts. Basically, this breaking change label says that this image should not be deployed automatically and requires manual review, checking the configurations. We decided to use the `breaking-change` name because we can generalize it for having blocking auto deployment and requiring adicional review for any kind of blocker.

Makefile
javanlacerda: Yes! You're right! on it

.github/workflows/container-build.yml
jku: > We sometimes hit rate limits when querying GitHub

`gh pr list` does not use the "search" endpoint which I believe is the problematic one (since search rate limit is 30/minute instead of 5000/hour like others so temporary spikes can easily hit it) so I guess this is fine.

cpanato: if i run that locally it does not work, need to do a setup before

```
$ gh pr list --state all --search "sha:7920be28f7f58020c4da8ab9bb78554ad7cd67ea" --label "breaking-change"
X No default remote repository has been set for this directory.

please run `gh repo set-default` to select a default remote repository.
```

cpanato: ok, so maybe is a local thing, not related to the gh actions env. thanks for the clarification

.github/workflows/container-build.yml
```diff
@@ -59,5 +60,21 @@ jobs:
...
+          FORMATED_LABELS="--image-label commit-hash=$GITHUB_SHA"
+
+          BRANCH_NUMBER=$(gh pr list --state all --search "sha:$GITHUB_SHA" --label "breaking-change" | awk '{print $1}')
+          echo "Branch Number: $BRANCH_NUMBER" 
+
+          # Check if a pull request number was found
+          if [ -n "$BRANCH_NUMBER" ]; then
+            FORMATED_LABELS+=" --image-label breaking-change=true"
+          fi
+          echo "FORMATED_LABELS='$FORMATED_LABELS'" >> $GITHUB_OUTPUT
```
cpanato, line 75: need to set the env var

```
GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

cpanato: check your fork execution: https://github.com/javanlacerda/fulcio/actions/runs/11242081243/job/31254992474#step:7:16

javanlacerda: My fork execution failed because I removed the steps that setup the gcloud SDK, as I just wanted to test that the `ko` command was being created correctly, without uploading the images.  Here you can check the how my action definition looks like: https://github.com/javanlacerda/fulcio/commit/1c8adf3ac29767bfe91a0f1ab334766743c83638

javanlacerda: You're right! adding here

javanlacerda: done, thanks!!

.github/workflows/container-build.yml
javanlacerda: It already addressed. We can check it in the deployment side, checking the tags or commits that were created after the last deployment, and check that any of them has or not the`breaking-change` label.

javanlacerda: We are adding in this same PR a new tag for the containers that will help us to address it, allowing us to sort them correctly

javanlacerda:
Do you think we can merge it? @cpanato @haydentherapper

Makefile
```diff
@@ -46,6 +46,13 @@ GHCR_PREFIX ?= ghcr.io/sigstore
 
 FULCIO_YAML ?= fulcio-$(GIT_TAG).yaml
 
+# It should be blank for default builds
+FORMATED_LABEL ?=
+
+GITHUB_RUN_NUMBER ?= "local"
+
+FULL_TAG := "0.$(shell date +%Y%m%d).$(GITHUB_RUN_NUMBER)-ref.$(GIT_VERSION)"
```
haydentherapper, line 54: Should this include git sha or git version?

.github/workflows/container-build.yml
```diff
@@ -59,5 +60,23 @@ jobs:
       - name: creds
         run: gcloud auth configure-docker --quiet
 
+      - name: Formatted labels
+        id: labels
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          FORMATED_LABELS="--image-label commit-hash=$GITHUB_SHA"
+
```
haydentherapper, line 69: Do we need commit hash as a label? Looks like we include that as a tag already.

haydentherapper (commented):
Thanks for your work on this, this looks good! Just a couple small comments

@cpanato any remaining comments?

Makefile
cpanato: this when we release will be used, not sure if we want that format

.github/workflows/container-build.yml
cpanato: if that does not hurt, it is an extra thing :)

PR #1814, opened 2024-09-24
Update chainguard-dev/actions to latest commit

haydentherapper:
One of the commit shas referenced a 2 year old commit, which caused CI to fail since it used a deprecated action.


haydentherapper:
@cpanato Do you think you could add release tags for this repo so dependabot can pick up updates?

cpanato:
I will review that and see what we can do

cpanato (approved):

PR #1849, opened 2024-10-28
Sync proto

cpanato:


hack/tools/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sigstore/fulcio/hack/tools
 
-go 1.22.5
+go 1.23
```
bobcallaway, line 3: should we keep this synced with `go.mod` in project root dir?

cpanato: no strong options as this will get any patch release of go

haydentherapper: Is there general guidance on `go 1.x.y` vs `go 1.x`? We used to only do the latter, and then at some point, Go forced us to specify each patch release.

cpanato: that will say the minimum go version, if we pin to a patch for example go 1.22.8 it will not build with go1.22.7, but if we say go 1.22 it will work for all

but if one of the deps have a patch pinned i think that will come the toolchain that will force to use a pinned go version

that is my understanding

haydentherapper:
Already updated.

PR #1903, opened 2025-01-02
Configure additional certificate extensions for Buildkite

### LINKED ISSUE #1074: Update Buildkite issuer to include new extensions

yob:
#### Summary

The Buildkite Issuer was added in #890, prior to the efforts to standardise certificate extensions for CI providers, and #1074 calls for the Buildkite issuer to be updated to use the new extensions (where applicable).

This is an early attempt to make those changes. I initially started these in #1307, however this is a new swing at it using the new CIProvider issuer (see #1729 and #1743).

I've added the extensions that make the most sense in a Buildkite context, like RunInvocationURI, RunnerEnvironment and SourceRepositoryDigest. Many of the other extensions don't apply because we're not a code host as well, or need further discussion.

I have tested this locally with a Buildkite agent and OIDC token, and the certificate was issued as expected. I started a local fulcio like this:

    $ go run main.go serve --port 5555 --ca ephemeralca --ct-log-url="" --config-path config/identity/config.yaml

... and signed git commits with gitsign. The relevant bits of the certificates look like:

```
    git cat-file commit HEAD | sed -n '/-BEGIN/, /-END/p' | sed 's/^ //g' | sed 's/gpgsig //g' | sed 's/SIGNED MESSAGE/PKCS7/g' | openssl pkcs7 -print -print_certs -text
    ...
    X509v3 extensions:
        X509v3 Key Usage: critical
            Digital Signature
        X509v3 Extended Key Usage:
            Code Signing
            36:D2:99:B9:BA:98:4B:3A:77:51:DC:08:05:83:12:9A:F4:EE:41:E5
            D2:41:21:29:23:AD:E9:27:69:6F:DB:85:6D:1B:3F:7E:A9:55:F3:02
            URI:https://buildkite.com/yob-opensource/oidc-signing-experiment
        1.3.6.1.4.1.57264.1.1:
            https://agent.buildkite.com
        1.3.6.1.4.1.57264.1.8:
            ..https://agent.buildkite.com
        1.3.6.1.4.1.57264.1.11:
            ..self-hosted
        1.3.6.1.4.1.57264.1.13:
            .(078a6dd4a32fa40592c21a40aedaf27105503140
        1.3.6.1.4.1.57264.1.20:
            ..ui
        1.3.6.1.4.1.57264.1.21:
            .khttps://buildkite.com/yob-opensource/oidc-signing-experiment/builds/52#01943a38-f93e-4355-abe8-90a30369c270
```

Fixes #1074

cc @sj26 

#### Release Note

None.

#### Documentation

None


config/identity/config.yaml
```diff
@@ -236,4 +236,13 @@ ci-issuer-metadata:
   *buildkite-type:
     default-template-values:
       url: "https://buildkite.com"
-    subject-alternative-name-template: "{{.url}}/{{.organization_slug}}/{{.pipeline_slug}}"
+    extension-templates:
+      # Link to the specific Buildkite job that the OIDC token was generated from
+      run-invocation-uri: "{{.url}}/{{.organization_slug}}/{{.pipeline_slug}}/builds/{{.build_number}}#{{.job_id}}"
```
yob, line 241: The `build_number` claim in Buildkite OIDC tokens is an integer, so this is generating URLs like:

```
https://buildkite.com/yob-opensource/oidc-signing-experiment/builds/%!s(float64=42)#01942921-7883-409b-81d0-3f6b20bcdabf
```

Is there an existing way to have `42` formatted as a string nicely? I had a poke at `pkg/identity/ciprovider/principal.go`, but can't see anything relevant

config/identity/config.yaml
```diff
@@ -236,4 +236,13 @@ ci-issuer-metadata:
...
+      # TODO Was the job executed on Buildkite operated compute or customer hosted compute? (valid values: self-hosted, buildkite-hosted)
+      # runner-environment: "runner_environment"
```
yob, line 243: I'm working on getting this claim added to our tokens

config/identity/config.yaml
```diff
@@ -236,4 +236,13 @@ ci-issuer-metadata:
...
+      # The git sha that job was running, available in the `build_commit` claim
+      source-repository-digest: "build_commit"
+      # TODO build_source: Event that triggered this workflow run. (valid values: api, ui, webhook, trigger_job, schedule)
+      # build-trigger: "build_source"
```
yob, line 247: I'm working on getting this claim added to our tokens

yob:
@haydentherapper @feelepxyz I've had another swing at this. It's not mergable yet, but I'd value your feedback!

config/identity/config.yaml
haydentherapper: Good find. We can fix this in https://github.com/sigstore/fulcio/blob/main/pkg/identity/ciprovider/principal.go#L33-L39. Either change the formatting string to `%v`, or do type reflection and use `%d` when the type is an integer.

haydentherapper (commented):
No additional tests are necessary if this is working locally! We'll roll this out to our staging environment first, `fulcio.sigstage.dev`, for you to test against.

haydentherapper:
> A specific question I had while working on this: if Buildkite identity is now managed by the CIProvider type, can we rip out the Buildkite specific issuer - either in this PR or in a follow up? Or is it needed for backwards compatibility?

The Buildkite specific issuer should no longer be necessary. I was planning to remove all specific CI issuers as part of a 2.x release, waiting for backwards compatibility just in case someone has deployed a local instance.

> Also: would you like this PR to update the table in docs/oid-info.md ?

Yes please! Thank you.

yob:
Brilliant, thanks! I'll look at getting those two new claims into our tokens this coming week, then update this and mark it ready for review

yob:
@haydentherapper this is now as functional enough to be mergable from our perspective. The new extensions are working in my local testing (see updated PR description), and I've fixed the URL construction to handle the integer `build_number` claim.

I assume you might want a test for the changes in `mapValuesToString()`?  I'm also a bit weirded out that our integer `build_number` is a `float64` by the time it reaches `mapValuesToString()`. Maybe the real fix is to make that function just use `%v` with no reflection, and find a way to get the claim value passed around as the integer it is?

haydentherapper:
@yob I looked into this, the root cause is how the JSON unmarshaller works. [`token.Claims`])https://github.com/coreos/go-oidc/blob/4b5f82d880aeabd881766d2d91bee8086d9a3b92/oidc/oidc.go#L468-L473) is just a wrapper around `json.Unmarshal`. https://pkg.go.dev/encoding/json#Unmarshal notes that JSON numbers are unmarshalled to floats. You can initialize a `json.Decoder` with `UseNumber()` which will instead map JSON numbers to strings, but since the raw claims byte[] is not public, you can't use your own JSON decoder object.

The solution you have here seems sufficient. A small test would be great.

haydentherapper:
Oh, I also wanted to pass along a link for being added to PyPI's list of accepted trusted publishers - 
https://docs.pypi.org/trusted-publishers/internals/#how-do-i-become-a-trusted-publishing-provider

Once this PR is in and rolled out, it should be straightforward to get Buildkite added to PyPI.

yob:
neat, thanks! I've started chatting to the rubygems folks about their trusted publisher support too: https://github.com/rubygems/rubygems.org/issues/5377

yob:
@haydentherapper done!
> The solution you have here seems sufficient. A small test would be great.

I ended up tweaking the float->string logic a little, so floats with a fractional part don't get truncated into an int. That's not required for Buildkite tokens (we have no claims with a float value), but it felt like the least surprising behaviour to leave for future travelers.

I was tempted to adjust `TestWorkflowPrincipalFromIDToken` to have tables with standard GHA **and** Buildkite tokens as input, and verify the correct extensions come out - to ignore the implementation of the package a bit. That would've required more surgery (the existing GHA flavoured test is not setup to easily work with other providers) so  ended up testing the `getTokenClaims` function, which felt like a nice place to test the conversion of various token shapes to string claims.

Maybe we could consider the `TestWorkflowPrincipalFromIDToken` changes in a follow up test-only PR if you're interested?

haydentherapper (approved):
Sorry for the delay, this is great! Thanks for the contribution!

Following up on `TestWorkflowPrincipalFromIDToken` SGTM.

yob:
@haydentherapper great timing, I was just thinking this morning about following this PR up 😁 

I've just flipped the feature flag that will add the new `runner_environment` and `build_source` claims to all our tokens. Let me know if I can help with testing in staging!

haydentherapper:
Just getting a few other PRs merged, then I'll cut a release and ping you here once it's in staging.

haydentherapper:
@yob This should now be rolled out in prod too.

yob:
Great! I just took it for a spin - here's a rekor entry for a signed git commit: https://search.sigstore.dev/?logIndex=168731119

This is interesting:

```
Run Invocation URI: https://buildkite.com/yob-opensource/oidc-signing-experiment/builds/%!s(float64=93)#0194d306-40f2-4865-ad0e-15d225770518
```

Is it possible the config change has shipped, without the code change to handle numeric claim values?

haydentherapper:
Whoops that's exactly it, will fix in a sec!

haydentherapper:
Try again!

yob:
:cook: :kissing_heart:  https://search.sigstore.dev/?logIndex=168744984

```
https://buildkite.com/yob-opensource/oidc-signing-experiment/builds/95#0194d313-e659-4cc9-b4e8-d8e1f554986a
````

Thanks!

PR #1910, opened 2025-01-15
cleanup some zizmor-flagged issues in actions

bobcallaway:


.github/workflows/codeql-analysis.yml
```diff
@@ -24,36 +24,37 @@ on:
   schedule:
     - cron: '45 10 * * 1'
 
-permissions:
```
cpanato, line 27: can we set here

```
permissions: {}
```


then we drop all for the workflow and just set for the job?

cpanato (approved):
some are here :)

https://github.com/sigstore/fulcio/pull/1909

cpanato (approved):

Issue #1161, opened 2023-05-08
Add SignatureAlgorithm to template during certificate creation


haydentherapper:
**Description**

This is necessary to specify a non-default hash algorithm - See https://github.com/golang/go/blob/master/src/crypto/x509/x509.go#L1412 for more information. Otherwise, only SHA-256 will be used for RSA. 

We need a table from signer-hash to SignatureAlgorithm, https://github.com/golang/go/blob/master/src/crypto/x509/x509.go#L215-L233. There might be something else we can reuse from x509.go or maybe smallstep/crypto.

FYI @codysoyland @malancas

haydentherapper:
Something like:

```
func ToSignatureAlgorithm(signer crypto.Signer, hash crypto.Hash) (x509.SignatureAlgorithm, error) {
  pub := signer.Public()
  switch pub := pub.(type) {
    // only supports pkcs1.5, not PSS
    case *rsa.PublicKey:
      if hash == crypto.SHA256 {
        return x509.SHA256WithRSA, nil 
     }  
     // etc
    case *ecdsa.PublicKey:
    case ed25519.PublicKey:
      return x509.PureEd25519, nil
    default:
      return x509.UnknownSignatureAlgorithm, errors.New("unknown signer")
  }
}
```

Horiodino:
Hey @haydentherapper, I’d like to take over this from now on.

haydentherapper:
Go for it! 

The updated relevant code is https://github.com/golang/go/blob/608acff8479640b00c85371d91280b64f5ec9594/src/crypto/x509/x509.go#L1517-L1569

PR #1929, opened 2025-01-27
Add signature algorithm to template during certificate creation

Horiodino:
#### Summary
 Added ToSignatureAlgorithm to support specifying non-default hash algorithms during certificate creation

#1161

### DIFF
--- a/pkg/certmaker/certmaker.go
+++ b/pkg/certmaker/certmaker.go
@@ -20,17 +20,14 @@ package certmaker
 import (
 	"context"
 	"crypto"
-	"crypto/ecdsa"
-	"crypto/ed25519"
-	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
-	"errors"
 	"fmt"
 	"os"
 	"strings"
 	"time"
 
+	"github.com/sigstore/fulcio/pkg/ca"
 	"github.com/sigstore/sigstore/pkg/signature"
 	"github.com/sigstore/sigstore/pkg/signature/kms"
 	"go.step.sm/crypto/x509util"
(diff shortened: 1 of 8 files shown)


pkg/certmaker/certmaker.go
```diff
@@ -522,3 +541,47 @@ func ValidateKMSConfig(config KMSConfig) error {
 
 	return nil
 }
+
+// ToSignatureAlgorithm returns the x509.SignatureAlgorithm for the given signer and hash algorithm.
+func ToSignatureAlgorithm(signer crypto.Signer, hash crypto.Hash) (x509.SignatureAlgorithm, error) {
```
haydentherapper, line 546: Can you move this to https://github.com/sigstore/fulcio/blob/main/pkg/ca/common.go?

For context, `pkg/certmaker` was just added to support a new CLI utility, and I'd prefer to add new functions to it that are only used by the utility.

We should still keep the places above where you updated `tmpl.SignatureAlgorithm` though, that LGTM.

haydentherapper (commented):
This looks good overall! We need to also invoke it wherever we call `x509.CreateCertificate`, like in https://github.com/sigstore/fulcio/blob/260872d5d90c0591ba4fac6625124b2376ca2f3d/pkg/ca/baseca/baseca.go#L60 or the other CA implementations in `pkg/ca`.

haydentherapper:
This looks good but I need to test this out locally. /hold for now

haydentherapper:
Holding on this - this is intertwined with another active discussion on cryptographic agility and if we want to support non-standard variants.

haydentherapper:
Thanks for creating this. In another track of work, we’ve decided we won’t need this feature as we will limit fulcio to a set of algorithms that do not differ from the standard library defaults. I’ll close this out.

PR #1931, opened 2025-01-28
feat: makes leaf optional and other changes. 

### LINKED ISSUE #1930: make leaf optional in certificate-maker
relates to https://github.com/sigstore/fulcio/pull/1870 / https://github.com/sigstore/timestamp-authority/pull/889

**Description**

it's been recommended to make the certificate-maker more agnostic so that it may live in one place rather two with the  respective templates living w/ the service (e.g. fulcio and tsa).

the certificate-maker would allow both leaf and intermediate certs be options rather than the current behavior of leafs being mandatory.

- [Roots are always required, intermediates are optional, and leafs are optional (for a CA vs TSA).
](https://github.com/sigstore/timestamp-authority/pull/889#issuecomment-2613482618)

cc @haydentherapper

ianhundere:
closes #1930 
#### Summary

makes leaf optional as per @haydentherapper's request as well as updates docs etc.

### DIFF
--- a/pkg/certmaker/certmaker.go
+++ b/pkg/certmaker/certmaker.go
@@ -279,16 +279,6 @@ func CreateCertificates(sv signature.SignerVerifier, config KMSConfig,
 			return fmt.Errorf("error getting leaf public key: %w", err)
 		}
 
-		leafCryptoSV, ok := leafSV.(CryptoSignerVerifier)
-		if !ok {
-			return fmt.Errorf("leaf signer does not implement CryptoSigner")
-		}
-
-		_, _, err = leafCryptoSV.CryptoSigner(context.Background(), nil)
-		if err != nil {
-			return fmt.Errorf("error getting leaf crypto signer: %w", err)
-		}
-
 		var leafTemplate interface{}
 		if leafTemplatePath == "" {
 			defaultTemplate, err := GetDefaultTemplate("leaf")
--- a/pkg/certmaker/certmaker.go
+++ b/pkg/certmaker/certmaker.go
@@ -141,15 +141,15 @@ var InitKMS = func(ctx context.Context, config KMSConfig) (signature.SignerVerif
 }
 
 // CreateCertificates creates certificates using the provided KMS and templates.
-// It creates 3 certificates (root -> intermediate -> leaf) if intermediateKeyID is provided,
-// otherwise creates just 2 certs (root -> leaf).
+// Root certificate is always required.
+// Intermediate and leaf certificates are optional based on provided key IDs and templates.
 func CreateCertificates(sv signature.SignerVerifier, config KMSConfig,
 	rootTemplatePath, leafTemplatePath string,
 	rootCertPath, leafCertPath string,
 	intermediateKeyID, intermediateTemplatePath, intermediateCertPath string,
 	rootLifetime, intermediateLifetime, leafLifetime time.Duration) error {
 
-	// Create root cert
+	// Create root cert (required)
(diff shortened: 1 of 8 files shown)


ianhundere:
this is ready for 👀 

cc @haydentherapper

pkg/certmaker/certmaker.go
```diff
@@ -200,8 +203,8 @@ func CreateCertificates(sv signature.SignerVerifier, config KMSConfig,
 	var signingCert *x509.Certificate
 	var signingKey crypto.Signer
 
+	// Create intermediate cert (optional)
 	if intermediateKeyID != "" {
-		// Create intermediate cert if key ID is provided
 		intermediateConfig := config
 		intermediateConfig.RootKeyID = intermediateKeyID
```
haydentherapper, line 209: Are https://github.com/sigstore/fulcio/blob/c828ca3aeee1ced92212fe38cceb8de08e243ad8/pkg/certmaker/certmaker.go#L55-L56 used? It looks like like `InitKMS` only reads from `RootKeyID`. Should we update this field name to just `KeyID` and remove the unused fields?

pkg/certmaker/certmaker.go
```diff
@@ -262,58 +265,60 @@ func CreateCertificates(sv signature.SignerVerifier, config KMSConfig,
...
-	}
+		leafCryptoSV, ok := leafSV.(CryptoSignerVerifier)
+		if !ok {
+			return fmt.Errorf("leaf signer does not implement CryptoSigner")
+		}
 
-	var leafTemplate interface{}
-	if leafTemplatePath == "" {
-		defaultTemplate, err := GetDefaultTemplate("leaf")
+		_, _, err = leafCryptoSV.CryptoSigner(context.Background(), nil)
```
haydentherapper, line 287: is this used?

pkg/certmaker/certmaker.go
ianhundere: these fields are used in `CreateCertificates`  / while `InitKMS` only reads from `RootKeyID`, both `IntermediateKeyID` and `LeafKeyID` are copied to `RootKeyID` when initializing their respective KMS instances. 

the pattern uses the same `InitKMS` logic, but i think it helps maintain a clear separation between diff cert types. the fields are also used in validation logic for each KMS provider type.

haydentherapper: Ah, I see my confusion, there's an inconsistency with how these are passed in. You pass in `intermediateKeyID` as a function parameter to `CreateCertificates` and read `config.LeafKeyID` from the config.

I think it would be cleaner if `KMSConfig`/`InitKMS` had no notion of whether the KMS key was for a root/intermediate/leaf, as it's just initializing the KMS signer.

I would suggest passing in both `intermediate-key-id` and `leaf-key-id` as parameters to `CreateCertificates`, and having `KMSConfig` have only a `KeyID` field which is set for each invocation of `InitKMS`.

Also, I noticed we call InitKMS from [`certificate_maker.go`](https://github.com/sigstore/fulcio/blob/5ac0b1689d347e5546c16def3af26590d493aee3/cmd/certificate_maker/certificate_maker.go#L189). What do you think about moving that into `CreateCertificates` for parity with where other KMS signers are initialized?

ianhundere: ^ ah, that makes "a lot" more sense / thanks for the feedback, should have the changes up at some point this week.

haydentherapper (approved):
Looks great! Thanks for the updates!

Issue #1388, opened 2023-10-02
Allow configurable client signing algorithms


tetsuo-cpp:
**Description**

Hi there! At Trail of Bits, we're looking at potentially implementing part of the [Configurable Crypto Algorithms](https://docs.google.com/document/d/1-wF1t6lmJO37BzStbyz8ZkDhNm86HyvKZSaDl7uyHxI) proposal (specifically Phase 1). I wanted to float this idea to each of the relevant Sigstore sub-projects so we can hash out the details in a more concrete way.

Across the Sigstore stack, we default to using ECDSA for signatures and SHA256 for hashing. There's more detail in the linked proposal but there are a number of motivations for wanting to customise the signatures that are generated, including paving the way for post-quantum signatures. The proposed design includes having a "supported algorithm" registry (perhaps this can go in the Protobuf specs) that outlines enumerates the approved signature/hash algorithm combinations. We specifically don't want to allow arbitrary mixing and matching of signature and hash algorithm to avoid some of the security pitfalls listed in the proposal.

I'm not super familiar with the Fulcio side of things but I expect that there's a few places where this ECDSA-SHA256 assumption is hardcoded, so we'll need to make those more flexible. We can also add a flag like `--client-signing-algorithms=alg-1,alg-2` that can be used in case a given Fulcio instance wants to constrain the list of supported algorithms to be more restrictive than the aforementioned registry.

haydentherapper:
There aren't many places where the signing algorithm is hardcoded. To give a quick overview, there's a few CA "backends" that implement signing logic given a key in some format - KMS, Tink, in-memory, etc. For each backend:

* KMS - This uses a cloud KMS to sign certificates. The signing algorithm and hash are provided from the sigstore/sigstore library. Once a KMS provider has added support for PQ signing algorithms, we'll just need to add that to s/s.
* Tink - In-memory signing where the signing key is encrypted with a KMS key. Tink will need to be updated to support PQ algorithms (which I assume is on their roadmap)
* File-based - Meant as a test CA since keys stored on disk (even password-protected) aren't ideal. This supports any signing key type supported in sigstore/sigstore.
* Ephemeral - This is a test CA which hardcodes ECDSA P-256 where the signing key lives in-memory. We can easily make this configurable.

[...]

PR #1938, opened 2025-02-05
Allow configurable client signing algorithms

ret2libc:
#### Summary
This PR adds a --client-signing-algorithms flag to Fulcio to restrict what key/hash combinations are allowed.

Closes https://github.com/sigstore/fulcio/issues/1388
This is based on https://github.com/sigstore/fulcio/pull/1517, but I could not mark it as Ready because done by @tetsuo-cpp .

### DIFF
--- a/cmd/app/serve.go
+++ b/cmd/app/serve.go
@@ -120,6 +120,9 @@ func newServeCmd() *cobra.Command {
 		v1.PublicKeyDetails_PKIX_ECDSA_P256_SHA_256,
 		v1.PublicKeyDetails_PKIX_ECDSA_P384_SHA_384,
 		v1.PublicKeyDetails_PKIX_ECDSA_P521_SHA_512,
+		v1.PublicKeyDetails_PKIX_RSA_PKCS1V15_2048_SHA256,
+		v1.PublicKeyDetails_PKIX_RSA_PKCS1V15_3072_SHA256,
+		v1.PublicKeyDetails_PKIX_RSA_PKCS1V15_4096_SHA256,
 		v1.PublicKeyDetails_PKIX_ED25519,
 	}), "the list of allowed client signing algorithms")
 
--- a/cmd/app/serve.go
+++ b/cmd/app/serve.go
@@ -33,6 +33,9 @@ import (
 	"syscall"
 	"time"
 
+	v1 "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
+	"github.com/sigstore/sigstore/pkg/signature"
+
 	"chainguard.dev/go-grpc-kit/pkg/duplex"
 	"github.com/goadesign/goa/grpc/middleware"
 	ctclient "github.com/google/certificate-transparency-go/client"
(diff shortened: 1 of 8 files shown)


ret2libc:
cc @haydentherapper

cmd/app/serve.go
```diff
@@ -113,6 +116,12 @@ func newServeCmd() *cobra.Command {
 	cmd.Flags().String("grpc-tls-key", "", "the private key file to use for secure connections (without passphrase) - only applies to grpc-port")
 	cmd.Flags().Duration("idle-connection-timeout", 30*time.Second, "The time allowed for connections (HTTP or gRPC) to go idle before being closed by the server")
 	cmd.Flags().String("ct-log.tls-ca-cert", "", "Path to TLS CA certificate used to connect to ct-log")
+	cmd.Flags().StringSlice("client-signing-algorithms", buildDefaultClientSigningAlgorithms([]v1.PublicKeyDetails{
+		v1.PublicKeyDetails_PKIX_ECDSA_P256_SHA_256,
```
haydentherapper, line 120: For the default, can we include the three RSA-PKCS1v15 algorithms as well, to match what Fulcio currently supports? (The public API says it supports PSS, but it doesn't...)

cmd/app/serve.go
```diff
@@ -437,7 +460,7 @@ func StartDuplexServer(ctx context.Context, cfg *config.FulcioConfig, ctClient *
 	)
 
 	// GRPC server
-	grpcCAServer := server.NewGRPCCAServer(ctClient, baseca, ip)
+	grpcCAServer := server.NewGRPCCAServer(ctClient, baseca, algorithmRegistry, ip)
 	protobuf.RegisterCAServer(d.Server, grpcCAServer)
 	if err := d.RegisterHandler(ctx, protobuf.RegisterCAHandlerFromEndpoint); err != nil {
 		return fmt.Errorf("registering grpc ca handler: %w", err)
```
haydentherapper, line 466: Can we either add a comment that the legacy path (`NewLegacyGRPCCAServer) will not be updated to support this feature, or if it's not much work, can we add it? I'm OK with either approach, with a slight preference for uniformity between the two if it's not much work.

ret2libc: Maybe I'm missing something, but I see that the legacy grpc server creates a v2 request from the legacy request (https://github.com/sigstore/fulcio/blob/main/pkg/server/legacy_server.go#L100) and then the new logic takes place anyway. So this is automatically supported on the legacy grpc server as well.

Are you referring to any edge case I'm not considering?

pkg/server/grpc_server.go
```diff
@@ -133,6 +145,22 @@ func (g *grpcaCAServer) CreateSigningCertificate(ctx context.Context, request *f
 		if err := challenges.CheckSignature(publicKey, proofOfPossession, principal.Name(ctx)); err != nil {
 			return nil, handleFulcioGRPCError(ctx, codes.InvalidArgument, err, invalidSignature)
 		}
+
+		// The proof of possession signature always uses SHA-256, unless the key algorithm is ED25519
+		hashFunc = crypto.SHA256
+		if _, ok := publicKey.(ed25519.PublicKey); ok {
+			hashFunc = crypto.Hash(0)
+		}
```
ret2libc, line 153: I was testing the Cosign changes and I've realized this is not good in general.

Something like this seems to work:
```golang
		switch pk := publicKey.(type) {
		case ed25519.PublicKey:
			// Fulcio only works with PureEd25519
			hashFunc = crypto.Hash(0)
		case *ecdsa.PublicKey:
			switch pk.Curve {
			case elliptic.P256():
				hashFunc = crypto.SHA256
			case elliptic.P384():
				hashFunc = crypto.SHA384
			case elliptic.P521():
				hashFunc = crypto.SHA512
			default:
				hashFunc = crypto.SHA256
			}
		case *rsa.PublicKey:
			hashFunc = crypto.SHA256
		default:
			hashFunc = crypto.SHA256
		}

		// Check proof of possession signature
		if err := challenges.CheckSignatureWithOpts(publicKey, proofOfPossession, principal.Name(ctx), options.WithHash(hashFunc)); err != nil {
		}
```

@haydentherapper what do you think about this?

haydentherapper: Is this a breaking change? Currently, it's assumed that the digest is always [SHA256](https://github.com/sigstore/fulcio/blob/8e982d3bc16dc2118d1ef0ed9f53e80c7b063e6f/pkg/challenges/challenges.go#L47) regardless of key type.

Can we work around this and note this as something to correct in a major version bump?

What about other clients as well? -python for example, is it always using sha256? Or has it just so happened that we didn't notice this because only RSA and ECDSA-P256 keys have been used?

haydentherapper: Tracking for Fulcio v2 in https://github.com/sigstore/fulcio/issues/1945.

haydentherapper:
Thank you!

Issue #1953, opened 2025-02-24
Algorithms that should not use SHA256 are not permitted


ret2libc:
**Description**

With https://github.com/sigstore/fulcio/commit/06e94fce5fd4a2d4580382641d0f0a8afdd572f4 we introduced a validation of the algorithms that Fulcio accepts https://github.com/sigstore/fulcio/blob/21f246a1c88a2c729c324e11dceb2586ee036132/pkg/server/grpc_server.go#L156-L164

However, we are also enforcing SHA256 (except for ed25519) because that's what all clients are using nowadays and we don't want to break compatibility with them until Fulcio v2 is done (see https://github.com/sigstore/fulcio/issues/1945). https://github.com/sigstore/fulcio/blob/21f246a1c88a2c729c324e11dceb2586ee036132/pkg/server/grpc_server.go#L149-L153

This means that when you specify ecdsa-p384-sha384 on the client side (e.g. cosign), the fulcio requests should be made with SHA256, however ecdsa-p384 + sha256 is not a permitted algorithm, thus `g.algorithmRegistry.IsAlgorithmPermitted(publicKey, hashFunc)` is going to fail and reject the certificate request. This behaviour is a regression, since before there was no check and the certificate request was performed.

**Version**
`21f246a1c88a2c729c324e11dceb2586ee036132`.

ret2libc:
I think the problem is only for `ecdsa-sha2-384-nistp384` and `ecdsa-sha2-512-nistp521`, because the RSA options only use SHA256 and the ED25519 options are already handled.

Possible solutions:
1) add some "custom" options in `PublicKeyDetails` like `PKIX_ECDSA_P384_SHA_256` and `PKIX_ECDSA_P521_SHA_256` and add those entries to the default list of accepted algorithms in Fulcio. In this way, the algorithms will be accepted and they can be removed at some point in the future.
2) add some special logic in `grpc_server.go`, similar to what we do for `ed25519` already. In other words, if we detect a ecdsa-p384/ecdsa-p512 public key, we set the hash to the correct value. 

In any case, the signature is checked with sha256 in `CheckSignature`, so this would be only for making the algorithmRegistry checks pass for compatibility reasons.

haydentherapper:
Good catch. What do you think about instead treating the PoP as the special case and pulling the digest algorithm based on the signature algorithm, like what you had originally proposed in https://github.com/sigstore/fulcio/pull/1938#discussion_r1952976376?

[...]

PR #1959, opened 2025-03-04
Proof of Possession agility

ret2libc:
#### Summary
Closes https://github.com/sigstore/fulcio/issues/1953

#### Release Note
* Proof of possession signatures now need to be signed with the "default" hash associated with a given public key type. The defaults comes from sigstore/sigstore (see https://github.com/sigstore/sigstore/pull/2014)

#### Documentation
This is a breaking change because before all algorithms used SHA256 (except ED25519 which cannot work with it). Now, based on the public key type, a different hash algorithm might be used. In practice, only ECDSA/P384 and ECDSA/P521 are going to be affected by this change because before they were using SHA256 and with this patch they are going to use SHA384 and SHA512.

Considering ECDSA/P256 is the de-facto standard so far for sigstore clients, most users won't notice any difference. However, Cosign supports BYOK, so some users might use ECDSA/P384 (or P521) keys to request Fulcio certificates. Cosign needs to be updated to use the correct hash algorithm based on the key.


haydentherapper (commented):
LGTM pending sigstore/sigstore PR being merged, and pending the discussion about ed25519 defaults.

ret2libc:
TBH I'm a bit unsure what's the problem with the golangci-lint. I don't have any issue locally and those reported issues seems to be false positives to me.

haydentherapper (approved):
Thanks!

PR #1968, opened 2025-03-13
Change grpc_prometheus dependency source

loosebazooka:
Old source has been deprecated and we're expected to pull the dependency from go-grpc-middleware now

#### Summary
See: https://github.com/grpc-ecosystem/go-grpc-prometheus?tab=readme-ov-file#deprecated-go-grpc-interceptors-for-prometheus-monitoring

#### Release Note
NONE

#### Documentation
NONE


pkg/server/metrics.go
```diff
@@ -22,22 +22,30 @@ import (
...
 
 var (
-	metricNewEntries = promauto.NewCounter(prometheus.CounterOpts{
+	metricNewEntries prometheus.Counter
+	MetricLatency    *prometheus.HistogramVec
+	RequestsCount    *prometheus.CounterVec
+)
+
+// InitMetrics should only be called once
+func InitMetrics(reg prometheus.Registerer) {
```
haydentherapper, line 31: Should we use https://pkg.go.dev/sync#OnceFunc to block against multiple inits?

loosebazooka:
Things I've discovered:
- The duplex server is still pulling in an old version of the prometheus go library
- All of our code expects the default server metrics and default registerer
- I think I can kinda fix this but wont be able to pull out the old library completely until duplex server is fixed

PR #1988, opened 2025-04-07
Use localhost address for internal gRPC client

### LINKED ISSUE #1987: Internal gRPC client connections using 0.0.0.0 fail in proxy environments
**Description**
Fulcio fails when deployed in an environment with an HTTP proxy enabled, where REST API calls trigger internal gRPC client connections to `0.0.0.0:5554`. This address is suitable for server binding but invalid for client connections.

In environments using HTTP proxies, requests to `0.0.0.0` typically bypass `NO_PROXY` settings and are incorrectly routed through the proxy. This causes the proxy to attempt a connection to `127.0.0.1:5554` on its own host, resulting in failures and breaking the expected internal communication flow.

**Reproducer**

1. Reproducer source code
 https://github.com/osmman/fulcio/commit/5f8e8fc29a4127c5cdd55146e7b57d70b85f87cc
2. Deploy infra 
`docker-compose up`
3. Trigger scenario by accessing the following URL 
https://127.0.0.1:5555/api/v2/trustBundle

**Expected Behavior**
Internal gRPC clients should connect to `127.0.0.1:5554` instead of `0.0.0.0:5554`, ensuring that the request bypasses the proxy via standard `NO_PROXY` rules.

**Version**
1.6.6

osmman:
#### Summary
This pull request resolves an issue where internal REST API calls in Fulcio would fail in proxy-enabled environments due to internal gRPC clients connecting to `0.0.0.0:5554`. This address, while valid for server binding, is not appropriate for client connections and is not filtered by `NO_PROXY` settings. As a result, requests were routed through the external proxy and failed. This change updates the client connection address to `localhost`, ensuring requests are properly excluded from proxy routing.

Closes #1987

#### Release Note
Fixed internal gRPC client to use `localhost` instead of `0.0.0.0`, avoiding proxy interference in REST-to-gRPC calls.

#### Documentation
No documentation updates are required for this change.

### DIFF
--- a/cmd/app/http.go
+++ b/cmd/app/http.go
@@ -20,6 +20,7 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"os/signal"
@@ -63,8 +64,14 @@ func createHTTPServer(ctx context.Context, serverEndpoint string, grpcServer, le
 	} else {
 		opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	}
-	grpcHealthEndpoint := fmt.Sprintf("localhost:%s", viper.GetString("grpc-port"))
-	cc, err := grpc.NewClient(grpcHealthEndpoint, opts...)
+
+	_, port, err := net.SplitHostPort(grpcServer.grpcServerEndpoint)
+	if err != nil {
+		log.Logger.Fatal(err)
+	}
+
+	grpcServerEndpoint := fmt.Sprintf("localhost:%s", port)
+	cc, err := grpc.NewClient(grpcServerEndpoint, opts...)
 	if err != nil {
 		log.Logger.Fatal(err)
 	}
@@ -73,7 +80,7 @@ func createHTTPServer(ctx context.Context, serverEndpoint string, grpcServer, le
 		runtime.WithForwardResponseOption(setResponseCodeModifier),
 		runtime.WithHealthzEndpoint(health.NewHealthClient(cc)))
 
-	if err := gw.RegisterCAHandlerFromEndpoint(ctx, mux, grpcServer.grpcServerEndpoint, opts); err != nil {
+	if err := gw.RegisterCAHandlerFromEndpoint(ctx, mux, grpcServerEndpoint, opts); err != nil {
 		log.Logger.Fatal(err)
 	}
 


cmd/app/http.go
```diff
@@ -63,8 +64,14 @@ func createHTTPServer(ctx context.Context, serverEndpoint string, grpcServer, le
...
 	}
-	grpcHealthEndpoint := fmt.Sprintf("localhost:%s", viper.GetString("grpc-port"))
-	cc, err := grpc.NewClient(grpcHealthEndpoint, opts...)
+
+	_, port, err := net.SplitHostPort(grpcServer.grpcServerEndpoint)
+	if err != nil {
+		log.Logger.Fatal(err)
+	}
+
+	grpcServerEndpoint := fmt.Sprintf("localhost:%s", port)
```
bobcallaway, line 73: I'm not following how this is substantially different from the current implementation given that `grpcServer.grpcServerEndpoint` is set at https://github.com/sigstore/fulcio/blob/7342c9725380764eb751bcea81b0aacedc432eaa/cmd/app/grpc.go#L197

Can you provide a bit more clarity here?

osmman: You're right that grpcServerEndpoint is initially set via `fmt.Sprintf("%s:%s", viper.GetString("grpc-host"), viper.GetString("grpc-port"))`. However, I can't rely on `viper.GetString("grpc-port")` later on in the code, because it may be set to `0`, especially in testing scenarios.

For example, in [http_test.go#L50–L61](https://github.com/sigstore/fulcio/blob/main/cmd/app/http_test.go#L50-L61), we explicitly pass `--grpc-port=0` to allow the system to bind to a random available port.

Since Viper will still return "0" in this case, I need to get the actual assigned port another way. That’s why I parse it from the resolved `grpcServer.grpcServerEndpoint` using `net.SplitHostPort`, which contains the effective bind address once it’s known.

Then I reconstruct the endpoint as `localhost:<actual-port>` to make sure we're rebinding locally with the correct port. This avoids relying on config values that may no longer reflect the runtime state of the server.

bobcallaway: ok; I can see at https://github.com/sigstore/fulcio/blob/4b34341ffdea3c2027ea71b53af3da56eaf66357/cmd/app/grpc.go#L215 where the request for an ephemeral port would be resolved... but TBH the message in the above PR doesn't really map up to this fix? 

Also, is there a simple way to add testing for this?

osmman: The main goal of this fix is to resolve issue #1987 by ensuring internal gRPC clients connect to `localhost:<port>` instead of `0.0.0.0:<port>`. Using `0.0.0.0` breaks in proxy-enabled environments, as it's not excluded by `NO_PROXY`.

The dynamic port part isn’t new. Fulcio already behaves this way, especially in tests. I used `grpcServer.grpcServerEndpoint` to get the actual bound port because relying on `viper.GetString("grpc-port")` can return "0".

For testing, I verified proxy problem manually by using a modified docker-compose setup ([commit](https://github.com/sigstore/fulcio/commit/c94d65ae64b86dd8d92a452295a1c512e04fcb25)). Not sure where this kind of test best fits in Fulcio — open to ideas.

bobcallaway: I think you mean it is trying to connect to `localhost:0` instead of `localhost:port` (which I agree is a bug)... but the linkage to `NO_PROXY` still isn't clear to me at all.

bobcallaway (approved):

PR #2014, opened 2025-04-23
Update docker compose deps, fix for macOS

haydentherapper:
Updates to the latest Trillian deps. Also sets the platform explicitly for the MySQL container, which will otherwise fail to startup on macOS since it's looking for ARM rather than emulating AMD64.


haydentherapper:
FYI @ramonpetgrave64 for getting fulcio working on macOS

Issue #2015, opened 2025-04-23
Migrate away from go-yaml


codysoyland:
**Description**

[go-yaml](https://github.com/go-yaml/yaml) is no longer maintained and the author archived the repo. We should consider switching to a different yaml package.

Some alternatives:
- https://github.com/goccy/go-yaml
- https://github.com/kubernetes-sigs/yaml

PR #2016, opened 2025-04-23
Switch yaml parser to goccy/go-yaml

codysoyland:
#### Summary

As [go-yaml/yaml](https://github.com/go-yaml/yaml) is no longer maintained, this PR switches the yaml parser to [goccy/go-yaml](https://github.com/goccy/go-yaml).

Fixes https://github.com/sigstore/fulcio/issues/2015


haydentherapper:
Does `define` not work anymore?

config/identity/config.yaml
```diff
@@ -12,11 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-define:
-  - &github-type "github-workflow"
-  - &gitlab-type "gitlab-pipeline"
-  - &codefresh-type "codefresh-workflow"
-  - &buildkite-type "buildkite-job"
```
codysoyland, line 19: Note that some tests were failing due to parse errors related to these anchors. If we merge this, we should also check public-good-instance/scaffolding for usage of yaml anchors.

haydentherapper:
In https://github.com/sigstore/timestamp-authority/pull/1049 we used a differently library, which only requires `json` rather than `yaml` tags. Thoughts on using this instead?

codysoyland:
I don't have a preference on library. Would be good if has better support for anchors. I'll investigate that lib.

codysoyland:
I tried using `sigs.k8s.io/yaml` but it's not really compatible with `FulcioConfig` because there are [different key names](https://github.com/sigstore/fulcio/blob/main/pkg/config/config.go#L54) for JSON config files.

haydentherapper:
Ah yes. I’m good with this library, though would prefer to not have a breaking change of anchors not being supported, but it’s too big of a deal.

haydentherapper:
Will close out for now and revisit later.

PR #2023, opened 2025-04-23
Do not HTML-escape extension values

haydentherapper:
This fixes a bug report where extensions contained HTML-escaped characters, e.g. &lt; rather than <. The root cause and the fix is simple: instead of using html/template, we should use text/template which does not escape any characters.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -22,13 +22,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


woodruffw (approved):
Thanks a ton!

haydentherapper:
Also LOL at the description, I typed `&lt;`

bobcallaway (approved):

PR #2111, opened 2025-07-22
fix: K8s API does not accept unauthorized requests

### LINKED ISSUE #2110: Faile to verify against kubernetes.default.svc
**Description**
Following the security hardening implemented in Kubernetes via pull request [#38708](https://github.com/kubernetes/kubernetes/pull/38708), all unauthenticated HTTP requests are now denied. This change prevents Fulcio from successfully retrieving JSON Web Tokens (JWTs) from the Kubernetes API, leading to authentication failures.

Fulcio is unable to verify signatures, resulting in the following error:
```
,"message":"failed to verify signature: fetching keys oidc: get keys failed: 403 Forbidden {\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"forbidden: User \\\"system:anonymous\\\" cannot get path \\\"/openid/v1/jwks\\\"\",\"reason\":\"Forbidden\",\"details\":{},\"code\":403}\n","requestID":"rB30_-IW","code":"InvalidArgument","clientMessage":"There was an error processing the identity token","error":"failed to verify signature: fetching keys oidc: get keys failed: 403 Forbidden 
```
This error indicates that the `system:anonymous` user is forbidden from accessing the /openid/v1/jwks path, which is necessary for Fulcio to fetch OIDC keys and verify identity token signatures.

bouskaJ:
#### Summary
Fixes #2110 
#### Release Note
* The cluster token is now automatically passed to the authentication header for the `kubernetes.default.svc OIDC` provider


haydentherapper:
Hey, thanks for the PR. Do you have more information about this issue? The PR linked in the issue was from almost 10 years ago so I'm uncertain why this is now coming up.

I also am not too familiar with Kubernetes setups, so can you confirm if `k8sTokenFile` is always present? Is whether or not the token is passed something that a deployer can configure?

bouskaJ:
Hi @haydentherapper thank you for the review. I can confirm that the token file is always present (as you can see in the Kubernetes doc https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/#directly-accessing-the-rest-api).

For more information about this issue, you can read https://www.redhat.com/en/blog/what-you-need-to-know-red-hat-openshift-416
Especially following paragraph is important for this context
```
Reduce unauthenticated user or group access
Beginning with OpenShift 4.16, there's a limit on the permissions given to the system:anonymous user and system:unauthenticated group. Only two roles are allowed by default:

system:openshift:public-info-viewer: required for OIDC workflows
system:public-info-viewer: read-only access to non-sensitive information about the cluster
This applies to new clusters only. Clusters that are upgraded are not affected.

For use cases where anonymous access is needed, the cluster administrator must explicitly add those permissions. For example, if you're using webhooks for BuildConfigs from an external system (like GitHub) that don't support sending auth tokens over HTTP, then you'd need to explicitly add system:webhook permissions. We strongly recommend using local rolebindings instead of clusterrole bindings in such scenarios. A cluster administrator can add back a ClusterRoleBinding for the anonymous user as needed, after evaluating the risks and benefits associated with it
```

pkg/config/config.go
```diff
@@ -292,55 +303,99 @@ func httpClientForIssuer(iss OIDCIssuer) (*http.Client, error) {
...
+		if err := fc.insertVerifier(iss); err != nil {
+			log.Logger.Errorf("error creating provider for issuer URL %q: %v", iss.IssuerURL, err)
+			continue
+		}
+	}
+
+	_, verifierPresent := fc.verifiers[k8sIssuerURL]
+	k8sIssuer, issuerConfigured := fc.GetIssuer(k8sIssuerURL)
+	if issuerConfigured && !verifierPresent {
+		// configure static validator for k8s that cover metaIssuers
```
cmurphy, line 317: Can you explain more why the k8s issuer would be present in the config but not already have its provider/verifier created in the loop above? This check doesn't look necessary to me.

pkg/config/config.go
```diff
@@ -292,55 +303,99 @@ func httpClientForIssuer(iss OIDCIssuer) (*http.Client, error) {
...
+
+	cache, err := lru.New2Q[string, []*verifierWithConfig](100 /* size */)
+	if err != nil {
+		return fmt.Errorf("lru: %w", err)
+	}
+	fc.lru = cache
+	return nil
+}
+
+var (
```
cmurphy, line 331: These could all be consts.

pkg/config/config.go
```diff
@@ -292,55 +303,99 @@ func httpClientForIssuer(iss OIDCIssuer) (*http.Client, error) {
...
-		client, err := httpClientForIssuer(iss)
-		if err != nil {
-			log.Logger.Errorf("error creating http client for issuer URL %q: %v", iss.IssuerURL, err)
-			continue
+		if _, err := os.Stat(k8sTokenFile); err != nil {
+			if errors.Is(err, os.ErrNotExist) {
+				log.Logger.Warnf("Kubernetes token file can't be found on path: %s", k8sTokenFile)
+			} else {
+				return err
+			}
```
cmurphy, line 368: (nit, Go style)
```suggestion
			}
			return err
```

pkg/config/config.go
bouskaJ: Yes, they could, but I make them `vars` because it allows me to change the value for testing purposes. https://github.com/bouskaJ/fulcio/blob/fix_2110/pkg/config/config_test.go#L905

pkg/config/config_test.go
```diff
@@ -717,6 +720,212 @@ func TestVerifierCache(t *testing.T) {
...
+				MetaIssuers: map[string]OIDCIssuer{
+					metaIssuer: {
+						ClientID: "sigstore",
+						Type:     IssuerTypeKubernetes,
+					},
+				},
+			},
+		},
+		{
+			name:          "Anauthorized",
```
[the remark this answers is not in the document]
bouskaJ: Assuming the k8s issuer is always an `OIDCIssuer` and never a `MetaIssuer ` significantly simplifies the process. I've now removed the confusing logic.

pkg/config/config.go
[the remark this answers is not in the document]
bouskaJ: I have removed the metaIssuer support based on https://github.com/sigstore/fulcio/pull/2111#discussion_r2331080146, but you are right. httpClientForIssuer seems like a much better place to solve this. :+1:

PR #2117, opened 2025-07-30
fix: vault now only expect the key name / sigstore library constructs…

ianhundere:
… the full path.


haydentherapper (approved):

haydentherapper (commented):
Can you revert the protobuf file changes? I’ll guess that your local environment has a different version of the protobuf compiler installed.

haydentherapper (approved):

haydentherapper:
Thank you!

PR #2147, opened 2025-09-10
server: Use warning log level for client errors

### LINKED ISSUE #908: Every failed request is logged as an error
See https://github.com/sigstore/fulcio/blob/main/pkg/server/error.go#L42, all request errors are logged as errors regardless of if it's due to the user (4xx) or server (429, 5xx). This leads to a lot of errors in the logs that are due to malformed JWTs.

We should differentiate between user errors and server errors. User errors should be at most WARN, but for JWT errors, I would remove logging entirely since there's nothing actionable to do.

jku:
Decide the log level of errors based on the grpc code: client mistakes should not be logged at "error" level.

Fulcio itself currently uses only Internal and InvalidArgument, grpc framework does not use most of the codes listed either: this code still tries to categorize all valid codes in order to be future proof.

Fixes #908.

### test

this client call:
```
$ curl -X POST -d '{}' http://localhost:5555/api/v2/signingCert
{"code":3,"message":"There was an error processing the identity token","details":[]}
```
results in:
```
fulcio-server-1        | 2025-09-10T09:39:58.799Z       WARN    server/error.go:57      oidc: malformed jwt, expected 3 parts got 1     {"requestID": "-03xA5aM", "code": "InvalidArgument", "clientMessage": "There was an error processing the identity token", "error": "oidc: malformed jwt, expected 3 parts got 1"}
fulcio-server-1        | github.com/sigstore/fulcio/pkg/server.handleFulcioGRPCError
fulcio-server-1        |        /opt/app-root/src/pkg/server/error.go:57
fulcio-server-1        | github.com/sigstore/fulcio/pkg/server.(*grpcaCAServer).CreateSigningCertificate

<snipped rest of stack trace>
```

### DIFF
--- a/pkg/server/error.go
+++ b/pkg/server/error.go
@@ -42,6 +42,21 @@ const (
 )
 
 func handleFulcioGRPCError(ctx context.Context, code codes.Code, err error, message string, fields ...interface{}) error {
-	log.ContextLogger(ctx).Errorw(err.Error(), append([]interface{}{"code", code, "clientMessage", message, "error", err}, fields...)...)
+	// Use log level "warning" for codes that are likely client errors, see https://grpc.github.io/grpc/core/md_doc_statuscodes.html
+	switch code {
+	case codes.InvalidArgument,
+		codes.NotFound,
+		codes.AlreadyExists,
+		codes.PermissionDenied,
+		codes.Unauthenticated,
+		codes.FailedPrecondition,
+		codes.OutOfRange,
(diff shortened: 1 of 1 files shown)


jku:
The issue has this suggestion: 
> for JWT errors, I would remove logging entirely since there's nothing actionable to do.

We could still do that... It requires adding some custom errors (since they are not used currently) and then handling them specifically in handleFulcioGRPCError.

PR #2190, opened 2025-10-29
Add extension for deployment environment

### LINKED ISSUE #1845: Fulcio certs: Encode the CI/CD "environment" in an extension?
## Problem / value statement

The current Fulcio X.509 extensions include [`Runner Environment`](https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md#13614157264111--runner-environment), which corresponds to a GitHub Actions *infrastructure* scope, e.g. `github-hosted` or `self-hosted` or similar.

However, there's another kind of "environment" in GitHub Actions (and possibly other CI/CD providers) as well: job within workflows can configure [deployment environments](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment), which have user-controlled names and can be mapped to deployment protection rules.

For example, here's an example of an environment (named `pypi`) used in an individual workflow job:

https://github.com/trailofbits/pypi-attestations/blob/464bd18b42b7ba207d6a77e9cca468116aa0406b/.github/workflows/release.yml#L39-L41

When a job references an environment, its OIDC claims include references to that environment. In particular, the `sub` claim contains the `environment:$NAME` clause, and there's also a separate `environment` claim that contains the environment name directly. Examples of this can be seen in GitHub's [OIDC docs](https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect#understanding-the-oidc-token).

At the moment, it looks like Fulcio issued certificates don't reflect these user-controlled deployment environments, probably because there's no dedicated OID/extension definition for them. For example, [this recent log entry](https://search.sigstore.dev/?logIndex=142610048) has no reference to the `pypi` environment it came from. This is true even for the subject, since Fulcio rewrites the SAN away from the `sub` claim provided by the GitHub Actions OIDC credential.

It would be useful to include the deployment environment's name in the Fulcio extensions for a few reasons:

1. A deployment environment can correspond to deployment protection rules, making it possible to confirm (by the repo owner at least) that a workflow ran with explicit user approval or other checks.
2. Other OIDC applications, like Trusted Publishing, use the deployment environment as a binding piece of metadata for the publisher's "identity". It'd be nice if these identities could fully align with what Fulcio can provide.

## Proposal

1. Define a new OID + extension for `Deployment Environment` or similar (maybe there's a better name)
2. Define `Deployment Environment` as corresponding to the `environment` claim on GitHub Actions and GitLab CI/CD (where it appears to have similar semantics)
3. Begin emitting Fulcio certs with the new extension where appropriate

CC @haydentherapper @segiddins as interested parties 🙂

Hayden-IO:
This defines a new OID for a "deployment environment". By itself, it's not very meaningful as it's a user-defined string, but this can be used by a monitor to validate artifact provenance, e.g. verifying that a deployment environment had certain protections enabled.

Fixes #1845


haydentherapper:
also cc @kommendorkapten

bobcallaway (approved):

PR #2210, opened 2025-11-15
Switch docker env from ct_server to TesseraCT

cmurphy:
Relates to https://github.com/sigstore/rekor-tiles/issues/73


haydentherapper (commented):
This looks great!

I realized, quite surprisingly, we don't have an e2e test in this repo that uses a non-mocked CT log. Could we throw together a minimal one, something like spinning up the containers, making a request using either https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon/ or a fake OIDC provider, and verifying the certificate and its SCT?

cmurphy:
I added https://github.com/sigstore/fulcio/pull/2230 to demonstrate working E2E tests with no changes , then based this PR on that commit.

haydentherapper (approved):

Issue #2219, opened 2025-11-21
Add support for disabling the email_verified claim requirement in fulcio


mogthesprog:
**Description**

Hello!

I've recently spent some time working with Fulcio, deploying it in an enterprise context, but have run into some issues integrating it with Microsoft Entra (Azure AD). Entra (and ADFS) does not support passing the email_verified claim. After some digging around, I've found these two other issues discussing the same issue in both [Dex](https://github.com/dexidp/dex/issues/1405) and [Kubernetes](https://github.com/kubernetes/kubernetes/issues/59496). A few of the enterprise IdPs don't support the email_verified claim, and its initial intent may have been to demonstrate that ownership of an email address was verified during sign-up, and that doesn't make much sense in an enterprise environment. 

I was hoping you may be interested in accepting a PR to add a field to the `github.com/sigstore/fulcio/pkg/config#OIDCIssuer` type,  `InsecureSkipEmailVerified bool`, with a comment explaining its relevance in some IdP scenarios? (happy for someone to suggest a better name, this is just how Dex exposes the configuration, and it seemed like a sensible starting point. It looks simple enough to implement, so I'll have a stab at implementing, and then we can decide what's best to do.

Cheers

PR #2220, opened 2025-11-21
feat: Add support for skipping the email_verified claim on email issuer types

mogthesprog:
#### Summary

This PR adds support for enterprise identity providers (Microsoft Entra/Azure AD, ADFS) that don't include the `email_verified` claim in their OIDC tokens.

**Problem:** Enterprise identity providers often handle email verification through centralized identity management systems rather than including an `email_verified` claim in OIDC tokens. Fulcio currently requires this claim to be present and set to `true` for all email-type issuers, which blocks valid tokens from these enterprise providers and prevents internal Fulcio deployments in organizations using Microsoft Entra, ADFS, or similar systems.

**Solution:** Added a new optional `skip-email-verification` configuration field per OIDC issuer. When set to `true`, Fulcio will skip the `email_verified` claim check while still validating email format. This maintains security by default (requires verification) while allowing explicit opt-in for trusted internal providers.

**Testing:**
  - All existing tests pass, confirming backward compatibility
  - New tests cover scenarios with missing `email_verified` claims, explicit `false` values, and `true` values when skip is enabled
  - Config parsing tests verify YAML and JSON deserialization of the new field

See issue #2219 for additional context.

#### Release Note

Added optional `skip-email-verification` configuration field for OIDC issuers to support enterprise identity providers (Microsoft Entra/Azure AD, ADFS) that don't include the `email_verified` claim in tokens. This field defaults to `false` to maintain existing security behavior. Operators can set it to `true` for trusted internal identity providers where email verification is handled through organizational identity management.

 Configuration example:
```yaml
  oidc-issuers:
    https://login.microsoftonline.com/TENANT_ID/v2.0:
      issuer-url: https://login.microsoftonline.com/TENANT_ID/v2.0
      client-id: your-client-id
      type: email
      skip-email-verification: true
```

#### Documentation

If we decide to go ahead with this, then i'll update the docs repo too.


pkg/config/config.go
haydentherapper: Can you also update this function? This is used to output the server's configuration, e.g. http://fulcio.sigstore.dev/api/v2/configuration

PR #2233, opened 2025-12-04
add protection to count token segments ahead of parsing

bobcallaway:
Count number of separators before calling Split to minimize memory allocation risk

Fixes: GHSA-f83f-xpx7-ffpw

### DIFF
--- a/pkg/identity/issuerpool.go
+++ b/pkg/identity/issuerpool.go
@@ -41,11 +41,11 @@ func (p IssuerPool) Authenticate(ctx context.Context, token string, opts ...conf
 }
 
 func extractIssuerURL(token string) (string, error) {
-	parts := strings.Split(token, ".")
-	if len(parts) != 3 {
-		return "", fmt.Errorf("oidc: malformed jwt, expected 3 parts got %d", len(parts))
+	if strings.Count(token, ".") != 2 {
+		return "", fmt.Errorf("oidc: malformed jwt, token must have 3 parts")
 	}
 
+	parts := strings.SplitN(token, ".", 3)
 	raw, err := base64.RawURLEncoding.DecodeString(parts[1])
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/identity/issuerpool.go
```diff
@@ -41,11 +41,11 @@ func (p IssuerPool) Authenticate(ctx context.Context, token string, opts ...conf
...
 
 func extractIssuerURL(token string) (string, error) {
-	parts := strings.Split(token, ".")
-	if len(parts) != 3 {
-		return "", fmt.Errorf("oidc: malformed jwt, expected 3 parts got %d", len(parts))
+	if strings.Count(token, ".") != 2 {
+		return "", fmt.Errorf("oidc: malformed jwt, token must have 3 parts")
 	}
 
+	parts := strings.SplitN(token, ".", 3)
```
adeinega: No need to return a slice of strings when you need to work with the header of a JWT in this function.

`header := token[:strings.Index(token, ".")]`, or say `strings.Cut(token, ".")` will do their job.

pkg/identity/issuerpool.go
```diff
@@ -41,11 +41,11 @@ func (p IssuerPool) Authenticate(ctx context.Context, token string, opts ...conf
 }
 
 func extractIssuerURL(token string) (string, error) {
-	parts := strings.Split(token, ".")
-	if len(parts) != 3 {
-		return "", fmt.Errorf("oidc: malformed jwt, expected 3 parts got %d", len(parts))
+	if strings.Count(token, ".") != 2 {
```
adeinega: You might want to be a bit more robust and handle not only regular JWTs but also SD JWTs, which have recently been graduated to a proposed standard ([RFC 9901](https://datatracker.ietf.org/doc/rfc9901/)).

An SD JWT has two "." or four "." in those cases when it includes a SD-JWT+KB proof (a cryptographic key binding proof).

PR #2343, opened 2026-05-19
feat: add support for unencrypted keys

JasonPowr:
#### Summary
Make --fileca-key-passwd optional for fileca. When omitted, keys are parsed directly via Go stdlib (x509.ParsePKCS8PrivateKey, etc.) instead of through pemutil.Read, which relies on the deprecated x509.DecryptPEMBlock. Encrypted keys continue to work when a password is provided.

#### Release Note

The --fileca-key-passwd flag is now optional. When omitted, the private key is parsed as an unencrypted PEM (PKCS#8, EC, or PKCS#1). Encrypted keys continue to work when a password is provided.

#### Documentation

The fileca backend now supports unencrypted private keys. To use one, omit --fileca-key-passwd


Hayden-IO:
Can you say more about why you would need this feature? Unencrypted keys on disk is a security risk and I'd prefer to not allow that.

JasonPowr:
> Can you say more about why you would need this feature? Unencrypted keys on disk is a security risk and I'd prefer to not allow that.

@Hayden-IO The main reason for this feature is to try to avoid the usage of the deprecated x509.DecryptPEMBlock used in pemutil.Read() which can panic in some restricted environments. Since the password flag was mandatory all deployments will hit this path. Users can still use encrypted keys if they wish its just optional now.

Hayden-IO:
Panics from usage of a deprecated function is surprising to me. Did you have a pointer to how this is happening?

Having an unencrypted key on disk is still a risk and not something I'd want to encourage for any real-world deployment. If it using a deprecated function is the problem, I'd rather look for a different solution, though using a third-party library wouldn't be preferred either given the STD library already implements this.

JasonPowr:
Apologies for the late reply, I was OOO for a bit :) 

> Panics from usage of a deprecated function is surprising to me. Did you have a pointer to how this is happening?

The panic isn't from the deprecation directly, DecryptPEMBlock uses MD5 for key derivation, which is not a FIPS-approved algorithm. In Go builds with fips140=only, calling non-approved crypto primitives panics at runtime.

Unfortunately there's no stdlib replacement for decrypting encrypted private keys (that I know of ) x509.DecryptPEMBlock is deprecated and x509.ParsePKCS8PrivateKey only parses unencrypted keys. Go issue [#8860](https://github.com/golang/go/issues/8860) has been open since 2014 with no resolution.

Making the password optional lets deployments that protect keys at the infrastructure level avoid hitting the deprecated path entirely, while encrypted keys still work for those who want them.

Hayden-IO:
Does https://pkg.go.dev/github.com/youmark/pkcs8 work for FIPS instead? From https://github.com/sigstore/sigstore/pull/2333 and https://github.com/sigstore/fulcio/pull/2357, I just added support for parsing private keys encrypted per RFC 1423 (the problematic one) and RFC 7468.

JasonPowr:
> Does https://pkg.go.dev/github.com/youmark/pkcs8 work for FIPS instead? From [sigstore/sigstore#2333](https://github.com/sigstore/sigstore/pull/2333) and #2357, I just added support for parsing private keys encrypted per RFC 1423 (the problematic one) and RFC 7468.

While youmark/pkcs8 does help avoid the DecryptPEMBlock issue, it uses golang.org/x/crypto/pbkdf2 and golang.org/x/crypto/scrypt for key derivation, which fall outside the Go standard library FIPS compliance boundary.

Hayden-IO:
I'm open to this feature for FIPS compliance, but I want to avoid just Fulcio supporting this. There's code duplication between the 4 Go Sigstore services currently for loading signing keys. I can work on moving it under the sigstore/sigstore library and adding support for an unencrypted key concurrently.

JasonPowr:
> I'm open to this feature for FIPS compliance, but I want to avoid just Fulcio supporting this. There's code duplication between the 4 Go Sigstore services currently for loading signing keys. I can work on moving it under the sigstore/sigstore library and adding support for an unencrypted key concurrently.

Sure that works :) thanks @Hayden-IO If I can help at all, just shout :)

PR #2356, opened 2026-05-29
Allow directly-configured Kubernetes issuers to use in-cluster auth path

### LINKED ISSUE #2355: K3s / non-default Kubernetes issuer no longer gets in-cluster CA+token handling
## Description

Commit [378c654](https://github.com/sigstore/fulcio/commit/378c654f48c3bafdced04ead7010aab2cb4c6ca1) (`Block cross-host redirects and restrict bearer token to expected host`, #2354), shipped in v1.8.6, changed the Kubernetes special-case in `httpClientForIssuer` from:

```go
if iss.Type == IssuerTypeKubernetes && hasK8SIssuer {
```

to:

```go
if iss.Type == IssuerTypeKubernetes && hasK8SIssuer && iss.IssuerURL == k8sIssuerURL {
```

In v1.8.6, the same change also added same-host redirect enforcement and restricted bearer-token attachment to requests whose host matches the configured issuer host, which makes sense as SSRF / token-leak mitigation. But the additional `iss.IssuerURL == k8sIssuerURL` clause appears to regress Kubernetes issuers that do not use the literal default issuer URL `https://kubernetes.default.svc`.

## Reproducer

K3s uses `--service-account-issuer=https://kubernetes.default.svc.cluster.local`, so projected service-account tokens carry:

```text
iss=https://kubernetes.default.svc.cluster.local
```

Fulcio config:

```yaml
OIDCIssuers:
  https://kubernetes.default.svc:
    IssuerURL: https://kubernetes.default.svc
    ClientID: sigstore
    Type: kubernetes
  https://kubernetes.default.svc.cluster.local:
    IssuerURL: https://kubernetes.default.svc.cluster.local
    ClientID: sigstore
    Type: kubernetes
```

This worked on v1.8.5, where any issuer of `Type: kubernetes` received the in-cluster CA + bearer-token client setup as long as the default Kubernetes issuer was present in config.

On v1.8.6, Fulcio resolves the matching issuer entry and then calls `httpClientForIssuer(fc, iss)` to build the OIDC discovery client. If the matched issuer is `https://kubernetes.default.svc.cluster.local`, the new equality check fails, so the HTTP client is built without the Kubernetes CA and without the projected service-account bearer token.

That means discovery against:

```text
https://kubernetes.default.svc.cluster.local/.well-known/openid-configuration
```

no longer uses the in-cluster auth path, so provider discovery fails (for example with TLS verification failure or 401), the verifier is never constructed, and signing fails.

## MetaIssuer case

This looks like it affects `MetaIssuers` too. `GetIssuer` materializes a concrete `OIDCIssuer` for a meta-issuer match and sets `IssuerURL` to the incoming token's `iss` claim. For a Kubernetes meta issuer that resolves to something like `https://kubernetes.*.svc.cluster.local`, that concrete issuer URL will also fail the new `iss.IssuerURL == k8sIssuerURL` check. In v1.8.6, meta issuers do preserve `CACert`, but they still miss the Kubernetes bearer-token attachment when this equality check fails.

## Impact

This appears to break deployments that use `Type: kubernetes` with a non-default issuer URL and rely on Fulcio's in-cluster CA / service-account-token discovery path, including K3s and similar configurations.

## Suggested fix

The redirect blocking and per-request host check in `bearerTokenTransport` already look like the important security control here: v1.8.6 only attaches the bearer token when the outbound request host matches the configured issuer host, and rejects cross-host use otherwise.

Because of that, it seems like the extra `iss.IssuerURL == k8sIssuerURL` gate in `httpClientForIssuer` is unnecessarily narrow and is the source of the regression. A minimal fix would be to revert that clause and keep:

```go
if iss.Type == IssuerTypeKubernetes && hasK8SIssuer {
```

That would restore the previous behavior while still keeping the new same-host redirect and bearer-token protections introduced by `378c654`.

If that is too broad, another option would be to make the special-case issuer configurable instead of hardcoding `https://kubernetes.default.svc`.

## References

- Regression commit: [378c654](https://github.com/sigstore/fulcio/commit/378c654f48c3bafdced04ead7010aab2cb4c6ca1)
- v1.8.5 code: [pkg/config/config.go#L298-L333](https://github.com/sigstore/fulcio/blob/v1.8.5/pkg/config/config.go#L298-L333)
- v1.8.6 code: [pkg/config/config.go#L318-L355](https://github.com/sigstore/fulcio/blob/v1.8.6/pkg/config/config.go#L318-L355)

kwmonroe:
#### Summary

Restore the in-cluster CA + projected service-account bearer-token auth path for directly configured Kubernetes issuers whose issuer URL is not literally `https://kubernetes.default.svc`, while preserving the v1.8.6 SSRF / token-leak protections added in 378c654.

This fixes the regression reported in #2355 for deployments such as K3s, where the service-account issuer may be `https://kubernetes.default.svc.cluster.local` and is explicitly listed in OIDCIssuers.

The existing `TestGetVerifier_MetaIssuerK8sTokenNotLeaked` continues to pass, and a new `TestGetVerifier_DirectConfiguredK8sIssuerGetsToken` covers the K3s case.

Closes #2355 

#### Release Note

* Fixed a regression introduced in v1.8.6 where Kubernetes-type issuers configured with a non-default URL (e.g. K3s's `https://kubernetes.default.svc.cluster.local`) no longer received the in-cluster CA and service-account bearer token during OIDC discovery.

#### Documentation

NONE

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -321,7 +321,9 @@ func httpClientForIssuer(fc *FulcioConfig, iss OIDCIssuer) (*http.Client, error)
 	}
 
 	_, hasK8SIssuer := fc.GetIssuer(k8sIssuerURL)
-	if iss.Type == IssuerTypeKubernetes && hasK8SIssuer && iss.IssuerURL == k8sIssuerURL {
+	_, isConfiguredOIDCIssuer := fc.OIDCIssuers[iss.IssuerURL]
+	isTrustedK8sIssuer := isConfiguredOIDCIssuer || iss.IssuerURL == k8sIssuerURL
+	if iss.Type == IssuerTypeKubernetes && hasK8SIssuer && isTrustedK8sIssuer {
 		// Add the Kubernetes cluster's CA to the client's CA pool
 		certs, err := os.ReadFile(k8sCA)
 		if err != nil {
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -1407,3 +1407,96 @@ func TestGetVerifier_MetaIssuerK8sTokenNotLeaked(t *testing.T) {
 		t.Fatal("vulnerability: sensitive local Kubernetes token was leaked to external meta-issuer URL")
 	}
 }
+
+// TestGetVerifier_DirectConfiguredK8sIssuerGetsToken verifies that a
+// directly-configured Kubernetes-type issuer whose URL is not k8sIssuerURL
+// (e.g. "https://kubernetes.default.svc.cluster.local" used by K3s) still
+// receives the in-cluster CA and projected service-account bearer token.
+func TestGetVerifier_DirectConfiguredK8sIssuerGetsToken(t *testing.T) {
+	origK8sCA := k8sCA
+	origK8sTokenFile := k8sTokenFile
+	origK8sIssuerURL := k8sIssuerURL
+	defer func() {
+		k8sCA = origK8sCA
+		k8sTokenFile = origK8sTokenFile
+		k8sIssuerURL = origK8sIssuerURL
+	}()
+
+	caPEM, serverTLSCert, err := createServerTLS()
+	if err != nil {
+		t.Fatal(err)
+	}
+	crtFile, err := mockFile("ca-*.crt", caPEM.String())
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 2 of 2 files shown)


Hayden-IO (approved):

Hayden-IO:
Thanks! I'll cut a new release shortly.

PR #2365, opened 2026-06-07
fix: add RSA-PSS signature algorithm support

### LINKED ISSUE #1858: Fulcio API incorrectly specifies support for RSA-PSS

AruneshDwivedi:
## Description

Fixes #1858. Fulcio's API documentation specifies RSA-PSS as a supported algorithm, but the `getHashFuncForSignatureAlgorithm` function was missing cases for RSA-PSS signature algorithms.

## Root Cause

The `getHashFuncForSignatureAlgorithm` function in `pkg/server/grpc_server.go` handled ECDSA, RSA (PKCS#1 v1.5), and Ed25519 signature algorithms, but did not handle:
- `x509.SHA256WithRSAPSS`
- `x509.SHA384WithRSAPSS`  
- `x509.SHA512WithRSAPSS`

This caused CSRs using RSA-PSS signature algorithms to be rejected with an "unrecognized signature algorithm" error.

## Changes

Added the three RSA-PSS signature algorithm cases to `getHashFuncForSignatureAlgorithm`, mapping each to its corresponding hash function (SHA-256, SHA-384, SHA-512).

## Testing

Existing tests should continue to pass. A new test case for RSA-PSS CSRs would be a good addition.

### DIFF
--- a/pkg/server/grpc_server.go
+++ b/pkg/server/grpc_server.go
@@ -331,6 +331,12 @@ func getHashFuncForSignatureAlgorithm(signatureAlgorithm x509.SignatureAlgorithm
 		return crypto.SHA384, nil
 	case x509.SHA512WithRSA:
 		return crypto.SHA512, nil
+	case x509.SHA256WithRSAPSS:
+		return crypto.SHA256, nil
+	case x509.SHA384WithRSAPSS:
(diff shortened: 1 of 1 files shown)


Hayden-IO:
There is another open PR to resolve this by documentation.

PR #2378, opened 2026-06-16
docs: clarify RSA_PSS algorithm and ignored field in proto

### LINKED ISSUE #1858: Fulcio API incorrectly specifies support for RSA-PSS

ogulcanaydogan:
## Summary

Fulcio's API advertises `RSA_PSS` as a supported algorithm, but as noted in #1858:

- the `algorithm` field is not used when parsing the public key (the algorithm is inferred from the submitted key), and
- RSA keys are treated as RSA PKCS#1 v1.5 rather than RSA-PSS.

This adds clarifying comments to `fulcio.proto` (and the regenerated output) so the
API contract is clear to clients, following @haydentherapper's suggestion in the issue
to document the current behaviour now and correct it properly in a future major version.

No behavioural change, comments only.

## Release Note

```release-note
NONE
```

Fixes #1858

### DIFF
--- a/fulcio.proto
+++ b/fulcio.proto
@@ -134,7 +134,9 @@ message PublicKeyRequest {
 
 message PublicKey {
     /*
-     * The cryptographic algorithm to use with the key material
+     * The cryptographic algorithm to use with the key material.
+     * This field is currently ignored; the algorithm is inferred from the
+     * public key when it is parsed.
      */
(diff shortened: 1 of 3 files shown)


Hayden-IO (approved):

Hayden-IO:
Thanks!

PR #2392, opened 2026-07-06
Add --ct-log-origin flag

cmurphy:
Since the move to TesseraCT, the logs are clogged with the warning "AddPreChain the request was received on a URL which is not prefixed with the configured origin". This is because Fulcio is accessing the CT log in-cluster from its Kubernetes service name instead of its public name. This is intentional to avoid an external network round trip, but TesseraCT is semi-strict about the host name it was reached on. This change adds an override to declare the CT log's origin name while still keeping the URL as the actual address to contact.

Generated by Gemini.

### DIFF
--- a/.github/workflows/verify-k8s.yml
+++ b/.github/workflows/verify-k8s.yml
@@ -67,6 +67,7 @@ jobs:
       KO_PREFIX: registry.local:5000/fulcio
       GIT_HASH: ${{ github.sha }}
       GIT_VERSION: test
+      INSECURE_REGISTRY: true
 
     steps:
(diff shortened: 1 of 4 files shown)


Hayden-IO:
Could you add the insecure flags to https://github.com/sigstore/fulcio/blob/7074b709ec6309d863f8789bdd32a75b49c5e127/Makefile#L117-L122 while we're here?

Hayden-IO (dismissed):

cmurphy:
@Hayden-IO done, this is now green

Hayden-IO (approved):
Thank you for fixing the tests!

PR #2406, opened 2026-07-20
fix(ciprovider): format large whole-number claims without scientific notation

arpitjain099:
## Summary

`mapValuesToString` formats whole-number float claims with `fmt.Sprintf("%v", ...)`. For a float64 that switches to scientific notation once the value reaches 1e6, so any numeric OIDC claim at or above a million ends up in the certificate in exponent form.

The sibling branch for fractional floats already does the right thing with `strconv.FormatFloat(value, 'f', -1, 64)`, so this just makes the whole-number branch consistent with it.

What it looks like today:

- `1000000` becomes `1e+06`
- `1753000000`, an ordinary `exp` / `iat` / `nbf`, becomes `1.753e+09`
- GitLab's `project_id` and `user_id` are affected once they pass a million

These values are what `applyTemplateOrReplace` then interpolates into the SAN and the extension values, so the mangled form is what gets written to the issued certificate.

Values below 1e6 are unchanged, and fractional claims keep their current formatting.

## Testing

`TestGetTokenClaims` already distinguishes ints from floats but never exercises magnitude, so I added two cases there. Without the change they fail with:

```
Expected 1753000000, received 1.753e+09
Expected 1000000, received 1e+06
```

With it, `go test ./pkg/identity/...` is green.

### DIFF
--- a/pkg/identity/ciprovider/principal.go
+++ b/pkg/identity/ciprovider/principal.go
@@ -42,10 +42,7 @@ func mapValuesToString(claims map[string]any) map[string]string {
 		case reflect.Float32, reflect.Float64:
 			value := vType.Interface().(float64)
 			if value == math.Trunc(value) {
-				// A float, but with no fractional part. Treat as an int.
-				// Format it explicitly rather than with %v, which switches to
-				// scientific notation once the value reaches 1e6 and would put
-				// e.g. "1.753e+09" in the certificate for an exp/iat claim.
+				// A float, but with no fractional part. Treat as an int
 				newMap[k] = strconv.FormatFloat(math.Trunc(value), 'f', -1, 64)
 			} else {
 				newMap[k] = strconv.FormatFloat(value, 'f', -1, 64)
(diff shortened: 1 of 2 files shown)


pkg/identity/ciprovider/principal.go
```diff
@@ -42,8 +42,11 @@ func mapValuesToString(claims map[string]any) map[string]string {
 		case reflect.Float32, reflect.Float64:
 			value := vType.Interface().(float64)
 			if value == math.Trunc(value) {
-				// A float, but with no fractional part. Treat as an int
-				newMap[k] = fmt.Sprintf("%v", math.Trunc(value))
+				// A float, but with no fractional part. Treat as an int.
+				// Format it explicitly rather than with %v, which switches to
```
Hayden-IO: No need to add this comment, since this is referencing the previous state of the codebase.

Hayden-IO (commented):
Thanks, this is a good find! Just a small comment about a comment.

pkg/identity/ciprovider/principal.go
arpitjain099: Good call, dropped it in cb2a67e. You're right that it reads as a note about what the code used to do rather than what it does now, which won't age well once this is merged. Kept the original one-liner.

Hayden-IO (approved):
Thanks again!

PR #2407, opened 2026-07-20
Remove unused chainops.dev OIDC issuer

loosebazooka:
Remove chainops.dev as it unused

This should eventually sync to prod when we update the pgi config, @codysoyland 's intial research indicated that this removal might be okay, but we'll have to see.

### DIFF
--- a/config/identity/config.yaml
+++ b/config/identity/config.yaml
@@ -80,12 +80,6 @@ oidc-issuers:
     type: chainguard-identity
     contact: mattmoor@chainguard.dev
     description: "Chainguard identity tokens"
-  https://issuer.chainops.dev:
-    issuer-url: https://issuer.chainops.dev
-    client-id: sigstore
-    type: chainguard-identity
-    contact: priya@chainguard.dev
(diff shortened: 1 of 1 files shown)


config/identity/config.yaml
```diff
@@ -80,12 +80,6 @@ oidc-issuers:
     type: chainguard-identity
     contact: mattmoor@chainguard.dev
```
Hayden-IO, line 81: While we're here, is there a group we could use as the contact instead of an individual?

Hayden-IO (approved):

config/identity/config.yaml
loosebazooka: @mattmoor ? although we can change this in a followup

codysoyland (approved):

loosebazooka:
If anyone has merge rights on fulcio, I can't merge here.

codysoyland:
Nit: rename PR "Remove unused chainops.dev OIDC issuer"

bobcallaway (approved):

PR #2410, opened 2026-07-22
Specify the environment the identity provider is supported

Hayden-IO:
This value will be used by the syncing job to know if an issuer should only be in the staging or production environment. No set value means both. I've added the Dex staging instance, and specified the staging label for the recently added Buddy CI dev providers.

Also alphabetize the config for the syncing job

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -137,6 +137,9 @@ type OIDCIssuer struct {
 	// Optional, the contact for the issuer team
 	// Usually it is a email
 	Contact string `json:"Contact,omitempty" yaml:"contact,omitempty"`
+	// Environment specifies whether this issuer is restricted to a specific environment
+	// (e.g. "staging" or "production"). If omitted, it is valid in all environments.
+	Environment string `json:"Environment,omitempty" yaml:"environment,omitempty"`
 
 	// CACert is an optional parameter that holds the CA certificate in PEM format.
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -137,6 +137,9 @@ type OIDCIssuer struct {
 	// Optional, the contact for the issuer team
 	// Usually it is a email
 	Contact string `json:"Contact,omitempty" yaml:"contact,omitempty"`
+	// Environment specifies whether this issuer is restricted to a specific environment
+	// (e.g. "staging" or "production"). If omitted, it is valid in all environments.
+	Environment string `json:"Environment,omitempty" yaml:"environment,omitempty"`
```
loosebazooka, line 142: It seems a bit odd to me that this option is only consumed by some action to filter out parts. The fulcio parser itself doesn't care. Should fulcio itself need to have code to identify these environments. Should this instead go into PGI or somewhere?

Hayden-IO: That's already the case with 'description' and 'contact', additional metadata just for whoever is reading this file.

Hayden-IO:
bump for review

loosebazooka (approved):
the fulcio codebase still feels like the wrong place to keep this config. But we can deal with that later.

bobcallaway (approved):

PR #2418, opened 2026-08-05
feat: serve TLS on HTTP/duplex and make TLS policy configurable

### LINKED ISSUE #2415: Support TLS on the HTTP endpoint and make TLS policy configurable
## Summary

Fulcio can present a server certificate on its **gRPC** listener only. The **HTTP/REST** endpoint, what most clients, including cosign, actually use never terminates TLS, in either of Fulcio's two serving modes (see below). The one path that does serve TLS (gRPC) also has its policy (`MinVersion: tls.VersionTLS13`) hardcoded, with no flag to change it.

This issue asks for:

1. **Serving TLS on the HTTP path**, in both serving modes.
2. **A configurable minimum TLS version and cipher suite list**, applied consistently
   to every serving path.

## Background: Fulcio's two serving modes

Fulcio can run as:
- **Two listeners** (default): a gRPC listener (`--grpc-host`/`--grpc-port`) and an HTTP listener (`--host`/`--port`, aliased as `--http-host`/`--http-port`) on  separate ports, each with its own `net.Listener`.
- **Duplex mode** (auto-selected when the HTTP and gRPC host:port match): a single  shared listener where `chainguard.dev/go-grpc-kit`'s `Duplex` type serves both the gRPC handler (via `grpc.Server.ServeHTTP`, over cleartext HTTP/2) and the REST  gateway mux from one `http.Server`. It's HTTP-layer request multiplexing, not a TLS/ALPN sniffer, so adding TLS here means wrapping that one shared listener  (e.g. `tls.NewListener` / `ServeTLS`), the same way any Go HTTP/2 server adds TLS.

## Current behaviour

* **gRPC listener (two-listener mode):** TLS is supported via `--grpc-tls-certificate`  / `--grpc-tls-key`, with `MinVersion: tls.VersionTLS13` hardcoded in  `cachedTLSCert.GRPCCreds()` ([cmd/app/grpc.go](https://github.com/sigstore/fulcio/blob/main/cmd/app/grpc.go)). No flag can change the minimum version or set a TLS 1.2 cipher list.
* **HTTP listener (two-listener mode):** always plaintext. `httpServer.startListener` ([cmd/app/http.go](https://github.com/sigstore/fulcio/blob/main/cmd/app/http.go)) calls  `ListenAndServe`; no `TLSConfig` is ever set and `ListenAndServeTLS` is never called.
* **Duplex mode:** always plaintext. `StartDuplexServer` ([cmd/app/serve.go](https://github.com/sigstore/fulcio/blob/main/cmd/app/serve.go)) constructs the shared listener with `insecure.NewCredentials()` and plain `net.Listen`; there's no TLS-material flag wired to this path at all.
* **Related, not a hardening ask:** when gRPC TLS is enabled, the in-process REST -> gRPC  gateway dial in `createHTTPServer` uses `InsecureSkipVerify: true` (already annotated `#nosec G402` as an accepted loopback trust boundary). Noted for completeness since it's the same trust boundary this issue is about.

## Why gRPC has TLS and HTTP/duplex don't

There's no architectural or notable design decision captured that could be referred to. TLS was added to the gRPC listener in PR #1252 (2023), scoped only to the two-listener gRPC path. It was never extended to the HTTP listener or to duplex mode. Both are solvable with the same standard Go primitives already used for gRPC (`tls.Config` + `credentials.NewTLS`, plus `cachedTLSCert`'s existing `fsnotify`-based
reload).

## Motivation

Fulcio's REST endpoint carries a live bearer credential in both directions:
* The **request** carries an OIDC identity token in the `Authorization` header  in the clear, on a plaintext hop, that's a bearer credential anyone on-path can observe and use to obtain a signing certificate within the token's validity window.
* The **response** carries the issued certificate; the request carries the CSR/public key being certified both substitutable by an on-path attacker without TLS.
* Under reencrypt/passthrough ingress (no transparent mTLS mesh), the terminator re-encrypts to the pod, so the pod itself must be able to serve TLS. Fulcio's HTTP endpoint can't, so that hop stays plaintext regardless of the front door.
* Compliance baselines requiring encryption-in-transit with a demonstrable minimum version (FIPS 140-x, FedRAMP, PCI-DSS, etc.) can't be evidenced on an endpoint with no TLS, or no configurable minimum, even when its actual posture is fine.

This applies to any operator terminating TLS in front of Fulcio and re-encrypting to the pod, or with a policy requiring a specific TLS version/cipher set - not tied to any specific platform.

## Proposed change

Both changes default to today's behaviour when unset — no existing deployment is
affected.

1. **HTTP TLS material:** add `--http-tls-certificate` / `--http-tls-key` (optionally `--http-tls-ca` for client-cert verification). Wire into `createHTTPServer` /  `startListener` (`ListenAndServeTLS`) and into `StartDuplexServer`'s shared listener. Reuse the existing `cachedTLSCert` watcher rather than adding a second mechanism.
2. **Policy flags:** add `--tls-min-version` and `--tls-cipher-suites`, applied to every serving path's `tls.Config`. Keep gRPC's default at TLS 1.3 (no silent downgrade on upgrade); default the new HTTP/duplex paths to Go's TLS 1.2 floor unless configured higher.

Note: `tls.Config.CipherSuites` only affects TLS 1.2. Go fixes the TLS 1.3 suite set. A cipher flag has no effect on gRPC today unless the minimum version becomes configurable too, so the two flags need to land together.

## Scope

* **In scope:** TLS on the HTTP listener and duplex mode; configurable min version and TLS 1.2 ciphers on all serving paths.
* **Optional:** client-certificate verification on the HTTP listener.
* **Out of scope:** the metrics listener (separate concern); outbound TLS to the CTlog (already configurable via `--ct-log.tls-ca-cert`); the `InsecureSkipVerify`  gateway dial (noted above, not requested here).

## Acceptance criteria

* Unconfigured: identical behaviour to today on all paths, including gRPC's TLS 1.3 minimum.
* Configured: HTTP serves HTTPS in both two-listener and duplex modes; a configured minimum version is enforced (verified against a client offering only a lower version); a configured TLS 1.2 cipher list is enforced on TLS 1.2 handshakes.
* Certificate rotation on disk is picked up without restart on the HTTP path, matching existing gRPC behaviour.
* Misconfiguration (cert without key, unparseable cipher name) fails at startup with a clear error.

anithapriyanatarajan:
Adds a shared TLS policy layer so the HTTP/REST listener, the gRPC listener, and the duplex listener can all terminate TLS with a consistent, configurable policy.

New flags:

```
--http-tls-certificate / --http-tls-key  enable TLS on the HTTP endpoint and the shared listener
--tls-min-version                        minimum TLS version for all serving paths (1.2 or 1.3)
--tls-cipher-suites                      allowed TLS 1.2 cipher suite names applied to all serving paths
```

The min-version and cipher-suite parsers reject unsupported versions and insecure/unknown suites, and `validateTLSFlags` fails fast at startup on mismatched certificate/key pairs. The gRPC listener keeps its existing fsnotify cert hot-reload; the HTTP/REST and shared duplex listeners load the certificate once at startup, so rotating it needs a restart.

## Summary

Closes #2415.

Today Fulcio can only terminate TLS on the gRPC listener (`--grpc-tls-certificate` / `--grpc-tls-key`); the HTTP/REST endpoint is always served in plaintext, and the minimum TLS version and cipher suites are not configurable on any path. This makes it hard to run Fulcio behind operators/platforms that need TLS terminated at the component and want to enforce an organization-wide TLS policy.

This PR introduces a small shared TLS policy layer (`internal/tlspolicy`) that is applied uniformly to all three serving paths, the two-listener HTTP and gRPC servers and the duplex (single shared listener) server:

- TLS on the HTTP endpoint via `--http-tls-certificate` / `--http-tls-key`. In duplex mode the same certificate secures the shared listener; ALPN advertises `h2` then `http/1.1` so gRPC-over-HTTP/2 still negotiates correctly, and the loopback gateway dial is upgraded to TLS.
- Configurable minimum TLS version via `--tls-min-version` (`1.2` or `1.3`). When unset, every serving path defaults to a TLS 1.3 floor. gRPC already required TLS 1.3, so this only raises the new HTTP/duplex paths and introduces no regression; deployments fronted by a peer that cannot negotiate TLS 1.3 can opt down with `--tls-min-version 1.2`.
- Configurable TLS 1.2 cipher suites via `--tls-cipher-suites`, restricted to the secure suites reported by the Go standard library (`crypto/tls`); insecure/unknown names are rejected. This has no effect at the default TLS 1.3 floor, whose suite set is fixed by Go, and only takes effect when the minimum version is lowered to 1.2.
- Fail-fast validation (`validateTLSFlags`) at startup: a certificate without its matching key (or vice versa) on either path, an unsupported version, or an unknown cipher suite name is rejected before serving.
- Certificate rotation: the gRPC listener keeps its existing fsnotify-based hot-reload. The HTTP/REST listener and the shared duplex listener load their certificate once at startup, so rotating it on disk requires a restart (per review feedback — mid-process reloads are avoided in favor of a restart).

All flags are opt-in and default to the current behavior, so this is backward compatible: without the new flags, Fulcio serves exactly as it does today.

## How to test

- Run `go test ./internal/tlspolicy/... ./cmd/app/...` — `internal/tlspolicy` covers version/cipher parsing and policy resolution; `cmd/app` covers flag validation plus end-to-end TLS: `TestDuplexTLS` starts a TLS-enabled duplex server and confirms `GET /healthz` succeeds over HTTPS while a client offering below the minimum version is rejected, and `TestCreateHTTPServerTLS` / `TestStartListenerTLS` cover the HTTP TLS-termination path.
- Manually: start `fulcio serve` with `--http-tls-certificate`/`--http-tls-key` (optionally `--tls-min-version 1.2`) and confirm the REST endpoint is served over HTTPS and that a sub-minimum client handshake is refused. To verify rotation behavior, replace the cert/key on disk while running — the listener keeps serving the old cert until Fulcio is restarted.

#### Release Note

```release-note
Added TLS support for the HTTP endpoint and a configurable TLS policy across all serving paths. New `serve` flags: `--http-tls-certificate` and `--http-tls-key` enable TLS on the HTTP/REST endpoint (and the shared listener in duplex mode), `--tls-min-version` sets the minimum TLS version (1.2 or 1.3, defaulting to 1.3), and `--tls-cipher-suites` restricts the allowed TLS 1.2 cipher suites. All flags are opt-in. Existing deployments are unaffected. The gRPC listener continues to hot-reload its certificate; the HTTP/REST and duplex listeners load the certificate at startup and require a restart to pick up a rotated certificate.
```

#### Documentation

Updated `docs/setup.md` with a new "Serving over TLS" section covering the `--grpc-tls-*`, `--http-tls-*`, `--tls-min-version`, and `--tls-cipher-suites` flags, certificate-rotation behavior, and an example. No change to https://docs.sigstore.dev is required.

### DIFF
--- a/cmd/app/grpc.go
+++ b/cmd/app/grpc.go
@@ -35,6 +35,7 @@ import (
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus"
 	grpc_recovery "github.com/grpc-ecosystem/go-grpc-middleware/recovery"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/sigstore/fulcio/internal/tlspolicy"
 	"github.com/sigstore/fulcio/pkg/ca"
 	"github.com/sigstore/fulcio/pkg/config"
 	gw "github.com/sigstore/fulcio/pkg/generated/protobuf"
@@ -107,7 +108,7 @@ func newCachedTLSCert(certPath, keyPath string) (*cachedTLSCert, error) {
 					return
 				}
 				if event.Has(fsnotify.Write) {
-					log.Logger.Info("fsnotify grpc-tls-certificate write event detected")
+					log.Logger.Info("fsnotify tls-certificate write event detected")
 					if err := cachedTLSCert.UpdateCertificate(); err != nil {
 						log.Logger.Error(err)
 					}
@@ -116,7 +117,7 @@ func newCachedTLSCert(certPath, keyPath string) (*cachedTLSCert, error) {
 				if !ok {
 					return
 				}
-				log.Logger.Error("fsnotify grpc-tls-certificate error:", err)
+				log.Logger.Error("fsnotify tls-certificate error:", err)
 			}
 		}
 	}()
@@ -142,20 +143,30 @@ func (c *cachedTLSCert) UpdateCertificate() error {
 
 	cert, err := tls.LoadX509KeyPair(c.certPath, c.keyPath)
 	if err != nil {
-		return fmt.Errorf("loading GRPC tls certificate and key file: %w", err)
+		return fmt.Errorf("loading tls certificate and key file: %w", err)
 	}
 
 	c.cert = &cert
 	return nil
 }
 
-func (c *cachedTLSCert) GRPCCreds() grpc.ServerOption {
-	return grpc.Creds(credentials.NewTLS(&tls.Config{
+// tlsConfig builds a *tls.Config that serves the watched certificate under the
+// supplied policy. nextProtos advertises the ALPN protocol list (nil for gRPC,
+// which negotiates HTTP/2 via its own credentials).
+func (c *cachedTLSCert) tlsConfig(p tlspolicy.Policy, nextProtos []string) *tls.Config {
+	/* #nosec G402 */ // MinVersion is a variable, but tlspolicy.ParseVersion constrains it to TLS 1.2 or 1.3 and callers default it to a literal >= TLS 1.2, so it is never below the G402 floor.
+	return &tls.Config{
 		GetCertificate: func(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
 			return c.GetCertificate(), nil
 		},
-		MinVersion: tls.VersionTLS13,
-	}))
+		MinVersion:   p.MinVersion,
+		CipherSuites: p.CipherSuites,
+		NextProtos:   nextProtos,
+	}
+}
+
+func (c *cachedTLSCert) GRPCCreds(p tlspolicy.Policy) grpc.ServerOption {
+	return grpc.Creds(credentials.NewTLS(c.tlsConfig(p, nil)))
 }
 
 func createGRPCServer(cfg *config.FulcioConfig, ctClient *ctclient.LogClient, baseca ca.CertificateAuthority, algorithmRegistry *signature.AlgorithmRegistryConfig, ip identity.IssuerPool) (*grpcServer, error) {
@@ -187,8 +198,13 @@ func createGRPCServer(cfg *config.FulcioConfig, ctClient *ctclient.LogClient, ba
 			return nil, err
 		}
 
+		policy, err := resolveTLSPolicy()
+		if err != nil {
+			return nil, err
+		}
+
 		tlsCertWatcher = cachedTLSCert.Watcher
-		serverOpts = append(serverOpts, cachedTLSCert.GRPCCreds())
+		serverOpts = append(serverOpts, cachedTLSCert.GRPCCreds(policy))
 	}
 
 	myServer := grpc.NewServer(serverOpts...)
--- a/cmd/app/http.go
+++ b/cmd/app/http.go
@@ -30,9 +30,11 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/fsnotify/fsnotify"
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/rs/cors"
+	"github.com/sigstore/fulcio/internal/tlspolicy"
 	gw "github.com/sigstore/fulcio/pkg/generated/protobuf"
 	legacy_gw "github.com/sigstore/fulcio/pkg/generated/protobuf/legacy"
 	"github.com/sigstore/fulcio/pkg/log"
@@ -49,6 +51,7 @@ import (
 type httpServer struct {
 	*http.Server
 	httpServerEndpoint string
+	tlsCertWatcher     *fsnotify.Watcher
 }
 
 func extractOIDCTokenFromAuthHeader(_ context.Context, req *http.Request) metadata.MD {
@@ -112,11 +115,31 @@ func createHTTPServer(ctx context.Context, serverEndpoint string, grpcServer, le
 		WriteTimeout:      60 * time.Second,
 		IdleTimeout:       viper.GetDuration("idle-connection-timeout"),
 	}
-	return httpServer{&api, serverEndpoint}
+
+	// Optionally terminate TLS on the HTTP listener. cachedTLSCert serves the
+	// certificate so on-disk rotation is picked up without a restart.
+	var tlsCertWatcher *fsnotify.Watcher
+	if viper.IsSet("http-tls-certificate") && viper.IsSet("http-tls-key") {
+		cachedTLSCert, err := newCachedTLSCert(viper.GetString("http-tls-certificate"), viper.GetString("http-tls-key"))
+		if err != nil {
+			log.Logger.Fatal(err)
+		}
+		policy, err := resolveTLSPolicy()
+		if err != nil {
+			log.Logger.Fatal(err)
+		}
+		api.TLSConfig = cachedTLSCert.tlsConfig(policy, tlspolicy.ALPNProtocols)
+		tlsCertWatcher = cachedTLSCert.Watcher
+	}
+	return httpServer{&api, serverEndpoint, tlsCertWatcher}
 }
 
 func (h httpServer) startListener(wg *sync.WaitGroup) {
-	log.Logger.Infof("listening on http at %s", h.httpServerEndpoint)
+	scheme := "http"
+	if h.TLSConfig != nil {
+		scheme = "https"
+	}
+	log.Logger.Infof("listening on %s at %s", scheme, h.httpServerEndpoint)
 
 	idleConnsClosed := make(chan struct{})
 	go func() {
@@ -135,7 +158,18 @@ func (h httpServer) startListener(wg *sync.WaitGroup) {
 
 	wg.Add(1)
 	go func() {
-		if err := h.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+		if h.tlsCertWatcher != nil {
+			defer h.tlsCertWatcher.Close()
+		}
+		// The certificate and key are supplied via TLSConfig.GetCertificate, so
+		// the file arguments are intentionally empty.
+		var err error
+		if h.TLSConfig != nil {
+			err = h.ListenAndServeTLS("", "")
+		} else {
+			err = h.ListenAndServe()
+		}
+		if err != nil && !errors.Is(err, http.ErrServerClosed) {
 			log.Logger.Fatal(err)
 		}
 		<-idleConnsClosed
--- a/cmd/app/serve.go
+++ b/cmd/app/serve.go
@@ -48,6 +48,7 @@ import (
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/sigstore/fulcio/internal/tlspolicy"
 	certauth "github.com/sigstore/fulcio/pkg/ca"
 	"github.com/sigstore/fulcio/pkg/ca/ephemeralca"
 	"github.com/sigstore/fulcio/pkg/ca/fileca"
@@ -70,6 +71,7 @@ import (
 	"goa.design/goa/v3/grpc/middleware"
 	"google.golang.org/api/option"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 	health "google.golang.org/grpc/health/grpc_health_v1"
 	"google.golang.org/grpc/keepalive"
@@ -122,6 +124,10 @@ func newServeCmd() *cobra.Command {
 	cmd.Flags().Duration("read-header-timeout", 10*time.Second, "The time allowed to read the headers of the requests in seconds")
 	cmd.Flags().String("grpc-tls-certificate", "", "the certificate file to use for secure connections - only applies to grpc-port")
 	cmd.Flags().String("grpc-tls-key", "", "the private key file to use for secure connections (without passphrase) - only applies to grpc-port")
+	cmd.Flags().String("http-tls-certificate", "", "the certificate file to use for secure connections on the HTTP endpoint (two-listener mode) and the shared listener (duplex mode)")
+	cmd.Flags().String("http-tls-key", "", "the private key file to use for secure connections (without passphrase) on the HTTP endpoint (two-listener mode) and the shared listener (duplex mode)")
+	cmd.Flags().String("tls-min-version", "", "minimum TLS version for all serving paths (1.2 or 1.3); when unset defaults to 1.3. Lower to 1.2 for peers that cannot negotiate TLS 1.3.")
+	cmd.Flags().StringSlice("tls-cipher-suites", nil, "allowed TLS 1.2 cipher suite names (crypto/tls spelling) applied to all serving paths; empty keeps the crypto/tls default. Has no effect on TLS 1.3.")
 	cmd.Flags().Duration("idle-connection-timeout", 30*time.Second, "The time allowed for connections (HTTP or gRPC) to go idle before being closed by the server")
 	cmd.Flags().String("ct-log.tls-ca-cert", "", "Path to TLS CA certificate used to connect to ct-log")
 	cmd.Flags().String("hsm-key-label", "PKCS11CA", "HSM key label for PKCS11 CA")
@@ -186,6 +192,24 @@ func (rt *hostRoundTripper) RoundTrip(req *http.Request) (*http.Response, error)
 	return inner.RoundTrip(req)
 }
 
+func resolveTLSPolicy() (tlspolicy.Policy, error) {
+	return tlspolicy.Resolve(viper.GetString("tls-min-version"), viper.GetStringSlice("tls-cipher-suites"))
+}
+
+// validateTLSFlags rejects a certificate supplied without its matching key on
+// either serving path before any listener starts. The minimum version and
+// cipher suite flags are validated by resolveTLSPolicy where the policy is
+// applied.
+func validateTLSFlags() error {
+	if viper.IsSet("grpc-tls-certificate") != viper.IsSet("grpc-tls-key") {
+		return fmt.Errorf("--grpc-tls-certificate and --grpc-tls-key must be set together")
+	}
+	if viper.IsSet("http-tls-certificate") != viper.IsSet("http-tls-key") {
+		return fmt.Errorf("--http-tls-certificate and --http-tls-key must be set together")
+	}
+	return nil
+}
+
 func runServeCmd(cmd *cobra.Command, args []string) { //nolint: revive
 	ctx := cmd.Context()
 	// If a config file is provided, modify the viper config to locate and read it
@@ -200,6 +224,10 @@ func runServeCmd(cmd *cobra.Command, args []string) { //nolint: revive
 	// Allow recognition of environment variables such as FULCIO_SERVE_CA etc.
 	viper.SetEnvPrefix(serveCmdEnvPrefix)
 	viper.AutomaticEnv()
+
+	if err := validateTLSFlags(); err != nil {
+		log.Logger.Fatal(err)
+	}
 
 	switch viper.GetString("ca") {
 	case "":
@@ -498,9 +526,38 @@ func StartDuplexServer(ctx context.Context, cfg *config.FulcioConfig, ctClient *
 		grpc_prometheus.WithServerHandlingTimeHistogram(),
 	)
 
+	// Resolve optional TLS on the shared duplex listener up front. The duplex
+	// gateway dials itself over the loopback to translate REST to gRPC, so when
+	// the listener terminates TLS that in-process dial must use TLS credentials
+	// too, otherwise the REST endpoint fails against its own gRPC backend.
+	var (
+		cachedCert *cachedTLSCert
+		policy     tlspolicy.Policy
+		tlsEnabled bool
+	)
+	if viper.IsSet("http-tls-certificate") && viper.IsSet("http-tls-key") {
+		var err error
+		cachedCert, err = newCachedTLSCert(viper.GetString("http-tls-certificate"), viper.GetString("http-tls-key"))
+		if err != nil {
+			return fmt.Errorf("loading duplex TLS certificate: %w", err)
+		}
+		defer cachedCert.Watcher.Close()
+		policy, err = resolveTLSPolicy()
+		if err != nil {
+			return err
+		}
+		tlsEnabled = true
+	}
+
+	loopbackCreds := insecure.NewCredentials()
+	if tlsEnabled {
+		/* #nosec G402 */ // InsecureSkipVerify is only used for the in-process loopback dial to the TLS-enabled duplex listener.
+		loopbackCreds = credentials.NewTLS(&tls.Config{InsecureSkipVerify: true})
+	}
+
 	d := duplex.New(
 		port,
-		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithTransportCredentials(loopbackCreds),
 		grpc.KeepaliveParams(keepalive.ServerParameters{
 			MaxConnectionIdle: viper.GetDuration("idle-connection-timeout"),
 		}),
@@ -558,6 +615,13 @@ func StartDuplexServer(ctx context.Context, cfg *config.FulcioConfig, ctClient *
 	lis, err := net.Listen("tcp", fmt.Sprintf("%s:%d", host, port))
 	if err != nil {
 		return fmt.Errorf("creating listener: %w", err)
+	}
+
+	// Wrap the shared listener with TLS when configured so the duplex server
+	// negotiates HTTP/2 (required by gRPC) over TLS via ALPN.
+	if tlsEnabled {
+		lis = tls.NewListener(lis, cachedCert.tlsConfig(policy, tlspolicy.ALPNProtocols))
+		logger.Info("Duplex server TLS enabled")
 	}
 	logger.Info("Starting duplex server...")
 	if err := d.Serve(ctx, lis); err != nil {
--- a/cmd/app/serve_test.go
+++ b/cmd/app/serve_test.go
@@ -17,13 +17,18 @@ package app
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"io"
 	"log"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+	"os"
+	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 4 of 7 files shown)


bobcallaway:
qq: if we never terminated tls on HTTP, why not just set the default to 1.3 across the board?

anithapriyanatarajan:
> qq: if we never terminated tls on HTTP, why not just set the default to 1.3 across the board?

Thank you. This now unifies on TLS 1.3 across all serving paths - gRPC was already 1.3, so nothing changes there. it only defaults/raises the new HTTP/duplex path, with no regression. And since it's configurable, any operator whose path includes a sidecar, load balancer, or proxy **that requires TLS 1.2** can **opt down** with --tls-min-version 1.2. Updating the Changes to reflect this.

cmd/app/serve.go
```diff
@@ -122,6 +123,10 @@ func newServeCmd() *cobra.Command {
 	cmd.Flags().Duration("read-header-timeout", 10*time.Second, "The time allowed to read the headers of the requests in seconds")
 	cmd.Flags().String("grpc-tls-certificate", "", "the certificate file to use for secure connections - only applies to grpc-port")
 	cmd.Flags().String("grpc-tls-key", "", "the private key file to use for secure connections (without passphrase) - only applies to grpc-port")
+	cmd.Flags().String("http-tls-certificate", "", "the certificate file to use for secure connections on the HTTP endpoint (two-listener mode) and the shared listener (duplex mode)")
+	cmd.Flags().String("http-tls-key", "", "the private key file to use for secure connections (without passphrase) on the HTTP endpoint (two-listener mode) and the shared listener (duplex mode)")
+	cmd.Flags().String("tls-min-version", "", "minimum TLS version for all serving paths (1.2 or 1.3); when unset defaults to 1.3. Lower to 1.2 for peers that cannot negotiate TLS 1.3.")
+	cmd.Flags().StringSlice("tls-cipher-suites", nil, "allowed TLS 1.2 cipher suite names (Go stdlib spelling) applied to all serving paths; empty keeps the Go default. Has no effect on TLS 1.3.")
```
Hayden-IO: Rather than go std lib, can we say crypto/tls?

cmd/app/tls.go
Hayden-IO, line 1: Can these functions go under an internal package like internal/tls? If they’re not exported, I’d prefer to not grow the cmd package.

cmd/app/grpc.go
```diff
@@ -182,8 +192,15 @@ func createGRPCServer(cfg *config.FulcioConfig, ctClient *ctclient.LogClient, ba
 			return nil, err
 		}
 
+		// Defaults to a TLS 1.3 floor unless an operator lowers it explicitly
```
Hayden-IO: Comment isn’t needed if values are set elsewhere

cmd/app/serve.go
```diff
@@ -122,6 +123,10 @@ func newServeCmd() *cobra.Command {
 	cmd.Flags().Duration("read-header-timeout", 10*time.Second, "The time allowed to read the headers of the requests in seconds")
 	cmd.Flags().String("grpc-tls-certificate", "", "the certificate file to use for secure connections - only applies to grpc-port")
 	cmd.Flags().String("grpc-tls-key", "", "the private key file to use for secure connections (without passphrase) - only applies to grpc-port")
+	cmd.Flags().String("http-tls-certificate", "", "the certificate file to use for secure connections on the HTTP endpoint (two-listener mode) and the shared listener (duplex mode)")
+	cmd.Flags().String("http-tls-key", "", "the private key file to use for secure connections (without passphrase) on the HTTP endpoint (two-listener mode) and the shared listener (duplex mode)")
+	cmd.Flags().String("tls-min-version", "", "minimum TLS version for all serving paths (1.2 or 1.3); when unset defaults to 1.3. Lower to 1.2 for peers that cannot negotiate TLS 1.3.")
```
Hayden-IO, line 129: I do wonder if there’s a need for 1.2. 1:3 is almost 10 years old. Like I mentioned, I don’t really want to be overly opinionated, but I also don’t want to support more than we have to.

cmd/app/serve.go
```diff
@@ -201,6 +206,11 @@ func runServeCmd(cmd *cobra.Command, args []string) { //nolint: revive
 	viper.SetEnvPrefix(serveCmdEnvPrefix)
 	viper.AutomaticEnv()
 
+	// Fail fast on obviously invalid TLS configuration before any listeners start.
```
Hayden-IO: Comment doesn’t add additional detail, can be removed

Hayden-IO (changes requested):

cmd/app/serve.go
Hayden-IO: There’s a number of comments that fall under this as well, can you prune comments that aren’t adding additional context?

cmd/app/serve.go
anithapriyanatarajan: I agree to your point regarding 1.3.  The main reason I  prefer 1.2 as non default option is TLS 1.2 is not yet deprecated. RFC 8996 only deprecated 1.0/1.1. 1.2 is frozen (RFC 9851), not sunset. Most Managed Kubernetes platforms and many load balancers still run a 1.2 minimum, so dropping it would block valid deployments pinned to 1.2. 

we could open a future issue to deprecate 1.2. Hope this is acceptable?

anithapriyanatarajan:
Marked as draft to rework the package layout as stated in comments. Thank you.

cmd/app/serve.go
anithapriyanatarajan: Updated.

cmd/app/grpc.go
anithapriyanatarajan: Done.

cmd/app/tls.go
anithapriyanatarajan: moved to internal/tlspolicy. To avoid confusion with crypto/tls named it as internal/tlspolicy

cmd/app/serve.go
anithapriyanatarajan: Addressed.

docs/setup.md
```diff
@@ -238,3 +238,53 @@ To get the root certificate, call `curl -o fulcio.crt.pem http://localhost:5555/
...
+terminator and the Fulcio pod is a separate trust boundary (for example a re-encrypting
+ingress), so that the OIDC bearer token and CSR are not sent in cleartext to the pod.
+
+All TLS flags are optional and default to today's behaviour when unset, so existing
+deployments are unaffected.
+
+### Serving material
+
+* `--grpc-tls-certificate` / `--grpc-tls-key`: serve TLS on the gRPC listener
+  (two-listener mode only). Unchanged from earlier releases.
```
Hayden-IO: Let's remove "Unchanged from earlier releases"

docs/setup.md
```diff
@@ -238,3 +238,53 @@ To get the root certificate, call `curl -o fulcio.crt.pem http://localhost:5555/
...
+* `--http-tls-certificate` / `--http-tls-key`: serve TLS on the HTTP/REST listener
+  (two-listener mode) and on the shared listener in duplex mode (when the HTTP and gRPC
+  host:port match).
+
+For each pair, the certificate and key must be provided together. Certificate rotation on
+disk is picked up without a restart, matching the existing gRPC behaviour.
+
+### Policy
+
```
Hayden-IO, line 263: This is a bit verbose, I think these flags are self-documenting.

cmd/app/serve.go
```diff
@@ -548,6 +605,13 @@ func StartDuplexServer(ctx context.Context, cfg *config.FulcioConfig, ctClient *
 	if err != nil {
 		return fmt.Errorf("creating listener: %w", err)
 	}
+
+	// Wrap the shared listener with TLS when configured so the duplex server
+	// negotiates HTTP/2 (required by gRPC) over TLS via ALPN.
+	if tlsEnabled {
```
Hayden-IO, line 629: Have you been able to test this using the duplex setup?

cmd/app/serve.go
```diff
@@ -186,6 +192,24 @@ func (rt *hostRoundTripper) RoundTrip(req *http.Request) (*http.Response, error)
...
 }
 
+func resolveTLSPolicy() (tlspolicy.Policy, error) {
+	return tlspolicy.Resolve(viper.GetString("tls-min-version"), viper.GetStringSlice("tls-cipher-suites"))
+}
+
+// validateTLSFlags rejects obvious TLS misconfiguration before any listener
+// starts: a certificate without its matching key, or an invalid minimum
+// version or cipher suite name.
+func validateTLSFlags() error {
```
Hayden-IO, line 203: Isn't resolveTLSPolicy called right after validate, is this needed? Can you just check in resolve?

cmd/app/serve.go
```diff
@@ -496,9 +524,38 @@ func duplexHealthz(_ context.Context, mux *runtime.ServeMux, endpoint string, op
...
+		policy     tlspolicy.Policy
+		tlsEnabled bool
+	)
+	if viper.IsSet("http-tls-certificate") && viper.IsSet("http-tls-key") {
+		var err error
+		cachedCert, err = newCachedTLSCert(viper.GetString("http-tls-certificate"), viper.GetString("http-tls-key"))
+		if err != nil {
+			return fmt.Errorf("loading duplex TLS certificate: %w", err)
+		}
+		defer cachedCert.Watcher.Close()
```
Hayden-IO, line 542: do you want to close the watcher here?

cmd/app/http.go
```diff
@@ -112,11 +115,32 @@ func createHTTPServer(ctx context.Context, serverEndpoint string, grpcServer, le
 		WriteTimeout:      60 * time.Second,
 		IdleTimeout:       viper.GetDuration("idle-connection-timeout"),
 	}
-	return httpServer{&api, serverEndpoint}
+
+	// Optionally terminate TLS on the HTTP listener; when unset the endpoint
+	// stays plaintext. cachedTLSCert serves the certificate so on-disk rotation
```
Hayden-IO: I know we do this in other another part of the code base, but I’m not certain we should continue this behavior. Updates mid-process are risky and I’d prefer to encourage a server restart instead.

@bobcallaway do you recall why we did this for the gRPC path? I thought that was because the TLS cert was just to appease the LB

cmd/app/serve.go
anithapriyanatarajan: Thank you for emphasizing on this. I now tested the duplex setup. That's how I ran into the startup panic go-grpc-kit's RegisterListenAndServeMetrics re-registers the gRPC collectors on the default registry and collides with grpc_prometheus. So in this PR I now serve /metrics from fulcio's own handler (like two-listener mode), which avoids the collision, and I raised #2419 to remove the deprecated go-grpc-prometheus dependency as a proper follow-up.

docs/setup.md
anithapriyanatarajan: Done. Thank you.

cmd/app/serve.go
anithapriyanatarajan: #2419 is closed and the changes were merged. Rebased this PR, removed the stop gap correction and retested. All modes start fine without panic.

cmd/app/serve.go
anithapriyanatarajan: validateTLSFlags, only checks that a cert and its key are supplied together (for both the grpc and http pairs). resolveTLSPolicy, evaluates the min-version and cipher-suite values, so it only runs once TLS is already on. If someone sets just --http-tls-certificate and forgets the key, the TLS branch is skipped and resolve is never reached,  the endpoint would silently serve plaintext. The separate check fails fast on that half-configured case instead, so shall we keep this way.

cmd/app/serve.go
anithapriyanatarajan: Yeah, I think this is fine.

cmd/app/http.go
anithapriyanatarajan: This just inherited the existing pattern introduced in #1256, so it's consistent with the current gRPC path rather than new behavior. Shall we leave it as-is  or if we feel restart-only is the right direction, I'll raise a follow-up that addresses all paths together.

anithapriyanatarajan:
@Hayden-IO @bobcallaway , I have addressed all the comments. Could you take a look and help merge this? Thank you

cmd/app/serve.go
bobcallaway: let's leave deprecation out of scope for this PR

cmd/app/http.go
bobcallaway: you're correct. when i wrote the other code i assumed (incorrectly) that the gcp load balancer checked the validity of the certificate presented by the server. we should probably skip this here as there are better ways to do this outside of checking

cmd/app/serve.go
```diff
@@ -186,6 +192,24 @@ func (rt *hostRoundTripper) RoundTrip(req *http.Request) (*http.Response, error)
...
+func resolveTLSPolicy() (tlspolicy.Policy, error) {
+	return tlspolicy.Resolve(viper.GetString("tls-min-version"), viper.GetStringSlice("tls-cipher-suites"))
+}
+
+// validateTLSFlags rejects a certificate supplied without its matching key on
+// either serving path before any listener starts. The minimum version and
+// cipher suite flags are validated by resolveTLSPolicy where the policy is
+// applied.
+func validateTLSFlags() error {
+	if viper.IsSet("grpc-tls-certificate") != viper.IsSet("grpc-tls-key") {
```
Hayden-IO: Can you use https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsRequiredTogether?

cmd/app/grpc.go
```diff
@@ -107,7 +108,7 @@ func newCachedTLSCert(certPath, keyPath string) (*cachedTLSCert, error) {
 					return
 				}
 				if event.Has(fsnotify.Write) {
-					log.Logger.Info("fsnotify grpc-tls-certificate write event detected")
```
Hayden-IO: isn't this still the grpc TLS cert?

cmd/app/grpc.go
```diff
@@ -142,20 +143,30 @@ func (c *cachedTLSCert) UpdateCertificate() error {
...
+	/* #nosec G402 */ // MinVersion is a variable, but tlspolicy.ParseVersion constrains it to TLS 1.2 or 1.3 and callers default it to a literal >= TLS 1.2, so it is never below the G402 floor.
+	return &tls.Config{
 		GetCertificate: func(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
 			return c.GetCertificate(), nil
 		},
-		MinVersion: tls.VersionTLS13,
-	}))
+		MinVersion:   p.MinVersion,
+		CipherSuites: p.CipherSuites,
+		NextProtos:   nextProtos,
```
Hayden-IO: Is NextProtos required? This stands out as unexpected.

docs/setup.md
```diff
@@ -238,3 +238,52 @@ To get the root certificate, call `curl -o fulcio.crt.pem http://localhost:5555/
...
+* `--http-tls-certificate` / `--http-tls-key`: serve TLS on the HTTP/REST listener, and on
+  the shared listener in duplex mode.
+
+Each certificate and key must be provided together. The gRPC listener picks up
+on-disk certificate rotation without a restart. For the HTTP/REST listener and the
+shared duplex listener, restart Fulcio to load a rotated certificate.
+
+### Policy
+
+* `--tls-min-version`: minimum negotiated TLS version, `1.2` or `1.3`. When unset, every
```
Hayden-IO: I'd just say this can be 1.3 or 1.2, I don't think we need more explanation.

docs/setup.md
```diff
@@ -238,3 +238,52 @@ To get the root certificate, call `curl -o fulcio.crt.pem http://localhost:5555/
...
+  TLS-enabled serving path defaults to a `1.3` floor. This is a floor, not a ceiling:
```
Hayden-IO: I'd drop the word 'floor' throughout this for clarity.

internal/tlspolicy/tlspolicy.go
```diff
@@ -0,0 +1,122 @@
...
+// under the policy. The cert is loaded once by the caller, so updating it on
+// disk needs a restart. nextProtos sets the ALPN list (use ALPNProtocols for
+// the HTTP and duplex listeners; nil for gRPC, which handles HTTP/2 itself).
+func (p Policy) ServerConfig(cert tls.Certificate, nextProtos []string) *tls.Config {
+	/* #nosec G402 */ // MinVersion is a variable, but ParseVersion constrains it to TLS 1.2 or 1.3 and Resolve defaults it to DefaultMinVersion (TLS 1.3), so it is never below the G402 floor.
+	return &tls.Config{
+		Certificates: []tls.Certificate{cert},
+		MinVersion:   p.MinVersion,
+		CipherSuites: p.CipherSuites,
+		NextProtos:   nextProtos,
```
Hayden-IO: In both places ServerConfig is called, ALPNProtocols is passed. Can the value just be hardcoded? Also, this stands out to me as unexpected, do we need NextProtos?

Hayden-IO (approved):
LGTM, thanks! @bobcallaway Anything else?

internal/tlspolicy/tlspolicy.go
```diff
@@ -0,0 +1,122 @@
...
+// under the policy. The cert is loaded once by the caller, so updating it on
+// disk needs a restart. It always advertises ALPNProtocols so gRPC-over-HTTP/2
+// still negotiates on the HTTP and shared duplex listeners.
+func (p Policy) ServerConfig(cert tls.Certificate) *tls.Config {
+	/* #nosec G402 */ // MinVersion is set by ParseVersion/Resolve to TLS 1.2 or 1.3 only.
+	return &tls.Config{
+		Certificates: []tls.Certificate{cert},
+		MinVersion:   p.MinVersion,
+		CipherSuites: p.CipherSuites,
+		NextProtos:   ALPNProtocols,
```
Hayden-IO: Do you know why we need this? From reading over the docs, it's still not clear to me why this is necessary.

cmd/app/serve.go
```diff
@@ -135,6 +141,9 @@ func newServeCmd() *cobra.Command {
 		v1.PublicKeyDetails_PKIX_ED25519,
 	}), "the list of allowed client signing algorithms")
 
+	cmd.MarkFlagsRequiredTogether("grpc-tls-certificate", "grpc-tls-key")
```
Hayden-IO: I was looking more into this, it looks like this checks flag values, but would not consider a config file or the env vars like FULCIO_SERVE_HTTP_TLS_KEY. Can you double check this, and if so, we'll have to go back to checking later on?

internal/tlspolicy/tlspolicy.go
```diff
@@ -0,0 +1,122 @@
...
+	if id, ok := supportedVersions[strings.TrimSpace(v)]; ok {
+		return id, nil
+	}
+	return 0, fmt.Errorf("unsupported --tls-min-version %q (supported: %s)", v, strings.Join(supportedVersionNames(), ", "))
+}
+
+// ParseCipherSuites resolves cipher suite names to their crypto/tls IDs. Only
+// suites reported as secure by crypto/tls are accepted; names must match the
+// crypto/tls spelling (e.g. TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256).
+func ParseCipherSuites(names []string) ([]uint16, error) {
```
Hayden-IO: Can you also verify that the cipher suite is actually for TLS 1.2 and not for TLS 1.3? For example, what happens when you provide TLS_AES_128_GCM_SHA256?

internal/tlspolicy/tlspolicy.go
```diff
@@ -0,0 +1,122 @@
...
+
+// DefaultMinVersion is applied when --tls-min-version is unset. TLS 1.3 keeps
+// every serving path secure by default; lower it to 1.2 only for peers that
+// cannot negotiate 1.3.
+const DefaultMinVersion = tls.VersionTLS13
+
+// Policy is the resolved TLS policy for a serving path.
+type Policy struct {
+	MinVersion uint16
+	// CipherSuites restricts the allowed TLS 1.2 suites; nil keeps the
```
Hayden-IO: Should we throw an error if you set the min TLS version to 1.3 but provide cipher suites?

internal/tlspolicy/tlspolicy.go
```diff
@@ -0,0 +1,122 @@
...
+import (
+	"crypto/tls"
+	"fmt"
+	"sort"
+	"strings"
+)
+
+// ALPNProtocols advertises HTTP/2 (required by gRPC) ahead of HTTP/1.1 on
+// TLS-enabled HTTP and duplex listeners.
+var ALPNProtocols = []string{"h2", "http/1.1"}
```
Hayden-IO: This can be unexported.

cmd/app/http.go
```diff
@@ -112,11 +112,29 @@ func createHTTPServer(ctx context.Context, serverEndpoint string, grpcServer, le
 		WriteTimeout:      60 * time.Second,
 		IdleTimeout:       viper.GetDuration("idle-connection-timeout"),
 	}
+
+	// Optionally terminate TLS on the HTTP listener. The certificate is loaded
+	// once at startup; rotating it on disk requires a restart.
+	if viper.IsSet("http-tls-certificate") && viper.IsSet("http-tls-key") {
```
Hayden-IO: Can you check if IsSet checks that the flag is not empty? IIRC, it doesn't, so you'd need to check `!= ""`.

cmd/app/serve_test.go
```diff
@@ -204,3 +210,272 @@ func TestServeCmdFlags(t *testing.T) {
...
+		if _, err := resolveTLSPolicy(); err == nil {
+			t.Fatal("expected error for unknown cipher suite")
+		}
+	})
+}
+
+// TestDuplexTLS exercises the duplex serving path with TLS enabled end to end:
+// the REST endpoint must be reachable over HTTPS and a client that caps below
+// the default TLS 1.3 minimum must be rejected by the server.
+func TestDuplexTLS(t *testing.T) {
```
Hayden-IO: Should we be doing the same registry swapping as in TestDuplex?

cmd/app/serve_test.go
```diff
@@ -204,3 +210,272 @@ func TestServeCmdFlags(t *testing.T) {
...
+	dir := t.TempDir()
+	certPath := filepath.Join(dir, "cert.pem")
```
Hayden-IO: You can also use writeTLSKeyPair

cmd/app/serve_test.go
```diff
@@ -204,3 +210,272 @@ func TestServeCmdFlags(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	if grpcServer.tlsCertWatcher != nil {
+		defer grpcServer.tlsCertWatcher.Close()
+	}
+
+	srv := createHTTPServer(context.Background(), addr, grpcServer, nil)
+	var wg sync.WaitGroup
+	srv.startListener(&wg)
+	defer srv.Shutdown(context.Background())
```
Hayden-IO: Can you run a goroutine leak checker and verify that the server is properly cleaned up after this test runs? Since it waits for idleConnsClosed, i'm not sure it will shutdown.

Hayden-IO (changes requested):
Sorry, did a second pass and have some small comments

PR #2428, opened 2026-08-10
Replace archived go-grpc-prometheus with go-grpc-middleware prometheus provider

### LINKED ISSUE #2419: Replace archived go-grpc-prometheus with go-grpc-middleware prometheus provider
## Summary

Fulcio still imports the archived, deprecated `github.com/grpc-ecosystem/go-grpc-prometheus` for gRPC server metrics. Upstream declares it deprecated and directs users to
`github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus`, which Fulcio already pulls in indirectly (via `chainguard.dev/go-grpc-kit`). We should migrate our direct usage over and drop the deprecated dependency.

Ref: https://github.com/grpc-ecosystem/go-grpc-prometheus#deprecated-go-grpc-interceptors-for-prometheus-monitoring

## Current usage

The deprecated package is imported directly in two places:

- `cmd/app/grpc.go` - two-listener mode: `grpc_prometheus.DefaultServerMetrics`, `EnableHandlingTimeHistogram()`, `grpc_prometheus.Register(...)`, and the
  `grpc_prometheus.UnaryServerInterceptor` in the interceptor chain.
- `cmd/app/serve.go` - duplex mode: the same registration pattern and interceptor.

Both rely on the deprecated library's *default* server metrics registered on the *default* Prometheus registry.

## Why it wasn't done already

This was attempted in #1968 but closed unmerged. The blocker, per that PR's own findings, was the duplex server: it pulled a conflicting registration path through go-grpc-kit, and Fulcio's code assumed the default server metrics / default registerer, so the deprecated library couldn't be fully removed while duplex
depended on that path.

That specific blocker has since been reduced: PR #2418 (Work In Progress while this Issue was raised)  changed duplex mode to serve `/metrics` from Fulcio's own `promhttp.Handler()` (default registry) instead of go-grpc-kit's `RegisterListenAndServeMetrics`, so Fulcio no longer depends on go-grpc-kit's metrics registration. The remaining work is swapping our direct interceptor + collector usage to the new provider.

## Proposed change

- Replace `grpc_prometheus.DefaultServerMetrics` / `Register` / `UnaryServerInterceptor` in both `cmd/app/grpc.go` and `cmd/app/serve.go` with the equivalents from `github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus`  (`NewServerMetrics()`, `metrics.UnaryServerInterceptor`, register the collector).
- Keep serving `/metrics` from the default registry so scrape output stays compatible across two-listener and duplex modes.
- Remove `github.com/grpc-ecosystem/go-grpc-prometheus` from `go.mod` once no direct references remain.

## Scope / non-goals

- No change to exposed metric names, ports, or the `/metrics` path.
- Decide explicitly whether to adopt the provider's optional labels (e.g. the `cgclientid` label go-grpc-kit adds); default is to preserve today's label set to avoid dashboard/alert churn.
- TLS work (#2415 / #2418) is unrelated and but this issue was identified during review (https://github.com/sigstore/fulcio/pull/2418#discussion_r3730901712) 

## Acceptance criteria

- No direct import of `go-grpc-prometheus` remains; it is dropped from `go.mod`.
- `grpc_server_*` metrics are exposed and increment in both two-listener and duplex modes, with an unchanged label set (unless a label change is explicitly agreed).
- `go build ./...`, `go vet ./...`, `golangci-lint run ./...`, and `go test ./...`  all pass.

## References

- #1968:  prior (closed) attempt at this migration
- #974:  Fulcio v2 (maintainers suggested revisiting there)
- #2418:  removed the duplex metrics-registration blocker

anithapriyanatarajan:
## Summary

Fulcio imported the archived, deprecated `github.com/grpc-ecosystem/go-grpc-prometheus` for gRPC server metrics.
Upstream directs users to `github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus`, which
Fulcio already pulled in indirectly via `chainguard.dev/go-grpc-kit`. This migrates the direct usage over and drops the deprecated dependency.

Fixes #2419

## Changes

- `cmd/app/grpc.go` (two-listener mode) and `cmd/app/serve.go` (duplex mode): build a single `*prometheus.ServerMetrics` via `NewServerMetrics` with  `WithServerHandlingTimeHistogram()` (equivalent to the old
  `EnableHandlingTimeHistogram()`), share it across the v2 and legacy gRPC servers, and wire its `UnaryServerInterceptor()` into the chain.
- Register the collector on the default registry and call  `InitializeMetrics()` after all services are registered.
- Duplex mode now serves `/metrics` from Fulcio's own `promhttp.Handler()`  on the default registry instead of go-grpc-kit's  `RegisterListenAndServeMetrics`, which registered its own `grpc_server_*`  collector on the default registry and caused a duplicate-collector panic.
- Dropped `go-grpc-prometheus` from `go.mod`.

## Metrics impact

No change to metric names, ports, the `/metrics` path, or the label set. The same five `grpc_server_*` families are exported with the same labels; the `cgclientid` label is not adopted, so there is no dashboard/alert churn.

## Notes

Independent of #2418 (TLS work), which tactically included fix for duplex metrics panic. Whichever merges first, I shall reconcile the other.


Hayden-IO:
@loosebazooka can you take a look?

loosebazooka:
Just a high level question before I dig into this: Are we unable to avoid using the Global Default Registries here? I remember that was one of the issues I was running into because of how duplex was set up. Ideally we register our own NewRegistry and work from there.

anithapriyanatarajan:
> Just a high level question before I dig into this: Are we unable to avoid using the Global Default Registries here? I remember that was one of the issues I was running into because of how duplex was set up. Ideally we register our own NewRegistry and work from there.

Thank you for the query. Summarizing below my analysis:

In duplex mode, two separate components were each registering the same grpc_server_* metrics on the default registry:
1. go-grpc-prometheus:  auto-registered them in its package init() just by being imported.
2. go-grpc-kit's RegisterListenAndServeMetrics (used only in duplex mode):  registered its own grpc_server_* collector on the default registry too. Prometheus only allows a metric to be registered once per registry, so in duplex mode the second registration panicked on startup (the panic #2418 stop-gapped) but we did not merge and proceeded to do this PR. 

This issue never affected the Two-listener mode since that mode never calls the go-grpc-kit helper, so it was unaffected.

This PR leverages the go-grpc-middleware/providers/prometheus (which does not auto-register,  we create and register one ServerMetrics instance ourselves), and duplex mode stops using go-grpc-kit's metrics path, serving /metrics from our own http.Server like two-listener mode already does. Net result: a single owner of grpc_server_*, no double registration, no panic. Had this verified running in both modes.

On the Global Default Registry. We are on the default registry today because Fulcio's own metrics (metrics.go) are created with promauto, which auto-registers on the default registry, and /metrics already served promhttp.Handler() (default).The gRPC metrics were also sent to that registry in 2-listener mode. This PR attempted to follow the existing registry pattern

Moving to our own prometheus.NewRegistry() is feasible now,  because this PR removed the duplex dependency on go-grpc-kit's metrics path (which served DefaultGatherer),  so nothing forces the default gatherer anymore. To do it we also need to migrate Fulcio's own metrics off promauto onto the custom registry and re-add the Go/process collectors, then serve promhttp.HandlerFor(reg, …) in both modes. That touches metrics.go beyond the original dependency swap.

Happy to include that in this PR if that is what is preferred.

cmd/app/serve.go
```diff
@@ -529,20 +531,29 @@ func StartDuplexServer(ctx context.Context, cfg *config.FulcioConfig, ctClient *
...
-	grpcMetrics := grpc_prometheus.DefaultServerMetrics
-	grpcMetrics.EnableHandlingTimeHistogram()
-	reg.MustRegister(grpcMetrics, server.MetricLatency, server.RequestsCount)
-	grpc_prometheus.Register(d.Server)
+	// InitializeMetrics must be called after all services are registered.
+	prometheus.MustRegister(grpcMetrics)
+	grpcMetrics.InitializeMetrics(d.Server)
 
 	// Healthz
 	health.RegisterHealthServer(d.Server, grpcCAServer)
```
loosebazooka, line 534: should RegisterHealthServer go before Initialize metrics?

loosebazooka (commented):
oh okay I see yeah. I guess we can follow up with moving the non-default global registry in a follow up PR. In that case this is basically LGTM, one minor thing.

Note to myself: And I should go update rekor-tiles to add in the go/process collectors as well.

cmd/app/serve.go
anithapriyanatarajan: Thank you, addressed this now. In duplex mode the health service is now registered before InitializeMetrics, so its metrics are pre-seeded to 0 from server start (matching two-listener mode).

loosebazooka (approved):
Thanks!

loosebazooka:
@anithapriyanatarajan as a followup can we see if it's possible to move away from the defaultregistry?

PR #2450, opened 2026-09-16
fix(gitlab): derive GitLab base URL from token issuer instead of hardcoding

### LINKED ISSUE #2445: GitLab provider hardcodes `https://gitlab.com/`, so three certificate extensions name the wrong host on self-hosted instances
## Description

`pkg/identity/gitlabcom/principal.go` sets the principal's base URL to the
string literal `https://gitlab.com/` (currently line ~175), even though the
token's issuer is already available on the same struct as `token.Issuer`:

```go
return &jobPrincipal{
    subject: token.Subject,
    issuer:  token.Issuer,          // correct instance URL
    url:     `https://gitlab.com/`, // hardcoded
    ...
```

`Embed` then parses that literal into `baseURL` and derives three extensions
from it:

```go
SourceRepositoryURI:      baseURL.JoinPath(p.repository).String(),
SourceRepositoryOwnerURI: baseURL.JoinPath(p.repositoryOwner).String(),
RunInvocationURI:         baseURL.JoinPath(p.repository, "/-/jobs/", p.jobID).String(),
```

On a self-hosted GitLab those three certificate extensions
(`1.3.6.1.4.1.57264.1.12`, `.1.16` and `.1.21`) therefore point at
`gitlab.com` paths that have nothing to do with the instance that actually
ran the build — and, in the general case, at a namespace on a public host
that the signing organisation does not control.

Every other field is correct, which is what makes this easy to miss: the SAN,
Build Signer URI (`.1.9`) and Build Config URI (`.1.18`) all derive from the
`ci_config_ref_uri` claim and carry the real host, and the non-URL extensions
(repository/signer digest, project id, ref, runner environment, visibility)
are accurate. So signature verification against SAN + issuer is unaffected;
the damage is to anyone using the source-repository OIDs as a defence-in-depth
check, where the value silently describes the wrong host.

There is a second, smaller consequence in the same function. The scheme
normalisation

```go
if baseURL.Host == ciConfigRefURL.Host {
    ciConfigRefURL.Scheme = baseURL.Scheme
}
```

can never match on a self-hosted instance, so the SAN scheme always falls back
to the `https` default rather than following the issuer.

## Reproduce

Sign anything with a self-hosted GitLab as the OIDC issuer, then decode the
signing certificate's extensions — `.1.12`, `.1.16` and `.1.21` will be
`https://gitlab.com/<path>` while the SAN and `.1.8` carry the real instance.

## Suggested fix

Derive the base URL from `token.Issuer` instead of the literal. That is
already the value used for the `Issuer` extension, and on gitlab.com it is
`https://gitlab.com`, so hosted behaviour is unchanged.

Worth confirming against the trust model: because the issuer is what the
provider is configured to accept, taking the host from it should not widen
what a token can assert. If that reasoning is wrong, an explicit
per-issuer configured instance URL would be the alternative.

AruneshDwivedi:
## Problem

The GitLab provider in `pkg/identity/gitlabcom/principal.go` hardcoded `https://gitlab.com/` as the base URL for source-repository certificate extensions, even though the actual issuer is available on the same struct as `token.Issuer`.

On self-hosted GitLab instances this caused three OID certificate extensions to point at `gitlab.com` paths instead of the real instance:
- `1.3.6.1.4.1.57264.1.12` (SourceRepositoryURI)
- `1.3.6.1.4.1.57264.1.16` (SourceRepositoryOwnerURI)
- `1.3.6.1.4.1.57264.1.21` (RunInvocationURI)

All other fields (SAN, Build Signer URI `.1.9`, Build Config URI `.1.18`) already derive correctly from the `ci_config_ref_uri` claim and carry the real host. The damage was limited to anyone using the source-repository OIDs as a defence-in-depth check.

There is also a secondary bug in the same function: the scheme normalisation block that checks `baseURL.Host == ciConfigRefURL.Host` can never match on a self-hosted instance, so the SAN scheme always falls back to `https` rather than following the issuer.

## Fix

Derive the base URL from `token.Issuer + "/"` instead of the literal. On gitlab.com the issuer is already `https://gitlab.com`, so hosted behaviour is unchanged. Self-hosted users now get correct extension values.

```release-notes
[BUGFIX] GitLab provider: use token issuer as base URL instead of hardcoded gitlab.com
```

Fixes sigstore/fulcio#2445


Hayden-IO:
See comment on issue.

